PR #1, opened 2016-10-29
Fixes DeleteDirectory infinite recursion calls

NerijusD:
Infinite recursion calls happens, when FtpListItem is referencing to parent or current directory and when `MLSD` command is used for `GetListing` method.


hgupta9:
Fantastic! Thanks for your contribution.

PR #2, opened 2016-10-29
Improvements delete directory method.

NerijusD:
These changes affect `DeleteDirectory` when called recursively with `force` flag set. 

Changes ensure that:
- `GetListing` method is called only once - all data regarding files / directories are downloaded in single big chunk instead of making separate small requests. Thus improves bandwidth usage and response time.
- Files and directories are deleted in right order, based on depth of directional hierarchy and type.

### DIFF
--- a/FluentFTP/FtpClient.cs
+++ b/FluentFTP/FtpClient.cs
@@ -10,6 +10,7 @@ using System.Diagnostics;
 using System.Web;
 using System.Security.Cryptography.X509Certificates;
 using System.Globalization;
+using System.Linq;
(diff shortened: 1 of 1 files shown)


hgupta9:
I merged this manually in [this commit](https://github.com/hgupta9/FluentFTP/commit/30d22cb126b4c3eabedadccb38f96170c5fb726c).

Because I could not test with multiple servers (`WasGetListingRecursive()`) I added a boolean `fastMode`. I don't want to break the DeleteDirectory method.

Issue #7, opened 2016-11-06
Proxy Implementation?


Cocotus:
Hi,

first thanks for your hard work, really appreciated. I like your library and it works fine for me, but can you show me how to implement proxy support for a connection? I have additional data from my VPN provider I would like to user for FTP connections (Proxy username, proxy server, proxy passwort) - in the past I could insert those in FTP client like Filezilla. Now I looked at avalaible properties in FTPFluent but right now, theres's no way to use proxy authentification data? 

I'm referring to this code:

```
public static class ConnectExample {
        public static void Connect() {
            using (FtpClient conn = new FtpClient()) {
                conn.Host = "localhost";
                conn.Credentials = new NetworkCredential("ftptest", "ftptest");
                conn.Connect();
            }
        }
}
```
Any chance for proxy support in future? Thanks!

hgupta9:
There is no support for FTP proxies atm, but it can be built. FtpClient will have to be subclassed, and the  [Authenticate](https://github.com/hgupta9/FluentFTP/blob/d7f0eae9c5f84d5ee1e9325ece42683c6986d490/FluentFTP/FtpClient.cs#L876) and Execute methods will have to be modified. Probably others too.

Cocotus:
Thanks for your reply! Sounds like some work, but I think would be cool feature in future! If you need someone to test it, leave me a message, I'm up for the task!

L3Z4:
Hello,
Is someone currently working on it ?
I can develop it, I just want to be sure the work will not be done twice :)

Cocotus:
> Hello,
> Is someone currently working on it ?
> I can develop it, I just want to be sure the work will not be done twice :)

Cool! I don't have any experience with FTP communication, so I cant and won't develop anything regarding this feature - but I'm psyched up nonetheless ;)

zharris6:
I am actually working on it and have it working with a SOCKS5 proxy atm. 

It is a very very very initial POC but it is working none the less.

As soon as I can fix this bug (man is it frustrating) I will get something committed.
https://github.com/hgupta9/FluentFTP/issues/9

zharris6:
L3Z4 - You can give it a go. My branch differs greatly from the master and the changes would be to much to overcome.

With that being said, i would be more then happy to help you out if you run into any issues!

hgupta9:
Proxy support has been added in FluentFTP 16.0.14. I'm closing this issue for now.

pwallner:
[...]

PR #11, opened 2016-11-18
New - Support proxy HTTP 1.1/ User@Host

L3Z4:
For #7 
HTTP1.1 proxy with or without authentication,
User@Host, with or without authentication.

Futur proxy implementation will benefit from FtpClientProxy abstract class.

Usage : 
using (FtpClient cl = new FtpClientHttp11Proxy(new Proxy { Host = "127.0.0.1", Port = 3128, }))
...

### DIFF
--- a/FluentFTP.Tests/Program.cs
+++ b/FluentFTP.Tests/Program.cs
@@ -25,7 +25,7 @@ namespace Tests {
             FtpTrace.AddListener(new ConsoleTraceListener());
 
             try {
-				/*foreach (int i in new int[] {
+                /*foreach (int i in new int[] {
                      (int)FtpDataConnectionType.EPSV,
                      (int)FtpDataConnectionType.EPRT,
                      (int)FtpDataConnectionType.PASV,
@@ -47,9 +47,9 @@ namespace Tests {
 
                 //StreamResponses();
 
-				//TestServer();
+                //TestServer();
 
-				//TestManualEncoding();
+                //TestManualEncoding();
 
                 //TestServer();
 
@@ -84,6 +84,8 @@ namespace Tests {
                 //TestUnixListing();
 
                 TestListPath();
+
+                TestListPathWithHttp11Proxy();
             }
             catch (Exception ex) {
                 Console.WriteLine(ex.ToString());
@@ -91,6 +93,24 @@ namespace Tests {
 
             Console.WriteLine("--DONE--");
             Console.ReadKey();
+        }
+
+        static void TestListPathWithHttp11Proxy()
+        {
+            using (FtpClient cl = new FtpClientHttp11Proxy(new Proxy { Host = "127.0.0.1", Port = 3128, })) // Credential = new NetworkCredential() 
+            {
+                cl.Credentials = new NetworkCredential(m_user, m_pass);
+                cl.Host = m_host;
+                cl.ValidateCertificate += OnValidateCertificate;
+                cl.DataConnectionType = FtpDataConnectionType.PASV;
+                cl.Connect();
+
+                foreach (FtpListItem item in cl.GetListing(null, FtpListOption.SizeModify | FtpListOption.ForceNameList))
+                {
+                    Console.WriteLine(item.Modified.Kind);
+                    Console.WriteLine(item.Modified);
+                }
+            }
         }
 
         static void TestListPath() {
@@ -112,8 +132,8 @@ namespace Tests {
                 cl.Host = m_host;
                 cl.EncryptionMode = FtpEncryptionMode.None;
                 cl.ValidateCertificate += new FtpSslValidation(delegate(FtpClient control, FtpSslValidationEventArgs e) {
-                        e.Accept = true;
-                    });
+                    e.Accept = true;
+                });
 
                 using (FtpDataStream s = (FtpDataStream)cl.OpenWrite("test.txt")) {
                     FtpReply r = s.CommandStatus;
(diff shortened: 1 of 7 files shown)


springcomp:
That would be really useful.
+1

hgupta9:
When you say HTTP1.1 Proxy, Is this for connecting to a FTP through an HTTP proxy? I have never heard of FTP proxy? And what is user@host? Never heard of that either.

L3Z4:
The FTP client you wrote is instantiated in a system inside a network. The outgoing communication can be going through a Proxy. FTP client must adapt its way to connect in this particular case.

For my tests, I used Squid proxy http://www.squid-cache.org/, this particular tutorial : https://docs.diladele.com/tutorials/installing_squid_windows/index.html

I hope it help ! Let me know if I could do anything.

hgupta9:
@L3Z4 - I'm going to test this today, but you could help by testing the new code as well. Since you have a genuine need for a proxy, you will provide valuable feedback. Why are there 2 classes for the Proxy? I think it would be better OOP design if there was just one FtpClientProxy class, or even better if we could merge functionality into the main FtpClient class. Which would you prefer?

L3Z4:
The option A is not acceptable : composition can be numerous. I implemented only two kind of proxy, I don't know how many exists. Idem for option B.

The option C is the closest of the implementation a pushed. I'm convinced that having a class dedicated for a proxy implementation is a good choice (S of SOLID). It is also an extension point of your library. 

I agree to add a factory method in the FtpClient class to construct a "proxified" client.
How do I update the pull request ? Should I update my branch and re-do it ? It is my first one to be honest.

hgupta9:
Just add more commits. Github will update the PR to use your latest commit in your trunk.

I agree. Please add a `ConnectionType` or similar property, and go with Option C. Do you mean a static factory method or a dynamic one?

L3Z4:
I add a property on the `FtpClient` to add the `ConnectionType`. The "set" is protected so the "proxified" implementation can declare when constructed.
I can not bring myself to create an enum to define the ProxyType ; it is too limiting.
I won't merge the proxies, they have segregated responsibilities and implementations.
The factory method in the FtpClient class didn't make sense neither.

I hope the code is acceptable as is. If not, I will let you modify it as you please.

springcomp:
@hgupta9 

I'm working with @LZ34 on this feature. Please, let me explain a bit more about the context of this pull request. We really love FluentFTP and would like to use it in our product, in replacement to the library we currently use. However, we need the FtpClient to support connecting through a proxy.

Many different kinds of proxy exist.

The most common one is an HTTP Proxy. However, there are many more customized protocols out there. For instance, if you check the FTP Proxy Settings dialog of FileZilla, you will find many different options.

FTP Proxy Settings

For one of our client, we needed to support the USER@HOST form of interaction with a proxy.

We figured there were two approaches to modify your library so that a proxy can be supported. One way to do that is to include support for the two supported proxies in your library. In that case, your suggestion to use a enumeration `ConnectionType` or `ProxyType` makes sense. However, this would hardcode the list of supported proxies. If one would like to add a third kind of interaction with a proxy, one would need to further modify the library.

The change that we are trying to make for this pull request is rather to make your library extensible with regards to which kind of proxy to use. Making the type of proxy a kind of _plugin_ if you like. That's why we are recommending using different concrete implementations of a simple base class and that's why there are two implementation in our pull request.

Getting back to the three suggestions, none of them allow us to make an extensible design. We would really love to be able to plug different proxy classes without modifying the library in the future. In this pull request, you have two different implementations, for two different forms of interaction with a proxy.

Perhaps, because an HTTP proxy is so common we could have it built in the library. And if you prefer, bring the less common USER@HOST form of interaction in a separate library. In that case, the library would be modified to make it extensible and support third-party plugins. We could make a simple interop library that defines the base class for third-party plugins. And bring an additional library for the USER@HOST proxy.

Please, let us know how we can help you move forward with this change.

You talked about ways to test this pull request. In our product, we have written a dummy proxy server that can be used to exercise the code. So that would be an addition to this project as well.

hgupta9:
Thanks for the detailed rationale. In response to your points:

> "If one would like to add a third kind of interaction with a proxy, one would need to further modify the library."

I don't understand why people simply couldn't take the source, add their changes and re-build the library. Its not closed source.

> Getting back to the three suggestions you made, none of them allow us to make an extensible design. We would really love to be able to plug different proxy classes without modifying the library in the future. In this pull request, you have two different implementations, for two different forms of interaction with a proxy.

Again, I don't understand but I accept. I'll merge this PR and think about refactoring later. 

Can the naming be improved? For example `FtpProxy` instead of `FtpClientProxy`? What do you suggest?

springcomp:
> I don't understand why people simply couldn't take the source, add their changes and re-build the library. Its not closed source.

I think that's the heart of the discussion. Yes, people can definitely modify the library themselves. After all, that's what we have currently done to help us move forward. However, our modifications are not part of the "standard" and official library. Therefore, we have the code for our modifications in one of our repo and we will need to maintain it.

We would much rather have the standard library merge our changes in the best possible way and modify our code to discard our modifications and use the next version of the library with the new features.

In making this pull request, it is not our intent to waste your time. I will happily change the naming to `FtpProxy` and push an updated pull request if you will. Please, let me know what you would plan to refactor and I will happily try and make the changes accordingly to make it easier for you to merge the pull request at a later time.

Does that work for you?

hgupta9:
I've merged the PR and cleaned it up in this [commit](https://github.com/hgupta9/FluentFTP/commit/30d22cb126b4c3eabedadccb38f96170c5fb726c). Can you test and see if it works fine for you?

hgupta9:
@springcomp and @L3Z4 - Thanks for your contribution. I appreciate your efforts in adding FTP proxy support to FluentFTP.

PR #17, opened 2016-12-06
Fixes for Proxy

elmar69:
Processed the Diff file from Zoltan666 to an Pull Request


hgupta9:
Thank you very much for the PR! I could not make sense of the diff files!

PR #18, opened 2016-12-08
LocalEndpoint can be null

elmar69:
I noticed that on Windows-Vista machine (Running on Mono Runtime) m_stream.LocalEndPoint migtht be null.

With the 2 changes the exceptions were eliminated. I did not see the reason for missing local endpoint data and there are much more references to be fixed. Perhaps it might be wise to create an dummy item to prevent Exceptions.


hgupta9:
Thank you for providing a PR! Thanks for the contribution.

PR #21, opened 2016-12-13
Add .NET Core Version of FluentFTP

hamu01:
Add .NET Core Version of FluentFTP,  there is a bug when upload file, sometimes it only upload partial document, I don't know why.

### DIFF
--- a/FluentFTP.Core/Extensions/GetChecksum.cs
+++ b/FluentFTP.Core/Extensions/GetChecksum.cs
@@ -1,0 +1,106 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace FluentFTP.Extensions {
+    /// <summary>
+    /// Retrieve checksum of file on the server
+    /// </summary>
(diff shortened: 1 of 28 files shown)


hgupta9:
Good job! I'll check out if the normal version is compiling fine and then merge it. Thanks for the work. However in the future I think its better we do this using compiler directives so we can maintain a single source code rather than 2 versions. I will try to merge it when I get time.

hamu01:
@hgupta9 Yes, I agree with you, I create a new version mainly because I don't want change too much. Expect a new version in nuget.

hgupta9:
New version in nuget? Are you publishing directly now??

hamu01:
I expect you publish a new version in nuget.
I don't publish any version to nuget, of course.

hgupta9:
Are you working on this yet?

hamu01:
No, the code in this feature still have problem

hgupta9:
Done it myself. Thanks for the idea though.

PR #24, opened 2016-12-22
Added Serializable attribute to defined Exceptions

RedDestroer:
All custom exceptions has to have [Serializable] attribute to make sure they can be serialized when go thru WCF communication for example, or if roaming between AppDomain's.


hgupta9:
Thanks!

PR #73, opened 2017-04-14
Sign assembly with strong key

amerkoleci:
In some scenario would be nice to have the assembly signed.


hgupta9:
Thank you for this! It has been requested for a long time...

Issue #76, opened 2017-04-25
Update to Async/Await Tasked-Based Asynchronous Programming Pattern


jblacker:
I know that this library was designed with the old style of asynchronous programming using the Begin & End pattern.  Is there currently any push to switch to the new `async`/`await` pattern?  

If so, I'd be happy to help out rewriting the asynchronous methods.  This library has helped me out a bunch and I'd like to contribute back if I could.

hgupta9:
Sure, your help would be greatly appreciated. If you don't have time to convert all the methods please feel free to submit a PR for whatever changes you have done, and I'll look and what you've done and repeat for the remaining methods..

jblacker:
I started working on this, but I've been leaving the current Begin & End methods.  Instead I'm creating new methods that end with`*Async()`.  I'm assuming you don't want this to be a breaking change right?

hgupta9:
> I started working on this, but I've been leaving the current Begin & End methods and wrapping them since they are already tested. 
Sounds okay. Perhaps later we can disable those/ make them private and see if the community is okay with that.

> All the new Task-based async methods will end with*Async(). I'm assuming you don't want this to be a breaking change right?
Yes, try to avoid breaking changes.

> Additionally, I've been updating the documentation within the code as I go through it. Would you rather me submit the updated comments as a separate pull request or keep it together with this?
Keep it together. Changing comments cannot cause any harm IMO :)

jblacker:
@hgupta9  Regarding making those private / disabling them.  I'd say the methods can probably be rewritten in a more "pure" asynchronous way sometime in the future.  The wrapping will work, but it's kind of a quick & dirty way of allowing this to follow the Task-based async pattern.  In the future these methods can probably be re-written async all the way down.

Here's my current plan.  I'm up to step 3 and will be submitting a pull request once I finish this step and test everything out.  

1.  Add overrides to the underlying stream sources (`FtpDataStream` & `FtpSocketStream`) to support the task-based async pattern.  As of .NET 4.5 virtual methods already exist on the `Stream` abstract class so they just needed to be added to those two `Stream` implementations
2.  Wrap all of the current Begin & End asynchronous methods with `Task.Factory.FromAsync()` to produce an awaitable Task that encapsulates those methods.
3.  Add awaitable Task methods to the UploadFile / DownloadFile (and multi-versions) 
[...]

PR #79, opened 2017-05-02
Adds Support for Task-based Asychronous Programming Pattern

jblacker:
This pull request contains parts 1 - 3 that I mentioned in my posted issue #76.  I have tested the functionality by uploading 1GB worth of files using both the `UploadFileAsync()` & `UploadFilesAsync()` methods.  Downloaded approx 500MB in test data using the `DownloadFileAsync()` method.   Performed file deletions and directory creations.  I do not have access to an FTP server that supports hashing, so that's untested.

Additionally, I updated the documentation throughout the FluentFTP project.

You'll need to update the nuspec and any scripts associated with nuget to property target .NET 2.0, .NET4.0, .NET4.5 - .NET4.6.1, and .NET CORE.  I added the Build Configurations, but I didn't touch any of the Nuget stuff.  I did not touch the versioning so nothing's been incremented.


hgupta9:
Fantastic work, Jordan. Thanks a lot for your contributions. Much appreciated :)

hgupta9:
I added your name into the credits :)

hgupta9:
Can you add your test methods into `Tests.cs` and submit that too?

hgupta9:
Do you think its a good idea to disable the Begin/End syntax for the .NET 4.5 & .NET core build with conditional compiling? Is there really any reason that the older syntax would be useful for newer projects?

jblacker:
@hgupta9 I tested it by adding it to an application I previously wrote using this library and modifying it to make it async.  However, if you'd like I can write some tests.

Regarding disabling the Begin/End syntax:  I wouldn't disable it yet, especially with conditional compiling for two reasons.
1)  I'm currently just wrapping those Begin & End methods so they are still required.
2)  I don't think we should issue a breaking change that abruptly. 

I'd recommend that we depreciate those methods and start re-writing the async methods to be "async all the way down" (and adding cancellation support).  This will require more work than using the wrapper methods since we'll need to write some new internal methods to issue commands on the socket/streams asynchronously, but still make sure everything is thread safe and ordered operations are still properly ordered.  It should improve pipelining & general performance though if we do get these written.  

In new projects I cannot really think of _why_ the older syntax would be used, but I'd be curious if anyone else in the community has a reason to use them.

hgupta9:
Good points. Oh and I didn't mean full blown tests. Just take like one or two methods (any method in Tests.cs) and create a copy of it with your async syntax. I can then run those "tests" on my PC to check roughly if everything looks okay.

jblacker:
@hgupta9  Ah, no problem.  I can do that.  I'll get it done a bit later today.

jblacker:
Doh, looks like I'm not going to get this done today.  I'll most likely have it by tomorrow though.

jblacker:
Submitted tests as pull request #80

hgupta9:
Thank you.

PR #100, opened 2017-05-15
Implementation of #81, #83, and #92

### LINKED ISSUE #81: Additional Logging Levels
Currently when you attach a `TraceListener` to `FtpTrace`, everything is issued at the DEBUG level.   Obviously, this is great for development and testing.  However, in production when an error occurs it would be nice to log the errors and ignore the tracing messages.  

I think the FtpTrace class would be greatly enhanced by simply adding levels so when a TraceListener is attached the logging can be filtered appropriately and errors can be handled by the appropriate people.  When using the low-level APIs this isn't a big deal, but this is especially important when a failure occurs using the high-level APIs.  For example, when using `UploadFile()` / `DownloadFile()`, there's no way to know what the failure if it returns false in production or in a release build.

Originally, I thought that maybe adding a "LastError" property to the `FtpClient` class would do the trick, but in a multi-threading environment this would be a nightmare.  What are your thoughts?

### LINKED ISSUE #83: Short Circuit Multi-File Upload/Download
What do you think about having the option to make the `UploadFiles()` and `DownloadFiles()` "transactional"?

There are situations where you'd want to transfer an entire directory or set of files, and a partial failure is not acceptable.  Since these methods swallow up exceptions they will still continue to upload/download until it completes.  In the cases that this is not acceptable I propose that an options flag is added to change the behavior if needed:

1. None (current behavior)
2. Stop Uploading/Downloading and return false immediately (do not upload remaining files)
3. Throw when an exception is encountered
4. Delete the files (or created directories) that uploaded/downloaded successfully

The deletion option would be able to be combined with either option 2 or 3.  This functionality along with the logging enhancement I proposed in #81 would really help handling problems in a production environment.  Maybe a boolean flag can also be added to `UploadFile()`/`DownloadFile()` to cause the method to throw if an unresolvable error is encountered if requested.

### LINKED ISSUE #92: Automatically verify checksum after upload/download
UploadFile() and DownloadFile() should provide an arg to verify the checksum after upload/download is complete. Failure of hash match should:

1. Throw an exception immediately
2. Retry a fixed number of times before giving up and throwing an exception
3. Keep the invalid file and return false
4. Delete the invalid file and return false

jblacker:
- Updated implmentation of the `FtpTrace` class to use a `TraceSource` instead of directly writing to `TraceListeners`.
- Implemented "Log Levels" on the `FtpTrace` class that translate to builtin levels that can be issued via a the `TraceSource`.  Resolves #81 
- Implemented user-defined error handling for Multi-File Upload / Download methods.  Resolves #83 
- Implemented verification and verification failure handing for File based Uploads & Downloads (not implemented on uploads/downloads that function directly on a stream object).  This needs some additional testing because I do not have access to an FTP server that has supports HASH functionality currently.  Should resolve #92 
- Updated XML documentation for all changes
- Updated XML documentation to correct spelling & grammar, as as well as creating some links to correctly reference other classes/methods.
- Depreciated the following methods:  `FtpListParser.LogWarn()`, `FtpListParser.LogDebug()`, `FtpTrace.Write(string msg)`, `FtpTrace.WriteLine(string msg/object msg)`
- Updated Multi-File Upload/Download method signatures to use `IEnumerable<string>` instead of `string[]` or `List<string>` to eliminate unnecessary overloads since these all implement `IEnumerable<T>`
- Corrected bug introduced in commit 20caf4b2ae13608fa0525f66647f28418a23f5c8

**Warning**: This introduces breaking changes to the API


jblacker:
Don't pull this yet.  I just realized I needed to add something.

jblacker:
Ok, now it's ready

hgupta9:
Wow. That's an amazing amount of work. I'll review it tomorrow and merge at the earliest. Thank you!

hgupta9:
And what was the bug introduced in commit https://github.com/hgupta9/FluentFTP/commit/20caf4b2ae13608fa0525f66647f28418a23f5c8?

hgupta9:
And what are the breaking changes in the API? You mean the extra args to the Upload/Download methods?

jblacker:
@hgupta9  `BeginFileExists()` had infinite recursion since it was calling `return BeginFileExists()`, but the overload it previously used was removed.

Yes the extra args to the upload/download method is the breaking change.

hgupta9:
Great work so far. Thanks for the heavy investment of your time and effort. Its appreciated :)

hgupta9:
Can you help me update the FAQ? Is this correct? Any more relevant questions regarding the new tracing method you built?

**How do I trace FTP commands for debugging?**
Do this ONCE at program startup (since its static it takes effect for all current FtpClient instances.)

   FtpTrace.AddListener(new ConsoleTraceListener());

**How do I log critical errors to a file?**
Do this ONCE at program startup (since its static it takes effect for all current FtpClient instances.)

   FtpTrace.AddListener(new TextWriterTraceListener("log_file.txt"));

hgupta9:
And I've simplified the hashing API, by omitting the `FtpVerify.Checksum` option. So users don't have to do `FtpVerify.Checksum | FtpVerify.Delete` instead you can just do `FtpVerify.Delete`. As far as I can tell I've not broken anything but can you check? https://github.com/hgupta9/FluentFTP/commit/b3fbd8635ef9648cfc901f001629e793292db857

hgupta9:
Readme improved to include error handling/hashing functionality. New [FAQ entry](https://github.com/hgupta9/FluentFTP#faq_verifyhash) for hashing.

jblacker:
@hgupta9 You actually wouldn't need to do `FtpVerify.Checksum | FtpVerify.Delete`.  Only having `FtpVerify.Delete` would be sufficient to make it do verification with deletion.  Having only `FtpVerify.Checksum` would just cause the function to return false if there was a verification failure.  I could see this functionality being useful if the end-user doesn't want any specific action taken by the library, but I can also see how it might seem a bit confusing.  Essentially `FtpVerify.Checksum | FtpVerify.Delete == FtpVerify.Delete` in terms of behavior (not actual value).  The reason being is that I do verification as long as `verifyOptions != FtpVerify.None`.  However, with that being said I can see how this could be confusing to end-users.  With that being said I looked at the code and no, that should not have broken anything.

hgupta9:
I could still keep the Checksum option and call it ReturnFalse or something, indicating that the user can take any action he wants.

Can you help me update the FAQ regarding tracing since I don't know how to use the new functionality you built? And how does one use the FtpTraceLevel setting?

**How do I trace FTP commands for debugging?**
Do this at program startup (since its static it takes effect for all current FtpClient instances.)
```cs
FtpTrace.FlushOnWrite = true;
FtpTrace.AddListener(new ConsoleTraceListener());
```
**How do I log critical errors to a file?**
Do this at program startup (since its static it takes effect for all current FtpClient instances.)
```cs
FtpTrace.FlushOnWrite = true;
FtpTrace.AddListener(new TextWriterTraceListener("log_file.txt"));
```

jblacker:
Personally, I'd prefer it as `Checksum` because `ReturnFalse` would seem more like this flag is *required* for it to return false upon failed verification, which isn't currently true as any set handling assumes Checksum/ReturnFalse.  If it would make more sense to make that a requirement I can update the implmeentation to prevent it from assuming a return false upon failed verification without the flag explicitly set.  However, as long as it's documented it doesn't matter as much.  Since you're the owner of the library I'll let you make that call.

Regarding updating the FAQ, you are correct that should enable it & yes it needs to be done at application start..  However, hold off on documentation it because although it was working when I submitted my PR it isn't functioning as expected on my system.  I'm going to make sure there's no bugs with using the `TraceSource` and setting up a few configurations to make sure it's fully functional.  I'll update the FAQ with my demo configs once it works to demonstrate how it's configured.

hgupta9:
For the flag, how about these other names :
- FtpVerify.Manual
- FtpVerify.OnlyCheck
- FtpVerify.OnlyChecksum
- FtpVerify.OnlyReturn

It we have proper docs saying that nothing will be done based on the checksum, and the method will return false upon failure.

Please update when the tracing is functioning. I will then push changes to nuget.

jblacker:
I have it fixed on my end, but I'm writing the documentation now.  I would avoid the word "Only" because it's a flag so it can be combined with others.  I'd recommend "Check" and then in the documentation say that using "Retry", "Delete", or "Throw" assumes "Check"

hgupta9:
Submit the PR whenever you want to, I'll make the change after. And if you can get .NET core support its beneficial. Adding an extra dependency won't be a problem. Anyways I'm using the latest .NET standard (1.6) so all modules are available.

jblacker:
@hgupta9 I was waiting on your call regarding the the additional enum value for `FtpVerify` before I submit the second push.

hgupta9:
@jblacker I prefer `FtpVerify.OnlyChecksum` as it indicates its a special use case and you don't need to consider it for normal usage (similar to `FtpExists.NoCheck`). I think calling it `Checksum` implies that you NEED that property to enable hashing and then you have to explain that `Reply = Checksum | Reply` and all that stuff which brings in unnecessary complication. And oh I tried to simplify the usage as much as possible. See the [FAQ entry](https://github.com/hgupta9/FluentFTP#faq_verifyhash) - I've filled in "all" the possible combinations to help users set the flags correctly.

jblacker:
Sounds good.

Issue #115, opened 2017-05-31
FtpClient.DownloadFile throws when downloading to current directory


ibsenrune:
Steps to reproduce:

```
var client = new FtpClient("some.ftp.server");
FtpListItem fileItem = /* get some file item */

//Download file to local current directory, maintaining the file name
client.DownloadFile(fileItem.Name, fileItem.FullName, true, FtpVerify.None);
```

Expected result: the file is downloaded to the local current working directory of the application.
Actual result:

```
Unhandled Exception: FluentFTP.FtpException: Error while crated directories. See InnerException for more info. ---> System.ArgumentException: Path cannot be the empty string or all whitespace.
   at System.IO.Directory.CreateDirectory(String path)
   at FluentFTP.FtpClient.DownloadFileToFile(String localPath, String remotePath, Boolean overwrite, FtpVerify verifyOptions)
   --- End of inner exception stack trace ---
   at FluentFTP.FtpClient.DownloadFileToFile(String localPath, String remotePath, Boolean overwrite, FtpVerify verifyOptions)
   at FluentFTP.FtpClient.DownloadFile(String localPath, String remotePath, Boolean overwrite, FtpVerify verifyOptions)
   at FtpTest.Program.Main(String[] args)
```

The problem arises because `fileItem.Name` is something like `"myfile.txt"`. That string is eventually passed as `localPath` [here](https://github.com/hgupta9/FluentFTP/blob/master/FluentFTP/FtpClient.cs#L3312). The `dirPath` becomes `""`, `Directory.Exists(dirPath)` returns `false` in the next line and the empty string is eventually passed to `Directory.CreateDirectory` [here](https://github.com/hgupta9/FluentFTP/blob/master/FluentFTP/FtpClient.cs#L3314), which causes the exception to be thrown.

I am happy to submit a pull request to fix this if you would like me to.

hgupta9:
I think you misunderstood the purpose of the function. DownloadFile() is only to download a file from the _FTP server_ to the _local file system_. If you want to move a file keeping it on the FTP server please use MoveFile(). There is no CopyFile() right now, but you can make it using DownloadFile() and then UploadFile().

ibsenrune:
That's exactly what I want: in a console application, I want to download the file to the local directory from which the console application is running.

hgupta9:
Yes, it should be an absolute path. I have not tested it with relative paths.

hgupta9:
[...]

PR #117, opened 2017-06-02
Check dirPath is not empty string

ibsenrune:
If `localPath` is just a filename, say 'file.txt', then the directory part returned by `Path.GetDirectoryName(localPath)` will be the empty string. Passing that to `Directory.Exists(dirPath)` will cause an exception.

This change ensures that we never pass the empty string to `Directory.Exists(dirPath)`. We can safely make this change, because if `dirPath` is the empty string, then we are downloading to the current directory, which definitely exists, so there is no need to create a new directory.

This fixes issue #115.


hgupta9:
Thank you!

PR #126, opened 2017-06-22
support .Net3.5

orisox:
hi,
i want  to use FluentFTP in Unity3D, and Unity3D only support dot net 3.5, so i make a patch to build dot net 3.5 version


hgupta9:
Thanks! I'll merge this and fix any issues.

Issue #122, opened 2017-06-14
Timed out trying to read data from the socket stream!


JWcodeDev:
**NOTE FROM ADMIN:**

This should be fixed in 29.0.0 and onwards.
But if you still have this issue, please add your post below instead of creating duplicate issues!

Thanks.

----

I am not able to upload a file to a FTPS site.  I keep receiving the following error.  If the file exists it will delete the file from the FTP but it will not upload the new file.

{"Timed out trying to read data from the socket stream!"}

Here is my code:

            FtpClient client = new FtpClient("ftp", "user", "pwd");

            client.DataConnectionType = FtpDataConnectionType.PASV;
            client.EncryptionMode = FtpEncryptionMode.Explicit;
            client.SslProtocols = System.Security.Authentication.SslProtocols.Tls11;
            client.ValidateCertificate += new FtpSslValidation(OnValidateCertificate);
            client.Encoding = System.Text.Encoding.ASCII;

            client.UploadFile(@"C:\Test\Test.txt", "/archive/Test.txt", FtpExists.Overwrite, true, FtpVerify.Retry);

             client.Disconnect();

It errors out on the client.UploadFile line.

Seems it is in the GetRequestStream().  Not sure how to over come this error.

Thanks,

JW

hgupta9:
Please send full logs in order for us to debug. Read the FAQ on how to print or save logs to a file.

JWcodeDev:
Thanks for the Response.

Here are the logs:

# UploadFile("C:\Test\Test.txt", "/Test.txt", Overwrite, True, None)

# FileExists("/Test.txt")

# Connect()
Status:   Connecting to ***:21
Response: 220 Welcome To  Solutions.  All traffic is monitored and scanned.
Response: 220-Microsoft FTP Service
Command:  AUTH TLS
Response: 234 AUTH command ok. Expecting TLS Negotiation.
Status:   FTPS Authentication Successful
Status:   Time to activate encryption: 0h 0m 0s.  Total Seconds: 0.3145793.
Command:  USER ***
Response: 331 Password required
Command:  PASSS ***
Response: 230 User logged in.
Command:  PBSZ 0
Response: 200 PBSZ command successful.
Command:  PROT P
Response: 200 PROT command successful.
Command:  FEAT
Response: 211 END
Response: 211-Extended features supported:
Response: LANG EN*
Response: UTF8
Response: AUTH TLS;TLS-C;SSL;TLS-P;
Response: PBSZ
[...]

PR #136, opened 2017-07-11
High level API - ASCII mode transfer fixes

rharrisxtheta:
We found two issues with the high level transfer functions.   This pull request fixes both.

1.  The UploadDataType and DownloadDataType properties are ignored in most cases.   So, if you set the DownloadDataType to ASCII - the file will still download in Binary unless the server aborts the connection.  (then the retry will set the correct datatype).

**Fix:** Honor the Upload/DownloadDataType properties in all cases.

2. Unknown size ASCII files won't transfer.  Some of FTP servers won't allow you to download a 0 byte file in ASCII mode.  There's code in the high level methods to work around this issue - but it also blocks transfer of 'unknown size' files in ASCII mode.   This creates a problem since some systems don't support the 'SIZE' subcommand (AS400, possibly S390), but require that you download files in ASCII mode.   

**Fix:**  Force user to use BINARY mode for 'known 0 byte' files, but allow them to use ASCII mode if the server doesn't support the 'SIZE' subcommand.

### DIFF
--- a/FluentFTP/Client/FtpClient_HighLevel.cs
+++ b/FluentFTP/Client/FtpClient_HighLevel.cs
@@ -1598,9 +1598,10 @@ namespace FluentFTP {
 
 			try {
 
-				// exit if file length not available
-				downStream = OpenRead(remotePath, DownloadDataType);
-				long fileLen = downStream.Length;
+                // exit if file length == 0
+			    long fileLen = GetFileSize(remotePath);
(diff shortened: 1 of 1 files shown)


rharrisxtheta:
This may relate to #122

hgupta9:
Thank you for this!

hgupta9:
Also should fix #137

Issue #137, opened 2017-07-12
GetFileSize - SIZE not allowed in ASCII mode


daantobk:
The FTP server returns 'SIZE not allowed in ASCII mode'. But I cannot change the FtpDataType to Binary, since this is protected code.
Please help??

hgupta9:
Try the latest. There are some fixes for ASCII

https://www.nuget.org/packages/FluentFTP/

daantobk:
I tried, but still not working.
Response is still:
550 SIZE not allowed in ASCII mode

2 possible fixes: 
- set to binary mode in function GetFileSize
or
- make make function SetDataType(FtpDataType type) public

hgupta9:
Try the latest. There are some fixes for this by rharrisxtheta.

https://www.nuget.org/packages/FluentFTP/

hgupta9:
And which server are you using?

daantobk:
Yes it works for version 17.5.8.
Thank you very much!

I don't know which server our client is using, because they don't send the version in the FTP session

hgupta9:
Glad it works for you. Special thanks to @rharrisxtheta for the fix.

Tekkion:
using 19.1.2 and it seems the error reappeared

Blackleones:
same error on 26.0.1

robinrodricks:
@Blackleones please check with this version?

https://we.tl/t-MUcikEcMsX

If you still get the error please send the logs so we can check (FAQ has steps how to get logs)

robinrodricks:
@Tekkion @daantobk @Blackleones  Can one of you guys check the version I posted? I tried to fix this...

robinrodricks:
Some fixes released in 26.0.2, please check:

https://www.nuget.org/packages/FluentFTP/26.0.2

If there are still issues, message here I'll reopen this thread.

Blackleones:
@robinrodricks problem still in 26.0.2 

Status:   Connecting to ***
Response: 220 Aruba FTP2S3 gateway 1.0.1 ready.
Command:  USER ***
Response: 331 Username ok, send password.
Command:  PASS ***
Response: 230 Login successful.
Command:  FEAT
Response: 211 End FEAT.
Response: 211-Features supported:
Response: AUTH SSL
Response: AUTH TLS
Response: EPRT
Response: EPSV
Response: MDTM
Response: MFMT
[...]

PR #138, opened 2017-07-12
Temporarily switch into BINARY mode before running 'SIZE' command.

rharrisxtheta:
This fixes some FTP servers that don't work correctly unless you do a SIZE in binary...

Possible fix for #137


hgupta9:
Thank you for this! Much appreciated..

PR #139, opened 2017-07-12
Misc. listing fixes

rharrisxtheta:
Add a listing fix for on the AS400, and make the listing methods automatically 'retry' if it turns out we picked the wrong date format.


hgupta9:
Thank you for this! Much appreciated..

rharrisxtheta:
Quite welcome.  If we come across other issues we'll get you some more fixes.

PR #140, opened 2017-07-14
Increase performance of GetListing method

sierrodc:
GetListing is based on FtpSocketStream.ReadLine to read the list of files returned by the ftp server.
ReadLine method is used across the library to read **only** the bytes required till the end of the line, so it is basically a FtpSocketStream.read of 1 char at a time.
Reading GetListing result char by char is a huge performance issue (considering also the AsyncWaitHandle.WaitOne method used for each Read).

I've added a method to FtpSocketStream class, ReadAllLines, which is able to read n characters in a single call (at the moment 100) and return the list of all lines retrieved by the server.

I've tested it with a big FTP folder (4846 files) with following results:
 - current implementation: 3.6-3.8 seconds to get the full list
 - modified implementation: 0.48-0.55 seconds

so, ~600% faster

Question: are there any unit tests?

### DIFF
--- a/FluentFTP/Client/FtpClient_Listing.cs
+++ b/FluentFTP/Client/FtpClient_Listing.cs
@@ -389,12 +389,9 @@ namespace FluentFTP {
 				using (FtpDataStream stream = OpenDataStream(listcmd, 0)) {
 					try {
 						FtpTrace.WriteLine(FtpTraceLevel.Verbose, "+---------------------------------------+");
-						while ((buf = stream.ReadLine(Encoding)) != null) {
-							if (buf.Length > 0) {
-								rawlisting.Add(buf);
-								FtpTrace.WriteLine(FtpTraceLevel.Verbose, "Listing:  " + buf);
-							}
-						}
+
+                        rawlisting = stream.ReadAllLines(Encoding).Where(l => l.Length > 0).ToList();
+
 						FtpTrace.WriteLine(FtpTraceLevel.Verbose, "-----------------------------------------");
 					} finally {
 						stream.Close();
--- a/FluentFTP/Client/FtpClient_Listing.cs
+++ b/FluentFTP/Client/FtpClient_Listing.cs
@@ -6,7 +6,6 @@ using System.Text.RegularExpressions;
 using System.Reflection;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Linq;
 using System.Security.Cryptography.X509Certificates;
 using System.Globalization;
 using System.Security.Authentication;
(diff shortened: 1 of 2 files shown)


hgupta9:
Good work and interesting implementation.
No unit tests atm but feel free to create some based on the code in the Tests project.

What I don't understand is how you know beforehand how many bytes exist in order to feed that value into your ReadAllLines method? Because if you read more than what the server sends I think the connection hangs waiting for bytes that are never gonna come?

And whats with 100 bytes? Is that chosen arbitrarily and why not 1000 for instance to result in larger speedups? Whats the reasoning behind the selection of that constant.

And whats the .Skip .Take? Is that LINQ? Is there a simpler implementation because LINQ is slower than straight code and may not play well with all the platforms we support.

sierrodc:
Hi,

 I've modified the code to get rid of System.Linq namespace.

About the size of the array (100) it is just a "feeling" and a "nice number". I did some tests changing the value and I cannot notice any change using a size of 50 or 150 bytes or even 10000. I  just noticed that using too small buffers (5 or 10 or 20 bytes) decreases performance.

About the amount of bytes received from server, you can read this documentation page: [NetworkStream.Read Method (Byte[], Int32, Int32)](https://msdn.microsoft.com/en-us/library/system.net.sockets.networkstream.read%28v=vs.110%29.aspx?f=255&MSPPError=-2147217396).

According to what it is written in Remarks: 

> This method reads data into the buffer parameter and returns the number of bytes successfully read. If no data is available for reading, the Read method returns 0. The Read operation reads as much data as is available, up to the number of bytes specified by the size parameter. If the remote host shuts down the connection, and all available data has been received, the Read method completes immediately and return zero bytes.

So, if the server replies with 152 bytes and you can create a buffer of 10000, the method will fill first 152 bytes of the array. No exception is thrown (at least if everything is fine with the connection).

hgupta9:
Looking good. Thanks for the changes. Just before I merge this I want to confirm: what are the possible use cases in which this can go wrong? Would it be necessary to add a boolean "UseLegacyListing" or something which uses the older line-by-line mode?

sierrodc:
Hello, this is a good question.

I've tested the modified assembly with 3-4 different FTP server and it worked fine.

If you want to add a "UseLegacyListing" parameter/config to be removed one day in the future if everything will work fine for me it's fine. Let me know if you want I add this kind of setting (I like more a property in FtpClient class).

Unfortunately I don't know exactly how to create a "integration" testing library to validate this changes. 
Maybe something like [MockFptServer](http://mockftpserver.sourceforge.net/fakeftpserver-getting-started.html) can be useful.

hgupta9:
Thanks for testing it on various servers. Yes it would be good if you can add the property on your side. Add it into the same file (`FtpClient_Listing.cs`) just below `RecursiveList`. Call it `BulkListing` with default value of true. Thanks.

hgupta9:
And please add another int property `BulkListingLength` with default of 128 to be used instead of the 100 constant.

sierrodc:
Hi, properties added.

Forgot to mention that my real issue was not the speed (I didn't care of 1-4 seconds) but the cpu trottling at 100% during file listing.

Thanks.

hgupta9:
Thank you for this!

Issue #145, opened 2017-07-25
AccessViolationException in FtpSocketStream when connecting


vishalpatel-te:
![image](https://user-images.githubusercontent.com/17013023/28590396-ed100c9c-714f-11e7-96d6-cc16e3fcca8a.png)

On one of our dev environment we are getting error that forces the process to exist.
I have tried to set the FTPTrace flags to false with no affect.
            FtpTrace.FlushOnWrite = false;
            FtpTrace.LogFunctions = false;
            FtpTrace.LogIP = false;
            FtpTrace.LogPassword = false;
            FtpTrace.LogUserName = false;
            FtpTrace.LogPrefix = false;

Also tried, 
            FtpTrace.AddListener(new TextWriterTraceListener("log_file.txt")
            {
                Filter = new EventTypeFilter(SourceLevels.Error)
            });

But getting same error all the time. Any help regarding this would be appreciated.

Same code gives no error on other developer's machine.

hgupta9:
Disabling the tracing will not solve the error! Its like disabling the logging system! The error is caused by the socket stream, not by the tracing/logging system! I don't know what the error is. Perhaps you should remove all the lines you showed above and enable logging on console (see the FAQ).. and then see at what point the error occurs. More than that there is nothing I can do from the armchair. Search google for AccessViolationException in Stream... FtpSocketStream is the main socket handling class, it has NO unsafe code so I'm not sure what is causing the AccessViolationException.. also the code in Connect() is very simple.. you can see for yourself in the source code. Maybe try running the debug version, it will give line numbers of the error as well...

https://we.tl/MQBg8vxViq

vishalpatel-te:
I am currently on 17.5.8 versrion before. Updating to 17.6.0 did not resolve the issue either. But the one you provided in link above did. So something is getting missed between DEBUG and RELEASE versions. 
May be .Net compiler is optimizing something it should not.
Will use the debug dll for now.

albertop:
Hi, all:

I am getting the exact exception while doing .Connect() while using the version 17.6.1 from nuget.

My app targets net40.  Could be the case that the app downloaded from Nuget is not for NET4?

Would you help me solve this?

Unhandled Exception: System.AccessViolationException: Attempted to read or write
 protected memory. This is often an indication that other memory is corrupt.
   at FluentFTP.FtpTrace.Write(FtpTraceLevel eventType, String message)
   at FluentFTP.FtpClient.Connect()
[...]

PR #164, opened 2017-08-15
Fix  for #145

ltines:
Marks the TraceSource as volatile thus avoiding compiler optimizations

### DIFF
--- a/FluentFTP/Helpers/FtpTrace.cs
+++ b/FluentFTP/Helpers/FtpTrace.cs
@@ -10,7 +10,7 @@ namespace FluentFTP {
     public static class FtpTrace {
 
 #if !CORE
-		private static readonly TraceSource m_traceSource = new TraceSource("FluentFTP") {
+		private static volatile TraceSource m_traceSource = new TraceSource("FluentFTP") {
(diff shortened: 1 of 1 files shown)


ltines:
Hi, the fact is it *is* accessed by multiple threads by being static (i.e. all instances of ```FluentFTP``` are sharing the same ```FtpTrace```). And I am not sure how ```TraceSource``` is thread-safe.

I went mostly with
> Fields that are declared volatile are not subject to compiler optimizations that assume access by a single thread

from docs because my test showed that debug builds are not experiencing #145 . So I assumed it will be caused by optimization done to ```m_traceSource```

andersonpimentel:
> So I assumed it will be caused by optimization done to `m_traceSource`

That's a long shot. More likely to be related to `FtpTrace.Write` method.

hgupta9:
@ltines Have you made a release build to check if it actually fixes #145??

ltines:
@hgupta9 yes i did. Has been deployed for few days. i do agree with @andersonpimentel that is a long shot. I might split it to two PRs; one for volatile and one for the switch (so people can turn it off until its properly fixed).

As a sidenote; for me it crashes only on Windows Server 2008 and only Release build

PR #165, opened 2017-08-16
Added a Interface for the FtpClient

Kris0:
Simply, just a Interface so we can write Unit Tests for our projects.


hgupta9:
Good, but I'm not willing to maintain the comments in this interface as well as the interface methods. Delete all comments.

Kris0:
Okay, done.

hgupta9:
Thanks!

Issue #167, opened 2017-08-23
Server Reply Not Read on FtpDataStream Dispose


Lukazoid:
This looks like it is related to issue #53 but occurs when running on the full .NET framework.

I first noticed this when attempting multiple uploads on one `FtpClient`, on some subsequent uploads I noticed the following trace messages:

> Command: TYPE I
> Response: 226 Transfer complete.

This looks like the server reply is not being read when an `FtpDataStream` is disposed.

When disposing of an `FtpDataStream` the call flow currently looks like:
> FtpSocketStream.Dispose() (due to shadowing IDisposable.Dispose()) -> FtpSocketStream.Close()

This means that `FtpDataStream.Close()` never gets called and thus `ControlConnection.CloseDataStream` is never called and no server reply read.

hgupta9:
Thank you for this notice. Yes I know, but the person who informed me before refused to help fix it. Thank you for fixing it!

hgupta9:
Published to nuget - https://www.nuget.org/packages/FluentFTP/18.0.1

canhantas:
I am still experiencing this exact problem after updating to 18.0.1

Lukazoid:
Is that using the `Upload`/`Download` methods? I've not used those so didn't realise this workaround existed in those.

I'll do some further testing today but we may be able to remove those blocks of code as the `FtpDataStream` should now handle reading the final reply itself.

PR #168, opened 2017-08-23
Dispose of FtpDataStream Now Reads Control Reply

Lukazoid:
This resolves issue #167.

The change here was to move all of the clean-up functionality to the `Dispose(bool)` overrides and simply use the `Close()` methods to call `Dispose()`.

As `FtpDataStream.Dispose(bool)` handles the reply from the control connection, we store this in a field so it may be returned from the shadowing `FtpDataStream.Close()` method.

### DIFF
--- a/FluentFTP/Stream/FtpDataStream.cs
+++ b/FluentFTP/Stream/FtpDataStream.cs
@@ -139,28 +139,55 @@ namespace FluentFTP {
 			m_position = pos;
 		}
 
+		private FtpReply m_closedDataStreamReply;
+
 		/// <summary>
 		/// Disconnects (if necessary) and releases associated resources
 		/// </summary>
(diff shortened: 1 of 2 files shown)


hgupta9:
Thank you for this! I have been unable to fix this for a while...

hgupta9:
This was a bad idea. Reverting.

hgupta9:
Bad idea. We should not change the API since lots of projects have already followed this pattern: OpenRead() and GetReply(). If you call GetReply() internally, there will be an extra call to GetReply() and it will hang on the 2nd call. All existing projects will break.

Lukazoid:
Is this not something which could be kept perhaps for the next major release and marked as a breaking change in the release notes?

If people are manually calling things such as `GetReply` it indicates the API surface is too large and should be reduced.

PR #170, opened 2017-08-25
Removed the fixes from high-level functions which read the final reply

Lukazoid:
An additional fix for issue #167.

This is no longer necessary now that FtpDataStream correctly handles the final reply itself.

### DIFF
--- a/FluentFTP/Client/FtpClient_HighLevel.cs
+++ b/FluentFTP/Client/FtpClient_HighLevel.cs
@@ -1137,13 +1137,7 @@ namespace FluentFTP {
 
 				// disconnect FTP stream before exiting
 				upStream.Dispose();
-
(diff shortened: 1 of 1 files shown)


Lukazoid:
I have tested this with both `EnableThreadSafeDataConnections = true` and `EnableThreadSafeDataConnections = false`.

I have tested the following functions from FtpClient: `OpenWrite`, `OpenRead`, `UploadFile` and `DownloadFile`, `UploadFiles` and `DownloadFiles`.

hgupta9:
Bad idea. We should not change the API since lots of projects have already followed this pattern: OpenRead() and GetReply(). If you call GetReply() internally, there will be an extra call to GetReply() and it will hang on the 2nd call. All existing projects will break.

Lukazoid:
So this issue will never be fixed because people already have hacks to work around it?

PR #175, opened 2017-08-28
Fix file upload EOS bug.

### LINKED ISSUE #174: FTPS file upload hangs for buffered Streams
Despite the FTP client API accepting a generic Stream as an argument, it assumes that the given Stream is seekable and that `fileData.Length` is equal to the length of the file.

In some cases, this is not true. While performing an asynchronous copy operation, for example, the provider of the Stream may not be aware of the total size of data being written into it; thus the Stream may have a fixed buffer size as its length rather than contain the full contents of the file (this is also wildly unreasonable for large file sizes).

Even worse, the current implementation ignores a very important aspect of the `Stream` API; i.e. that a zero return value from `Read` _always_ indicates end-of-stream (EOS). Thus, it is an error, no matter the circumstances, to continue calling `Read` after EOS has been reached. If the Stream has been disposed, it will throw an exception. If not, it will just keep reading forever.

The simple fix is to break after reaching EOS. But it should be considered as a future implementation improvement to remove the dependence on `Stream.Length` entirely.

bgroenks96:
Per the .NET Stream API, Read returns zero ONLY on end-of-stream.

Thus, the file upload should not continue after the inner read/write
loop terminates without error.

Stream.Length should not be relied upon for this purpose, as many
implementations of Stream may be buffered, and there is no guarantee
that the length of the Stream is the size of the file.

Resolves #174

### DIFF
--- a/FluentFTP/Client/FtpClient_HighLevel.cs
+++ b/FluentFTP/Client/FtpClient_HighLevel.cs
@@ -1077,6 +1077,8 @@ namespace FluentFTP {
 								offset += readBytes;
 							}
 
+							// zero return value (with no Exception) indicates EOS; so we should terminate the outer loop here
+							break;
 						} catch (IOException ex) {
 
 							// resume if server disconnected midway, or throw if there is an exception doing that as well
@@ -1117,7 +1119,9 @@ namespace FluentFTP {
 								}
 							}
 
-						} catch (IOException ex) {
+							// zero return value (with no Exception) indicates EOS; so we should terminate the outer loop here
+							break;
+                        } catch (IOException ex) {
 
 							// resume if server disconnected midway, or throw if there is an exception doing that as well
 							if (!ResumeUpload(remotePath, ref upStream, offset, ex)) {
@@ -1231,7 +1235,10 @@ namespace FluentFTP {
 								await upStream.FlushAsync(token);
 								offset += readBytes;
 							}
-						} catch (IOException ex) {
+
+							// zero return value (with no Exception) indicates EOS; so we should terminate the outer loop here
+							break;
+                        } catch (IOException ex) {
 
 							// resume if server disconnected midway, or throw if there is an exception doing that as well
 							if (!ResumeUpload(remotePath, ref upStream, offset, ex)) {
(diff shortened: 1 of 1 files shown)


artiomchi:
Good PR, but could you make sure your code uses the same whitespace character as the project? You're using spaces, and the project uses tabs, which makes reading the diff a big pain :)

In reality, this PR just added 4 `break;` calls, but it looks like the change was MUCH more extensive

artiomchi:
Could you apply the same fix to the `DownloadInternal` and `DownloadInternalAsync` methods in the same way? They have the same problem (with a poor attempt to overcome it).

I'm just testing a private build of the library where I removed the `readToEnd` variable entirely, and replaced it with `break` statements, same as in your code. Hoping it'll fix the issue we're seeing.

I'd create a PR of my own, but I just did the same thing you did, so want to give proper credit to you :)

bgroenks96:
@artiomchi Absolutely! I'll make the changes as soon as I am able.

bgroenks96:
@artiomchi I cleaned up the commit to remove some of the whitespace changes you referred to. I hope that makes it clearer.

Unfortunately, after further investigation, I don't think adding the unconditional `break` to file _downloads_ is a good idea (and also isn't likely the source of your problem).

Unlike the upload case, we _know_ the expected file length as reported by the server. The `readToEnd` flag correctly covers the case under which me must rely on the EOS signal to finish the download.

That being said, it's still an error to continue to read from the stream after it returns zero. I have modified the download methods to throw an exception and attempt to resume the download in this case. I would be interested to see if that solves your issue.

Best of luck!

artiomchi:
@bgroenks96 ah, I see, that's a great observation! I'll apply this patch instead of mine then. Thanks for that!

Unfortunately, I learned that a lot of the FTP servers that we have to connect to are pretty bad. To the point that `AutoActive` and `AutoPassive` modes don't work (server and client negotiate to a newer transfer mode, that the server reports as being supported, but it really doesn't), so I have to force it to use `PASV` and `PORT` instead. At this point I wouldn't be surprised if they sent the wrong file size as well :)

hgupta9:
Thank you very much @artiomchi and @bgroenks96 for this fix! Much appreciated!

PR #190, opened 2017-09-16
Migrating to a single solution and CSProj 2017 format

### LINKED ISSUE #185: UWP System.IO.FileNotFoundException
When trying to connect in a UWP app, the following error occurs: 

> "Could not load file or assembly 'System.Console, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'"

Code:
```
FluentFTP.FtpTrace.EnableTracing = false;
FluentFTP.FtpClient conn = new FluentFTP.FtpClient();
conn.Host = "{IP address}";
conn.Connect();
```

Exception throws at "conn.Connect();"

Version: NuGet 18.0.2

artiomchi:
Looking through the project, I noticed that you have three different solutions and three different csproj files. Besides making it less manageable, it also adds inconsistencies, and I've spotted some issues too. 

Moving to the new project format introduces a single solution/project that covers ALL the supported frameworks that you use, greatly simplifies configuration and management, removing clutter and keeping all the configuration in a single place. This also removes a LOT of potential for error.

Furthermore, makes development a lot easier, since you'll now be able to use/test all frameworks in one go, and building / packing the nuget package is as simple as building the project in release mode.

## Summary of the changes:

**Fixes:**
- The test project mapped the ReleaseNET4 configuration to the Release config. So it was never build against .NET 4.0 ([ref](https://github.com/hgupta9/FluentFTP/blob/974e4e2866070a75e4f705488ac50729eb0c83a0/FluentFTP_NET_VS2012.sln#L68))
- `FtpExtensionAttribute.cs` was completely forgotten and not referenced in the projects. It was also irrelevant, since the class from inside was also present in `Utils\NET2Compatibility.cs`
- `FluentFtp.csproj.user` seems to have been erroneously committed. I've removed it, and added it to `.gitignore`
- Net standard compiled code never included the .xml docs for the library
- The project is using custom compile time constants for each framework - I've switched them with pre-built ones (e.g. NETFX35 becomes NET35). I left the NETFX and CORE ones for wide matching rules to prevent making a massive revamp (although they might be not a great practice)
- Fixes #185, which is a missed reference to System.Console for .NET core projects

**Benefits:**
- I added a .editorconfig file to ensure that contributors use the same whitespace character as the rest of the project (tab). This is only applied to this project, so the rest of their environment is untouched.
- There's a single solution and project file that covers all the frameworks. The whole file is just under 60 lines, and covers all the previous csproj combined + nuspec settings + ApplicationInfo.cs ([FluentFTP.csproj](https://github.com/artiomchi/FluentFTP/blob/e83391b37510345ec095836f905278bbe18ffe82/FluentFTP/FluentFTP.csproj))
  - Uses confitional attributes to add individual configuration for different frameworks
  - Defines a single point for project configuration and metadata
  - Much easier to manage due to being so concise and easily editable in VS (or any text editor)
  - Builds code against all frameworks in one go!
  - Also correctly packages all relevant resources in a nupkg file, ready to be uploaded to nuget directly!
- New VS and combined project format makes managing multi framework code a LOT easier to manage:  
image
Framework selection in the file editor
image
NO_SSL code not compiled in most frameworks
image
NO_SSL code compiled in .NET Standard 1.4
image
Great visibility of which frameworks contain individual methods/properties

**Differences:**
- I've removed the DNX Core compilation, as it should be irrelevant at this point. I've just left the .net full frameworks and .NET Standard configurations
- While the new project format can build the .nupkg file, there are some things that it doesn't set, specifically the Summary and Language fields. If that's important to the author, an external .nuspec file can be used as well, but it does cause a disconnect between the project contents and the package file. I'm happy to change it back (release notes and summary can easily be amended in the nuspec file before uploading it, or after uploading it on the NuGet website)


artiomchi:
**Suggestions:**
- The tests project should really be running actual tests. This could easily be switched to use XUnit, for example
- The tests project only runs against the .NET framework. Currently it would fail against .NET standard (hence they are not added in `Tests.csproj`
- The repo could do with a CI setup, to build each commit. AppVeyor provides that for free for OSS projects :)

If you're interested in any of the suggestions, I could help with that too

artiomchi:
Pinging @hgupta9 @jptrosclair 

I'm hoping this project is alive, but there's a bunch of issues piling in, and we're hoping this project won't go under.

Are any of the project members still around? :)

hgupta9:
Yes it's still alive. I had to take a break due to personal reasons but now I'm going to be active again. Thank you @artiomchi for the brilliant work. I will merge this hope it compiles fine on my end. So all I need is VS 2017, and I can compile for all frameworks and nuget too?

hgupta9:
Tests are hard because we need a free FTP server donated and my company cannot provide that atm. CI is very interesting but again that can only work after tests are configured.

artiomchi:
Regarding unit tests, I was thinking it would be nice to find a "sponsor" company that would be willing to host an FTP server, or allow us to use their server for unit tests.

One server that I used when doing some testing was Rebex: http://test.rebex.net/

I guess we could drop them an e-mail, and if they'd be OK with this - use them in test projects.

The unit tests would cause hardly any traffic at all, but would help drastically!

Issue #125, opened 2017-06-21
.NET Core Disable output to Console


Yannick183:
Hi,

Is there a way to disable the logging to Console or Debug on the .NET Core version ?

hgupta9:
There is now. Please update your nuget package. And see this for [usage info](https://github.com/hgupta9/FluentFTP#logging).

Does that work for you?

Yannick183:
I don't see the updated NuGet package. The latest version is still 17.5.0

hgupta9:
Sorry, must have been a mistake. I've pushed the latest version now.

Yannick183:
It's all good now! And it's no longer writing to the console. 

Thank you!

hgupta9:
Good to hear that.

PR #196, opened 2017-09-22
Fix download bug created by PR #125

taoyouh:
PR #125 just created a new bug. When downloading a file, if the DownloadFileInternal method successfully gets the file size, readToEnd is false, and it will go into the newly created "throw" statement, even with all the contents have finished downloading.


robinrodricks:
Thank you very much :)

taoyouh:
#175 Actually

PR #197, opened 2017-09-23
Task-based async methods for .NET Standard and .NET Framework 4.5

taoyouh:
Most network based methods now have a new asynchronous version for .NET standard and .NET Framework 4.5. They used asynchronous network API instead of synchronous ones. If .NET Framework 4.5 already has same methods implemented by wrapping Begin/End methods, they are replaced.
By using these new asynchronous methods, apps will not freeze when waiting for server reply, and there will be no threads created like Task.Run() does.

The newly written methods should run as synchronous ones do, because their codes are only slightly different with synchronous ones. However, I did not test all of them yet because I didn't get the usage of the Test project.

Compatibility change:

- For .NET Standard, current methods are unchanged.
- For .NET Framework:
    - Some methods are changed from wrapping Begin/End methods to a new implementation and their locks are removed
    - Some methods changed return type from Task to Task<bool> for consistency with synchronous version


robinrodricks:
Very cool and great work so far. Thank you for this.

robinrodricks:
I will bump the major version number when releasing this as there are breaking changes. Thanks for the extensive work.

PR #198, opened 2017-09-23
Basic CI using AppVeyor

artiomchi:
Following on from my previous suggestion re: automated builds, I decided to split this into two stages:

1. Implement Continuous Integration using AppVeyor (Free for open source projects)
2. Add unit testing with a live server

Since we don't have a live server that we can test against, I thought it would be good to start with at least the first stage. (while I'm looking into converting the Tests project into a working xUnit proj)

The main benefit is that each commit will have a build run against it, so you can immediately see whether there are compilation errors. Better yet, it will also be displayed on the PR screens, to make sure that pull requests won't bring in any compilation errors :)

To get this working, the only thing you need to do is create an AppVeyor account and point it to this repository. The `appveyor.yml` defines the commands needed to run the build and pack commands.

You can see an example build for this commit here, including the full output, and the build artifact (which can be directly uploaded to NuGet!)

### DIFF
--- a/FluentFTP.Tests/Tests.cs
+++ b/FluentFTP.Tests/Tests.cs
@@ -7,259 +7,74 @@ using System.Threading;
 using System.Collections.Generic;
 using System.Text.RegularExpressions;
 using System.Linq;
-#if NETFX45
+#if !NOASYNC
 using System.Threading.Tasks;
 #endif
 using System.IO.Compression;
 using System.Text;
 using FluentFTP.Proxy;
 using System.Security.Authentication;
+using Xunit;
 
-namespace Tests {
-	class Tests {
+namespace Tests
+{
+	public class Tests
+	{
+		private const string Category_Code = "Code";
+		private const string Category_PublicFTP = "PublicFTP";
+		private const string Category_CustomFTP = "CustomerFTP";
 
 		// SET THESE BEFORE RUNNING ANY TESTS!
-		static string m_host = "";
-		static string m_user = "";
-		static string m_pass = "";
+		const string m_host = "";
+		const string m_user = "";
+		const string m_pass = "";
 
-	    private static readonly int[] connectionTypes = new int[] {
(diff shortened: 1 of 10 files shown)


artiomchi:
P.S. I've implemented the same for the R4Mvc project, and in there we also have Continuous Deployment to MyGet for CI builds, and to NuGet for final release builds. It's pretty easy to implement, but it's up to you whether you want it, or whether you want to manually deploy to NuGet :)

robinrodricks:
Thanks for this. I don't mind automated builds but when exactly does it push to nuget? Everytime someone pushes to trunk? Or can I manually run the command? Also versioning, will it auto increment or can I manually set the version? Manual is preferred...

artiomchi:
Oh no, this PR just adds CI, CD is just something we could extend this to in the future.

This config will only run a build for each commit, making sure that the code compiles, then runs unit tests, and creates a nupkg file ready for deployment. I've just pushed in a tests branch which has some of the basic tests turned on (mostly synthetic code tests, and a couple tests against public ftp servers). It runs them against two frameworks (.NET 4.5.2 and NetCoreApp 1.0, which is the same api level as .NET Standard 1.6).

Version number is not affected (it's manually set in the project's .csproj), and no auto deployments are done (it just gives you the file to do as you wish as an "artifact). The main benefit of this is visibility of the branch stability and continuous testing. It'll also show the status of the build on each Pull Request as well on GitHub

artiomchi:
I forgot to mention that while working on this I found several methods that had their async versions visible only to net45, and not to .net standard. I've tested them, and they work perfectly fine, so I re-introduced them in this PR.

robinrodricks:
Sounds cool. Excellent work as usual. I'm having some internet issues so I am writing this on a mobile device. When my computer is online I will setup the AppVeyor account. Thanks for all the contributions.

artiomchi:
My pleasure! It's a great project, and I'm using it too, so the last I can do is help make it better :)

PR #204, opened 2017-09-26
Async compiler keyword + More secure SSL protocols

artiomchi:
In a number of places in code the library uses `#if NET45 || CORE` to check if the framework supports await/async keywords. To make it more readable, I've added a `ASYNC` compiler keyword.

Also, I noticed that the library defaults to SSL3 + TLS 1.0 by default on .NET core. SSL3 is now considered obsolete and insecure, and the default since .NET 4.6 is to use TLS 1.0-1.2, so I've updated that.

The FtpClient class has an override to customise that (say you want to connect to older servers), but it was ignored by the `FtpSocketStream` class, so I've rectified that as well, making sure that we store the default `SslProtocols` value in a single place.

### DIFF
--- a/FluentFTP/Client/FtpClient_Connection.cs
+++ b/FluentFTP/Client/FtpClient_Connection.cs
@@ -571,7 +571,7 @@ namespace FluentFTP {
 #endif
 
 #if CORE
-		private SslProtocols m_SslProtocols = SslProtocols.Tls11 | SslProtocols.Ssl3;
+		private SslProtocols m_SslProtocols = SslProtocols.Tls12 | SslProtocols.Tls11 | SslProtocols.Tls;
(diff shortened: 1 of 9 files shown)


robinrodricks:
Thank you!!

robinrodricks:
New version published to nuget at last. I was able to easily build it and create the nuget package during the build. Credits added into [release notes](https://github.com/robinrodricks/FluentFTP#release-notes) for the 19.0.0 version. Credits also added [here](https://github.com/robinrodricks/FluentFTP#credits). Thanks for all the help.

robinrodricks:
Can you please check the nuget package for .NET apps and .NET standard apps since its the first time I'm building via your new workflow?

PR #226, opened 2017-10-09
Fix a DownloadFileAsync/UploadFileAsync bug

taoyouh:
CurrentDataType was not modified in SetDataTypeAsync(), so if async methods are used, the downloading may not switch to correct data type. Causing some files to be downloaded in ASCII mode accidentally and sometimes cause unexpected end of stream exception.
In SetDataType(), CurrentDataType is modified, and now SetDataTypeAsync is set to the same behavior.


robinrodricks:
Thank you for this!

PR #199, opened 2017-09-24
Fixing incorrect stream closing/disposing

artiomchi:
While working on the project testing, I stumbled upon a problem with some operations when running in .NET Standard. The library worked fine in .NET Framework, although it only did because it worked around the bug in the first place.

The .NET `Stream` class has three methods (relevant to this issue): `Close()`, `Dispose()` and `Dispose(bool)`. Both dispose methods are blank.  
`Dispose()` calls `Close()`  
`Close()` calls `Dispose(bool)`

The .NET Core's `Stream` class only has `Dispose()` and `Dispose(bool)`, both are blank.

To compensate for the missing `Close()`, the `FtpSocketStream` creates a new `Dispose()` method that calls `Close()`, overrides the `Close()` method (**not calling base.Close()**, and so not calling `Dispose(bool)`). For .NET Core, it creates a `Close()` method, and calls `Dispose()` inside. So we *already* have different paths in here.

In the `FtpDataStream` (derives from `FtpSocketStream`), the library overrides `Dispose(bool)` and **calls `Close()`**. This would actually cause the Close to call Dispose, but (un)luckily, the Close method doesn't call the base Close method, avoiding the circular call. This is the reason the error wasn't caught while running in .NET framework!

The main side-effect of this? In `FtpDataStream`'s `Dispose(bool)` method, we're clearing the `m_control` variable, without checking if the data stream has been properly cleared. In fact, just below, the `Close()` method checks if `m_control`'s property is not null, and if it's not, clears the control channel before nulling it.

If `Dispose(bool)` is called before `Close()` in this class, the library will never clear the data stream. That's exactly what happens when running under .NET core.
image

Fixes #160

PR #229, opened 2017-10-10
Bug fixes for file downloading

taoyouh:
1. fix wrong type checking in DownloadFileInternal and DownloadFileInternalAsync
2. fix the bug that sometimes FtpSocketStream and FtpDataStream are not disposed in Dispose()

### DIFF
--- a/FluentFTP/Client/FtpClient_HighLevel.cs
+++ b/FluentFTP/Client/FtpClient_HighLevel.cs
@@ -1795,6 +1795,8 @@ namespace FluentFTP {
 
 							// if we reach here means EOF encountered
 							// stop if we are in "read until EOF" mode
+							downStream.Dispose();
+
 							FtpReply reply = GetReply();
 							if (!readToEnd && offset != fileLen)
 								throw new IOException($"Unexpected EOF for remote file {remotePath} [{offset}/{fileLen} bytes read]");
@@ -1853,6 +1855,8 @@ namespace FluentFTP {
 
 							// if we reach here means EOF encountered
 							// stop if we are in "read until EOF" mode
+							downStream.Dispose();
+
 							FtpReply reply = GetReply();
 							if (!readToEnd && offset != fileLen)
 								throw new IOException($"Unexpected EOF for remote file {remotePath} [{offset}/{fileLen} bytes read]");
@@ -1879,7 +1883,6 @@ namespace FluentFTP {
 
 				// disconnect FTP stream before exiting
 				outStream.Flush();
-				downStream.Dispose();
 
 				return true;
 
@@ -1949,6 +1952,8 @@ namespace FluentFTP {
 
 							// if we reach here means EOF encountered
 							// stop if we are in "read until EOF" mode
+							downStream.Dispose();
+
 							FtpReply reply = await GetReplyAsync();
 							if (!readToEnd && offset != fileLen)
 								throw new IOException($"Unexpected EOF for remote file {remotePath} [{offset}/{fileLen} bytes read]");
@@ -2003,6 +2008,8 @@ namespace FluentFTP {
 
 							// if we reach here means EOF encountered
 							// stop if we are in "read until EOF" mode
+							downStream.Dispose();
+
 							FtpReply reply = await GetReplyAsync();
 							if (!readToEnd && offset != fileLen)
 								throw new IOException($"Unexpected EOF for remote file {remotePath} [{offset}/{fileLen} bytes read]");
@@ -2028,7 +2035,6 @@ namespace FluentFTP {
 
 				// disconnect FTP stream before exiting
 				await outStream.FlushAsync(token);
-				downStream.Dispose();
 
 				return true;
 
--- a/FluentFTP/Client/FtpClient_HighLevel.cs
+++ b/FluentFTP/Client/FtpClient_HighLevel.cs
@@ -1769,7 +1769,7 @@ namespace FluentFTP {
 
 				// if the server has not reported a length for this file
 				// we use an alternate method to download it - read until EOF
-				bool readToEnd = (fileLen <= 0);
+				bool readToEnd = (fileLen <= 0) || CurrentDataType != FtpDataType.Binary;
 
 
 				// loop till entire file downloaded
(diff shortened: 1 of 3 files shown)


taoyouh:
CurrentDataType is not changed before OpenRead, so I changed CurrentDataType to DownloadDataType.
This is a bug fix irrelevant with the PR description.

robinrodricks:
I appreciate all your improvement, however modifying the Dispose/Remove set of methods would conflict with the work done by @artiomchi in #199. Unless he approves of this I cannot accept this. Or you understand thoroughly his fixes especially for .net core and ensure you don't break his fixes.

taoyouh:
@artiomchi 
On .NET core, when you do something like the following, nothing happens:
```
Stream stream = someFtpSocketStream;
stream.Dispose();
```
So downstream.Dispose() in DownloadFileInternal does nothing. My change fix that. While I am trying to understand the disposal, please help check if this fix cause other problems.

taoyouh:
It seems that in .NET Framework a circular call happens.

taoyouh:
I read the docs. In.NET Framework, Dispose() calls Close(), and Close() calls Dispose(bool). In.NET Core, Dispose() calls Dispose(bool). .NET Framework recommends that we should override Dispose(bool) but not Close().
I plan to move all closing logic to Dispose(bool) and call Dispose(bool) in .NET Core Close().

artiomchi:
@taoyouh 
> I read the docs
. In.NET Framework, Dispose() calls Close(), and Close() calls Dispose(bool).

That's correct.

> In.NET Core, Dispose() calls Dispose(bool).

Nope, it actually doesn't. If you check the source code, both Dispose() and Dispose(bool) are blank methods in .NET Core.

So what happens is, if you have a stream and wrap it in a `using()` block, `Dispose(bool)` will never be called!

To be perfectly honest, I'm not sure if that's a bug in .NET core or not, so I'll raise it with them, but based on the current source code - the current behaviour should be the correct one.

Also..
> So downstream.Dispose() in DownloadFileInternal does nothing.

What do you mean it does nothing? Didn't you just say that `Dispose()` calls `Close()` which calls `Dispose(bool)`?

Could you clarify what the problem is, please?

artiomchi:
See issue at dotnet/corefx#24543

taoyouh:
@artiomchi
My test on UWP shows that Stream.Dispose() calls Stream.Dispose(true). But FtpDataStream doesn't have Dispose(true) overridden. And downstream.Dispose() calls Stream.Dispose() rather than FtpSocketStream.Dispose(). Using "new void Dispose()" means you can't access the method when it is a Stream class variable.

Problem is that the downstream.Dispose() in DownloadFileInternal and DownloadFileInternalAsync doesn't really reach disposing logic in UWP (Tested) and .NET Standard (Logically)

artiomchi:
@taoyouh I've gotten used to decompiling the source code for assemblies to check the inner workings of a library, and that's where I got misinformed. Seems like the reference assemblies were not accurate.

Also, the Stream class got it's `Close()` method restored at some point, although it seems like it's not always visible in all the packages (the reference assemblies aren't making it visible, but the core framework seems to actually have had it since .NET Core 1.0).

Anyway, my point is - I think you were right to move the dispose code to `Dispose(bool)`, since it will be called in both .NET Framework and .NET Core. `new void Dispose()` was definitely a bad idea, but it was a workaround around the issue I thought I found, which was a misunderstanding.

I think your original PR is probably a good idea (you seem to have reverted parts of it in f8331d7), so I'd suggest you that last commit :)

I'd probably completely avoid using the `Close()` method, since it's not visible pre .NET Core 2.0. So you should remove any references/calls to it, and just use `Dispose(bool)` instead.

We shouldn't need to have to call it anyway, since all the Streams are wrapped in a `using` block in code, making sure they'll be disposed!

Good job on the PR and the analysis!

robinrodricks:
@taoyouh After you make the changes requested by artiom I will merge the PR. Keep your new Close/dispose logic. Hopefully no new issues about it come up.

artiomchi:
@robinrodricks I'm guessing you haven't set up the appveyor account?

Could you set one up and link it to this repository? It would be useful to see the results of all the unit tests in PRs, especially for pull requests like this. Even though we don't have many tests, there are actually a couple tests there that connect to public servers and where I managed to reproduce the errors with Dispose(). 

The unit tests run against both the .NET framework and .NET Core, making sure we don't break any existing behaviour in either framework.

artiomchi:
@taoyouh I think your change to the `FtpSocketStream` class is good and a definite improvement.  I'd leave it as is now.

Regarding the changes in FtpClient_HighLevel, is that really the best way to do this?
- Why not just check the response code after the download around line 1870, just after the original dispose? This would simplify the code. It seems like you were making these extensive changes before you fixed the Dispose() method, and they might be unnecessary.
- You changed the CurrentDataType to DownloadDataType check. Will this have any side-effects?

robinrodricks:
@artiomchi Ah yes, totally forgot about that as I have been very busy this last month. Will set it up soon.

@taoyouh I think the DownloadDataType change is correct, because it's the type that the user has chosen to download in...

taoyouh:
@artiomchi The story is long. First I tried to fix ASCII download problem, so I added checking for reply code. I misunderstood the resume so I added the code there. Then I found that the code hangs at Dispose() so I fixed the disposal.

taoyouh:
There are some problems with the reply code check, which I have reverted.
Now there are two fixes:
1. fix wrong type checking
2. fix incorrect disposal which (sometimes) cause hanging

artiomchi:
Oh yeah, this looks great! I don't think it'll have any side-effects, but I wonder if you could run the unit tests locally, just to make sure they all succeed, since Robin didn't get a chance to set up the automated testing yet. If the unit tests are successful, I think Robin can merge this branch straight away :)

Just run these two commands from the root of the repository:

```
dotnet test fluentftp.tests -f net452
dotnet test fluentftp.tests -f netcoreapp1.0
```

P.S. In the FtpDataStream class, you changed the call from Dispose() to Dispose(false).. The end result is the same, but I was wondering
- Which method do you think makes more sense to call here?
- Do we even need the finaliser, or is that just as a safety net, in case someone didn't wrap the stream in a using block?

taoyouh:
ObjectDisposedException on .NET Core App 1.0

taoyouh:
ObjectDisposedException reason:
Ftp.netbsd.org is slow here, it needs more than 2 seconds to get its listing. Before the client in thread 1 gets the listing, thread 2 disposed it. I think this is not a problem.

taoyouh:
@artiomchi As for the finaliser, the default finaliser calls Dispose(false). I think the override here is just to catch the exception.

robinrodricks:
@taoyouh Please tell me when you are ready to merge.

taoyouh:
Bugs are fixed and no side effects are left. Ready to merge now.

robinrodricks:
Thank you. Can you edit this list to the final list of changes made?

1. Listen for 2xy reply when data connection is closed to ensure download is success.
2. Close data connection when the server closes the connection. Because some servers (ProFTPD) wait until the connection is totally shut down by two sides before sending a reply.
3. Override Dispose(bool) instead of create new Dispose() to make sure disposal is correct. In the previous version, if FtpSocketStream is in type Stream, disposal does nothing. Calling Stream.DIspose() automatically calls Dispose(true), so we don't need to create a new Dispose().

robinrodricks:
@taoyouh Can you send the final list of changes made in this PR?

taoyouh:
1. fix wrong type checking in DownloadFileInternal and DownloadFileInternalAsync
2. fix the bug that sometimes FtpSocketStream and FtpDataStream are not disposed in Dispose()

robinrodricks:
What about "Listen for 2xy reply"? Did you revert that change?

taoyouh:
I reverted "listen for 2xy reply" because I found some problems there.

robinrodricks:
Okay, thanks.

PR #233, opened 2017-10-20
Make IFtpClient inherit from IDisposable

repl-andrew-ovens:
This enables the interface to be used in a using() block


robinrodricks:
Released on nuget as https://www.nuget.org/packages/FluentFTP/19.1.2

PR #234, opened 2017-10-21
Check file existence in background thread

taoyouh:
UWP does not allow File.Exists() to run in UI thread and throws an exception when doing that. [See StackOverflow entry](https://stackoverflow.com/questions/35747635/perform-synchronous-operation-on-ui-thread) 
The File.Exists() in asynchronous methods are wrapped in Task.Run to fix this.
The ones in synchronous methods are not changed because synchronous methods are not supposed to be used in UI thread.


robinrodricks:
Thanks. No need to slow down the desktop version however, so please duplicate that line within an #if #else #endif block, keeping one line for desktop and the other line for .Net standard.

robinrodricks:
Thank you for this!

robinrodricks:
Released on nuget as https://www.nuget.org/packages/FluentFTP/19.1.2

PR #242, opened 2017-11-10
Throw OperationCanceledException when cancellation is requested

taoyouh:
Throw OperationCanceledException instead of FtpException when cancellation is requested.


robinrodricks:
Thanks!!

PR #261, opened 2018-01-02
moved check for 220 before the Authenticate()

gone4wd:


robinrodricks:
Good, but are you using the bluecoat server? If now, create another proxy file and modify that instead so it doesn't interfere with existing code that uses the bluecoat class.

PR #276, opened 2018-02-26
Fixes for using active FTP and specifying the ActivePorts

ToniMontana:
* FtpClient_LowLevel.cs need to break the foreach loop when a port is successful otherwise we get errors.
* FtpClient_Connection.cs needs to clone the ActivePorts as well when using EnableThreadSafeDataConnections.


robinrodricks:
Thanks for this!

Issue #294, opened 2018-05-02
Asynchronous methods error when callback invoked before IAsyncResult added to dictionary


Lukazoid:
**FTP OS:** All

**FTP Server:** All

**Computer OS:** Windows 10 and Windows Server 2008 R2

When using `FtpClient.CreateDirectoryAsync(string path)` I observed the following exception and stack trace:

    System.InvalidOperationException: The specified IAsyncResult could not be located.
       at FluentFTP.FtpClient.GetAsyncDelegate[T](IAsyncResult ar)
       at FluentFTP.FtpClient.<CreateDirectoryAsync>b__53(IAsyncResult ar)
       at System.Threading.Tasks.TaskFactory`1.FromAsyncCoreLogic(IAsyncResult iar, Func`2 endFunction, 
    Action`1 endAction, Task`1 promise, Boolean requiresSynchronization)

We have been using what was System.Net.FtpClient and now is FluentFtp in production heavily for several years now and have never seen this error before.

This led me to delve into the internals of the `FtpClient` async methods (`CreateDirectoryAsync` in particular). As the async methods use `Task.Factory.FromAsync` I took at look at the `BeginCreateDirectory` and `EndCreateDirectory` implementations. 

It appears that the `Begin` methods add the `IAsyncResult` to `m_asyncmethods` after `BeginInvoke` returns (see [source](https://github.com/robinrodricks/FluentFTP/blob/ac6ef5c5dc4a4de81910f686056af5e1337ba97c/FluentFTP/Client/FtpClient_Management.cs#L836)), this does not work in the case when the asynchronous method completes quickly (i.e. before adding to `m_asyncmethods`), this results in the `AsyncCallback` being invoked before the item is added to `m_asyncmethods`.

This behaviour of `BeginInvoke` calling the callback before the item is added to `m_asyncmethods` can be demonstrated with a small sample, this also replicates what is happening in `FtpClient`:

    Func<int, int> callback = x => x * 2;
    var m_asyncmethods = new Dictionary<IAsyncResult, object>();

    while (true)
    {
        await Task<int>.Factory.FromAsync((cb, state) =>
        {
            var asyncResult = callback.BeginInvoke(5, cb, state);
            lock (m_asyncmethods)
                m_asyncmethods.Add(asyncResult, callback);

            return asyncResult;
        }, r =>
        {
            lock (m_asyncmethods)
            {
                if (!m_asyncmethods.ContainsKey(r))
[...]

PR #296, opened 2018-05-03
Fix a race condition with all async methods.

Lukazoid:
Fix a race condition when BeginInvoke calls the callback before the IAsyncResult is added to m_asyncmethods.

By moving the BeginInvoke inside of the lock it is no longer possible for the End method to be called before the IAsyncResult is added to m_asyncmethods.

This fixes issue #294.


robinrodricks:
Thank you for this!

robinrodricks:

Issue #308, opened 2018-06-15
Unexpected EOF for remote file


psp589:
**FTP OS:** Windows 

**FTP Server:** Filezilla

**Computer OS:** windows 10 running an android emulator with Nougat 7.1

When I try to download a file from UWP it works without problems, when I try to download the same file using an android emulator I got an exception : 

FluentFTP.FtpException: Error while downloading the file from the server. See InnerException for more info. ---> System.IO.IOException: Unexpected EOF for remote file CAPT0004/db.zip [3744141/3744181 bytes read]
  at FluentFTP.FtpClient.DownloadFileInternal (System.String remotePath, System.IO.Stream outStream, System.IProgress`1[T] progress) [0x000c6] in <1bd16e2bef914c27b4aadfb191e1fd86>:0 
   --- End of inner exception stack trace ---
  at FluentFTP.FtpClient.DownloadFileInternal (System.String remotePath, System.IO.Stream outStream, System.IProgress`1[T] progress) [0x00283] in <1bd16e2bef914c27b4aadfb191e1fd86>:0 
  at FluentFTP.FtpClient.DownloadFileToFile (System.String localPath, System.String remotePath, System.Boolean overwrite, FluentFTP.FtpVerify verifyOptions, System.IProgress`1[T] progress) [0x00073] in <1bd16e2bef914c27b4aadfb191e1fd86>:0 
  at FluentFTP.FtpClient.DownloadFile (System.String localPath, System.String remotePath, System.Boolean overwrite, FluentFTP.FtpVerify verifyOptions, System.IProgress`1[T] progress) [0x0005b] in <1bd16e2bef914c27b4aadfb191e1fd86>:0 
  
I read something about setting the data type to Binary  so I execute this couple of lines before calling DownloadFile method : 
            _FtpClient.DownloadDataType = FtpDataType.Binary;
            _FtpClient.UploadDataType = FtpDataType.Binary;
But I get the same error 

**Logs :**

```
06-15 12:43:28.976 I/mono-stdout(24763): 
06-15 12:43:28.976 I/mono-stdout(24763): # FileExists("CAPT0004/db.zip")
06-15 12:43:28.976 I/mono-stdout(24763): 
06-15 12:43:28.977 I/mono-stdout(24763): # GetWorkingDirectory()
06-15 12:43:28.977 I/mono-stdout(24763): Command:  PWD
Response: 257 "/" is current directory.
Command:  SIZE /CAPT0004/db.zip
[...]

PR #314, opened 2018-07-03
Fixes Many Unexpected EOF for remote file IOException on Android

jersiovic:
Related with issue https://github.com/robinrodricks/FluentFTP/issues/308
In a multiplatform app for UWP and Android that shares assemblies for accesing ftp code I found that for the same files (of approximattely 6 MB each file) and same server (Filezilla server) the UWP app doesn't fails downloading  those files, but Android app fails almost every time with this exception:  IOException "Unexpected EOF for remote file ..." https://github.com/robinrodricks/FluentFTP/blob/master/FluentFTP/Client/FtpClient_HighLevel.cs#L1814

Problem could be caused by this behavior of the java bytecode to which is translated C# in a xamarin code: https://wiki.sei.cmu.edu/confluence/display/java/FIO10-J.+Ensure+the+array+is+filled+when+using+read%28%29+to+fill+an+array

Initially I tried to fix it with the proposed solution at that post, I mean retrying a read  when readbytes is 0 and the total size of the stream was not readed but despite initially it looked it worked, later I found it doesn't work cause at some point next reads always return a 0 value for bytes read.
What finally has worked properly is to include code in ResumeDownload method to also resume download when this exception is thrown.

### DIFF
--- a/FluentFTP/Client/FtpClient_HighLevel.cs
+++ b/FluentFTP/Client/FtpClient_HighLevel.cs
@@ -2066,12 +2066,12 @@ namespace FluentFTP {
 
 		private bool ResumeDownload(string remotePath, ref Stream downStream, long offset, IOException ex) {
 			// resume if server disconnects midway (fixes #39)
-			if (ex.InnerException != null) {
+			if (ex.InnerException != null || ex.Message.StartsWith("Unexpected EOF for remote file")) {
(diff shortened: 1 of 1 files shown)


robinrodricks:
Looks good! Thanks for this!

robinrodricks:

PR #316, opened 2018-07-16
Update FtpClient_LowLevel.cs

olivierSOW:
workaround for ftpd which response "550 No files found" when folder exists but is empty


robinrodricks:

PR #333, opened 2018-09-11
Make Dispose method of FTPClient virtual

martinbu:
In this PR the Dispose method of the FtpClient is made virtual. This makes it possible to add custom dispose logic in an inherited class. So far I have overriden the virtual Disconnect method. But I think making the dispose method virtual is more accurate.


martinbu:
The really correct pattern is also described here. [Correct Dispose Pattern](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose)

robinrodricks:

Issue #248, opened 2017-11-20
Get All Directory needs a manual recursion fallback if the server doesn't support recursion


datar12345:
I having issues.
How to get all directory, subdirectory and file from path root.

ex:
using: ftpClient.GetListing("home/")
result: "home/dir1"
           "home/dir2"
           "home/dir1/csv"
           "home/dir2/csv"
           "home/dir1/csv/test.txt"
           "home/dir2/csv/test.txt"

robinrodricks:
You can add flags to getlisting, check the other overloads.

datar12345:
I use ASP .NET CORE.
I tried to use FluentFTP as after:

var ftpClient = new FtpClient(Configuration["FTPServices:Host"]
                            , Configuration["FTPServices:User"]
                            , Configuration["FTPServices:Password"]) {
                    DataConnectionType = FtpDataConnectionType.AutoActive,
                    DataConnectionEncryption = false,
                    EncryptionMode = FtpEncryptionMode.None,
                    Encoding = System.Text.Encoding.UTF8,
                });

var list = ftpClient.GetListing("home/", FtpListOption.AllFiles | FtpListOption.ForceNameList)
// or -> var list = ftpClient.GetListing("home/", FtpListOption.AllFiles)
// or -> var list = ftpClient.GetListing("home/", FtpListOption.ForceList)

-> But it only returns 2 paths("home/dir1", "home/dir2"), not 6 paths.

I tried to use FtpListOption.UseLS, but an error occurred: command unknown.

May, server not support MLSD, LS command.

Is there any other way.

Please guide help me.

robinrodricks:
Can you remove all your options and only add ftplistoption.recursive??

datar12345:
I tried, but not successful.

I know that the dir -r command is used to get lists all files in current directory and subdirectories.
And i tried use cmd win 7 to connect ftp server and use dir -r command, but it only returns 2 paths("home/dir1", "home/dir2").

Is this the cause I not success with ftplistoption.recursive?

I think you know lot about FTP, so can you explain help me.

beantaxi:
I am having the same issue ... 

ftpClient.GetListing("/", FtpListOption.Recursive);

Only gets me the top level. This is not specific to '/' - I tried other folders and got the same result.

I am very new to FluentFtp, so no clue if this is a regression on not. Any more info I can provide to help?

diegoaraujolima:
I am having the same issue....

[...]

PR #334, opened 2018-09-12
Exceptions Deleting Non-Empty Directories

digitalcoyote:
Altered Recursive Directory deletion so that if it fails to delete a directory it will attempt recursively attempt to empty the directory before trying to delete it again. This allows FTP Servers that appear to not support the Recursive List to fallback to recursively calling List and deleting files.


digitalcoyote:
If #248 is fixed, this may not be necessary as a recursion fallback there would allow us to not use recursion here

PR #355, opened 2019-01-07
Added BufferedStream between SslStream and NetworkStream

### LINKED ISSUE #354: Huge performance improvement for uploads
**FTP OS:** Unix

**FTP Server:** Various

**Computer OS:** Windows Server 2008 R2

It has been observed how much slower FluentFTP uploads are compared to that of other FTP applications (https://github.com/robinrodricks/FluentFTP/issues/332), in our application, concurrent uploads were taking an hour with FluentFTP and about 6 minutes with FileZilla.

By looking at the code I have determined a major source of FluentFTP's slowdown, there is no buffering between the SSL stream and the Network Stream, this means we end up with a lot of small writes from the SSL stream to the Network Stream. Writing to a network stream is incredibly slow as this requires marshalling, by using a BufferedStream we can vastly reduce the amount of marshalling involved.

Through the use of reflection I was able to inject a BufferedStream into the correct locations, after this change our uploads now actually upload faster than FileZilla, taking 5 minutes instead of the previous hour.

Here is the code I used to achieve this:

```
private static BufferedStream InjectBufferedStream(Stream ftpDataStream, int bufferSize = 81920)
{
    Contract.Requires(ftpDataStream != null);
    Contract.Ensures(Contract.Result<BufferedStream>() != null);

    // A gross method to inject a BufferedStream between the SslStream and the NetworkStream
    // remove this if we ever update to a version of FluentFtp with this built in

    const BindingFlags privateField = BindingFlags.NonPublic | BindingFlags.Instance;

    var sslStreamField = typeof(FtpSocketStream).GetField("m_sslStream", privateField);
    var sslStateField = typeof(SslStream).GetField("_SslState", privateField);
    var sslStateInnerStreamField = sslStateField.FieldType.GetField("_InnerStream", privateField);
    var authenticatedInnerStreamField = typeof(AuthenticatedStream).GetField("_InnerStream", privateField);

    var sslStream = (SslStream)sslStreamField.GetValue(ftpDataStream);
    var sslState = sslStateField.GetValue(sslStream);

    var networkStream = (NetworkStream)authenticatedInnerStreamField.GetValue(sslStream);

    var bufferedStream = new BufferedStream(networkStream, bufferSize);
    authenticatedInnerStreamField.SetValue(sslStream, bufferedStream);
    sslStateInnerStreamField.SetValue(sslState, bufferedStream);

    return bufferedStream;
}
```

With this reflection version, the `BufferedStream` must NOT be disposed and must have `Flush` or `FlushAsync` called before the ftp data stream is disposed. This will not be required in the final merge request version.

For example:

```
using (var remoteStream = await ftpClient.OpenWriteAsync(remoteFilePath, FtpDataType.Binary).ConfigureAwait(false))
{
    var bufferedStream = InjectBufferedStream(remoteStream);

    await fileContents.CopyToAsync(remoteStream, 81920, CancellationToken.None).ConfigureAwait(false);

    await bufferedStream.FlushAsync(CancellationToken.None).ConfigureAwait(false);
}
```

I will follow this issue with a merge request to actually make this change.

**Logs :**

```
N/A
```

Lukazoid:
This reduces the large number of small marshalled writes to the NetworkStream, massively increasing performance when using SSL.

During testing of these changes a file which was previously taking 1 minute 5 seconds to upload is now taking 12 seconds to upload.

Closes #354

### DIFF
--- a/FluentFTP/Stream/FtpSocketStream.cs
+++ b/FluentFTP/Stream/FtpSocketStream.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.IO;
 using System.Net.Sockets;
 using System.Net.Security;
@@ -147,6 +147,8 @@ namespace FluentFTP {
 		}
 
 #if !NO_SSL
+		BufferedStream m_bufStream = null;
+		
 		SslStream m_sslStream = null;
 		/// <summary>
 		/// The encrypted stream
@@ -683,6 +685,15 @@ namespace FluentFTP {
 		/// </summary>
 		protected override void Dispose(bool disposing) {
 			FtpTrace.WriteStatus(FtpTraceLevel.Verbose, "Disposing FtpSocketStream...");
+
+#if !NO_SSL
+			if (m_bufStream != null) {
+				// ensure the last of the buffered bytes are flushed
+				// before we close the socket and network stream
+				m_bufStream.Flush();
+			}
+#endif
+
 			if (m_socket != null) {
 				try {
 					if (m_socket.Connected) {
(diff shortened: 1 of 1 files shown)


robinrodricks:
Excellent work!! I'll check this out soon and let you know!

robinrodricks:
Thanks a lot!

robinrodricks:

PR #362, opened 2019-01-27
Added new upload option to AppendNoCheck so that we can append with no check.

everbalovas:
Added new upload option to AppendNoCheck so that we can append a file on the server without checking if it exists yet.  This is required if the server you are uplaoding to does not allow you to list the contents of the folder, but the file does exist.

The call will throw an exception when the NLIST command is used, this will bypass the exception since we are not checking the directory for the file.

### DIFF
--- a/FluentFTP/Client/FtpClient_HighLevel.cs
+++ b/FluentFTP/Client/FtpClient_HighLevel.cs
@@ -1023,8 +1023,18 @@ namespace FluentFTP {
 
 				// check if the file exists, and skip, overwrite or append
 				if (existsMode == FtpExists.NoCheck) {
+					checkFileExistsAgain = true;	
+				}
+				else if (existsMode == FtpExists.AppendNoCheck) {
 					checkFileExistsAgain = true;
-				} else {
+					
+					offset = GetFileSize(remotePath);
+					if (offset == -1)
+					{
+						offset = 0; // start from the beginning
+					}
+				}
+				else {
 					if (!fileExistsKnown) {
 						fileExists = FileExists(remotePath);
 					}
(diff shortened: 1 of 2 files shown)


robinrodricks:
Why is this needed? Aren't the existing options enough?

everbalovas:
This is needed because we a have a major chain that customers order drugs from and they do not allow us to List from their FTP.

We need to appended to the file because an order file could be sent and updated.

Thanks,

Eric Verbalovas
Sr. Quality Assurance Analyst - THPS
TELUS Health
25 York Street
Toronto, ON M5J 2V5
T 1-800-263-5876 | C 647-217-5963

[telus_2col_600w]
The information contained herein, including any attachments, is proprietary and confidential and is intended for the exclusive use of the addressee. It also may contain privileged information and/or personal information subject to privacy legislation. The authorized addressee of this information, by its retention and use, agrees to protect the information contained herein from loss, disclosure, theft or compromise with at least the same care it employs to protect its own confidential information. Any dissemination or use of this information by a person other than the intended recipient is unauthorized and may be illegal. If you have received this e-mail in error, please notify us immediately by reply e-mail and destroy all copies.

everbalovas:
>           Why is this needed? Aren't the existing options enough?
When we send the order file, the current options do not let us append without running a list, we cannot just replace the file or overwrite it, they want us to append to the file only because of weird permission issues.

everbalovas:
Is it possible to have this merged in and published on NuGet? This will definitely be useful in this type of scenario and would be greatly appreciated

robinrodricks:
I'm still trying to understand it!!

robinrodricks:
Okay, lets do this. Thanks!

everbalovas:
Thanks robinrodricks, let me know when you build out a version on NuGet, I see 19.2.4 but it doesn't include my change.

Thanks again!

robinrodricks:
Released in [FluentFTP 20.0.0](https://www.nuget.org/packages/FluentFTP/20.0.0).

PR #368, opened 2019-02-01
Added New Property FtpReply to Client updated where function get reply is called

brucemcmillan:
Hi,

I've added a new property to the Client called FTPReply.

this as per the name is a FTpReply classs that is set by other functions where the GetReply function is called.

Helpful to store and pass back the results of the ftp call to the calling code in the client class

I needed a way to get retruned information from the response to a file upload


brucemcmillan:
request cancelled, too much un needed code

PR #369, opened 2019-02-02
Added FTPReply Class as Property to Client, Set in GetReply 

brucemcmillan:
Added FTPReply Class as Property to Client, Set in GetReply .

Results of calls can be accessed in Client


brucemcmillan:
reverted, not working in all instances

PR #370, opened 2019-02-02
Added FTPReply class as property to FTPClient

brucemcmillan:
Added FTPReply Class as Property to Client, Set when get reply is called.

Revised version, initial version was refactored to remove code but took out too much..

Results of calls can be accessed in Client


robinrodricks:
I'm sorry but this is a mess. I've implemented the `LastReply` property in the latest version.

robinrodricks:
Released in [FluentFTP 20.0.0](https://www.nuget.org/packages/FluentFTP/20.0.0).

brucemcmillan:
Hi,

Sorry it's a mess :-(  was trying to he helpful and save some time!

What happened was that I tried to create a pull request for what I thought was too much code, so tried to cancel that request. Not sure how that works here as i'ts all new to me..

I then reverted that change, and made a couple of other changes / commits.

In that time you also made a change which caused git / sourctree to have a fit and then when i fixed that locally I found that reshaarper when it formatted the files moved every { down below the () in the methods

At that stage i thought sod it, more commits etc would only make it worse, hence i left it where it was as a guideline for what i trying to achieve..

I'll kill this fork later now, just wanted to help !!

brucemcmillan:
Hi,

What sort of thing were you looking at creating, a simple test harness to support the functions of the DLL or a more full blown filezilla style client ??

The great thing about what you've got at present is that it can be implemented in other projects. I for example needed to create, strangely enough, a compliant automation server that could only be passed either simple data types or a stream of data. (yes there is still a need for this ! )

my concerns with a client are direct comparisons with filezilla / cuteftp etc all which do similar things and whilst the end result might be lovely in our eyes it might just be seen as a me too solution that will take a lot of time / maintenance to keep still.

Some more time on other features might be better spent . e.g. a auto detect function for connecting ? In got my solution working by running cuteftp and reviewing the results of the new site wizard ??

I don't want to sound at all negative here, just trying to gauge where you are in your views ??

B...

robinrodricks:
Yeah, a full blown FTP client. Well I thought of going about it bit by bit.
Start off simple and then add features as time permits. No pressure! And
also I think it would expose a lot of flaws in the underlying FTP client
implementation. Plus it would expose this lib to a whole new crowd, even
non developers.

As for the redundancy, true, filezilla is pretty good but it's unmanaged
and quite buggy. I tried using it at work and it frequently crashes with
not so much as an error dialog. Just boom! So I think there may yet be a
space for this.

brucemcmillan:
So thinking about, the whole client should be able to be completely abstracted away from what you've got in the solution so far as the UI is probably only going to need to call methods on the high level class and just reference FluentFtp as a DLL ..

It might be interesting to have a play at seeing how simple it is to create a client based on this approach as it won't affect any work you're doing under the hood. just use the nuget published version as a component ?

I'd go for the simpler winforms option as the dev time is much quicker than the WPF route...

brucemcmillan:
I've had play this evening with creating a gui client based on the sample code from the master / read me..

It's got a very basic ui with the options to connect to a host, view local / remote files and then transfer between the two places.

I've left all the worrying bits in the bit you normally do and have tried only to use the nuget package as an entry point to ftp'ing stuff.
Needs more work ,but it's a start.

have a play. you will have got a collaboration mail at some point to.

robinrodricks:
Oh nice. That was quick! I'll definitely have a look and let you know.

brucemcmillan:
HI,
Thanks for the feedback. 
Agree, it's not the prettiest yet, but it sort of proved the concept of a basic client.
WPF , of course prettier but more work. It's been years since i worked with that.. I think i started working with it when MVVM light was released.
We can certainly have a look at it.

I might move these conversations to the other repo so we can keep them in sync etc.

BTW.. I guess you not in the UK time zone ??

brucemcmillan:
had the chance to spend a bit more time playing with the app.
It's not WPF as of yet as I'm still playing with the winforms version. working out what i can create just by referencing the nuget package and the help files. I've added icons, folder navigation, upload / download functions and an output panel to show the information messages on transfers etc.

I've been looking at moving the non ui elements of the solution into a helpers library so it could be re-used..

It's been an interesting experience working with it.

I've been testing this connecting to an IIS ftp server and one thing i notice which might be config or a bug is the getremotelisting function with the additional options to add . and .. to the listing, It didn't work on this win 10 version of FTP on IIS.

robinrodricks:
How interesting. I'll take a look at your progress. Yes, it will definitely
reveal bugs as more people will be willing to use the app version than the
lib.

>

robinrodricks:
Its looking much better now! Good work with the UI. I'll try to get a basic WPF version setup today so you can begin contributing to that.

robinrodricks:
I built a basic WPF project at https://github.com/robinrodricks/FluentFTPClient

Still working on getting the basic class structure in place along with the fancy UI components I'd like to use.

brucemcmillan:
Hi,

Got a few busy days coming on so not sure what time i'm going to have.. Will have a look when i get time.

probbaly going to concentrate on tidying up the other thing first and then look at WPF, very rusty on WPF and never used the caliburn stuff just the mvvm stuff..

cheers

robinrodricks:
Oh np. I'm not going to bother using mvvm since I think its far to complicated for such a simple app. I'll just use the same principles I use to create WinForms apps. I'm just using Wpf for the graphics capabilities.

brucemcmillan:
OK cool..

I’ve been nudging my winforms app along and committing changes along the wwway.

When that’s a bit more stable I’ll see what can be done in the WPF version..

At the moment I’m using synchronous calls only, so when I’ve got the drag drop finished I’ll look at the async stuff.

Hopefully there is going to be a lot of stuff re-usuable in the helpers lib I started writing, but guess it’s going to need more refactoring to suit WPF requirements.

I’ll keep an eye on your version as time permits

Cheers

B.....

Bruce McMillan
07786 266431

________________________________

robinrodricks:
Yeah, well, we could always use more clients. Besides it will only help
build a better base library.

Just yesterday I improved the logging to support the kind of logging I was
looking for, ie. Being able to identify which FtpClient the logs were sent.

A long time ago I began on a winforms client which I abandoned as I
discovered it's graphics capabilities were very limited.

I've got the basic settings window UI ready. After that's functional I'll
begin on the main FTP client window. I'm just trying to find a decent WPF
explorer control for local and remote views. Then I'll use a listbox that
supports line coloring and set the line coloring based on the log type.

brucemcmillan:
Be good to see how you implemented logging.

I added my own customer Traclistener and then raised an event to update a list box.

Ideally it should write to a data store for the UI to access, ( perhaps later ) but wondered how you would do it..

cheers

Bruce McMillan
07786 266431

________________________________

robinrodricks:
I hooked onto FtpClient.OnLogEvent which is a new property.

See here:
https://github.com/robinrodricks/FluentFTPClient/blob/master/Managers/ClientManager.cs#L120

Which fires this:
https://github.com/robinrodricks/FluentFTPClient/blob/master/Managers/ClientManager.cs#L192

Is that what you needed?

robinrodricks:
Moved all discussion to https://github.com/robinrodricks/FluentFTPClient/issues/1

PR #375, opened 2019-02-21
created property RequestSupportedFeatures

nhh-softwarehuset:
added a property in which it is possible to define whether the client should request feature support when calling Connect. This is a neat feature in the case that you have a secured FTP user with limited access to a minimum subset of commands on the ftp server.


robinrodricks:
Looks good. However can you rename the property to CheckCapabilities?

nhh-softwarehuset:
can be done without any problems ;)

robinrodricks:
You know that many commands will fail because of this right?

robinrodricks:
FluentFTP will think that the server does not support X and those commands will fail..

robinrodricks:
Thanks anyways!

WolfspiritM:
What's with that NCrunch files? Are these really supposed to be there?

robinrodricks:
Which NCrunch files?

>

WolfspiritM:
@robinrodricks This PR added two *.cache files inside the folder "_NCrunch_FluentFTP". I don't think they're supposed to be in the repository.

robinrodricks:
Oh I see. I'll delete those from master.

robinrodricks:
Released in [FluentFTP 22](https://www.nuget.org/packages/FluentFTP/22.0.0). Please verify and tell me if it works?

PR #383, opened 2019-03-18
Swap SslStream and BufferedStream

rmja:
This swaps the stream order of SslStream and BufferedStream, so that instead of SslStream->BufferedStream->NetworkStream the order becomes BufferedStream->SslStream -> NetworkStream.

This fixes a regression from #355 where proxied connections with FtpClientHttp11Proxy are unable to connect.

Failing example:
```
var proxyCredentials = new NetworkCredential("username", "password");
var proxyInfo = new ProxyInfo { Host = "1.2.3.4", Port = 5000, Credentials = proxyCredentials };

var ftpCredentials = new NetworkCredential("ftpuser", "ftppassword");

using (var client = new FtpClientHttp11Proxy(proxyInfo) { Host = "ftphost", Port = 990, Credentials = ftpCredentials, EncryptionMode = FtpEncryptionMode.Implicit })
{
    client.Connect();
}
```

The call to `Connect()` times out during read if the the BufferedStream is between the SslStream and the NetworkStream

### DIFF
--- a/FluentFTP/Stream/FtpSocketStream.cs
+++ b/FluentFTP/Stream/FtpSocketStream.cs
@@ -1093,16 +1093,14 @@ namespace FluentFTP {
                 DateTime auth_start;
                 TimeSpan auth_time_total;
 
-	            m_bufStream = new BufferedStream(NetworkStream, 81920);
-
 #if CORE
-				m_sslStream = new SslStream(m_bufStream, true, new RemoteCertificateValidationCallback(
+				m_sslStream = new SslStream(NetworkStream, true, new RemoteCertificateValidationCallback(
 					delegate(object sender, X509Certificate certificate, X509Chain chain, SslPolicyErrors sslPolicyErrors) {
 						return OnValidateCertificate(certificate, chain, sslPolicyErrors);
 					}
 				));
 #else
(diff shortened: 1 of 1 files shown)


rmja:
Sorry about the indentation problems, the file seems to be a mix between spaces and tabs - oh nooo :-)

robinrodricks:
Thanks!

robinrodricks:
Done. Published in FluentFTP [24.0.0](https://www.nuget.org/packages/FluentFTP/24.0.0). 

Please test it and see if everything works as expected.

robinrodricks:
Hi. I need to revert this PR as it is causing https://github.com/robinrodricks/FluentFTP/issues/397 and https://github.com/robinrodricks/FluentFTP/issues/396.

Do you know a way to fix it? If you don't reply shortly I'm reverting this. Thanks.

rmja:
You can revert it, but maybe we should remove the BufferStream as it seems as if either of the variants cause different issues?

robinrodricks:
Okay I'll add an option to enable/disable it and keep it off by default. Perhaps "BoostSSLSpeed" or something?

rmja:
hehe, sounds good to me:) An alternative is the more boring EnableBufferStream

robinrodricks:
Fixed in 25.0.0.

Issue #399, opened 2019-05-17
Separate connect and authenticate or use specific "auth failed" exception


erik-wramner:
**FTP OS:** All

**FTP Server:** All

**Computer OS:** All

The current code combines Connect and Authenticate in a single operation. If credentials are available the login will be performed as part of the connect. As far as I can see there is no way to omit the credentials and do the authentication manually later? Or is there such a method that I missed?

A failed connection should typically be retried as it may have been caused by a transient network error. A failed authentication on the other hand should typically NOT be retried as that may lock out the user and if the password is wrong the first time it will be wrong the second time as well. For that reason it would be nice to either be able to connect in one operation and authenticate in a separate operation (first one with retry, second one without retry) OR to use the existing code as-is but with a clear exception that can be caught on authentication failures. The current code throws FtpCommandException and that is thrown for all failed commands, not just authentication.

I would prefer two different operations (Connect and Authenticate) but a custom exception that is a sub-class of FtpCommandException for backwards compatibility would work too.

erik-wramner:
@robinrodricks please let me know if you need something more for this.

robinrodricks:
Not sure how easy that would be with SSL, connect and authenticate are quite deeply woven. But if you can separate it, fine, but it would break all existing users' code where they only call `Connect()`. So even if you manage to seperate it you'll need a property that controls if authenticate happens automatically with connect or not.

I think the best way is to throw a different exception if the login details are wrong, and a different exception if the connection failed.

erik-wramner:
@robinrodricks, have you seen the push request for this issue? It adds a new exception for "authentication failed" that is a sub-class of the old exception so that it won't break existing code. I wasn't asking you to spend time building my pet feature, I was asking if there is anything more you need from me to accept the push request. I fully realize that your time is limited and that you have other priorities.

robinrodricks:
Hi erik, sorry bout that. I wasn't aware you had made a PR as well. Thanks for that. Normally its better to have such discussions on the PR itself.

robinrodricks:
Thanks. Release in 25.0.1.

https://www.nuget.org/packages/FluentFTP/25.0.1

PR #404, opened 2019-05-31
#399 Use specific "auth failed" exception

erik-wramner:
Issue 399, use a sub-class of the old exception for authentication errors. That way it is possible to catch and not retry them. Opted for quick solution to get this in place fast.


robinrodricks:
Thanks.

PR #412, opened 2019-06-14
Re-added bugfix for issue #288 (Upload Hang with only a couple bytes left)

cw-andrews:


robinrodricks:
Thanks for your work!

PR #429, opened 2019-07-11
Fix empty folder listing with encryption

Mortens4444:
When trying to list an empty folder with encryption on, an exception occurred in the client.


robinrodricks:
Good but how are you sure that every Win32 exception with code = 10053 is a "empty folder exception"?

Mortens4444:
I cannot be sure, but I couldn't find a better way, to handle this error. Do you have any idea, how to do it in a better way?

robinrodricks:
Its complicated. Could you rename the exception to `FtpMissingSocketException` and I'll accept this PR in the spirit of improvement. 

Can you also implement the exception throwing and exception handling in the async version?

Mortens4444:
Oh, we had the same idea, I have already renamed the exception. I will check the async version also.

robinrodricks:
Great, thanks!

Issue #328, opened 2018-08-26
FileExists return false when file exists on server


kfrid:
FileExists is returning false on a file that exists on the server. I use relative path for this.
Use the below code flow, here the exists variable is false but the Bok.xlsx-file exists on the server in the specified working directory.

```
var ftpClient = new FtpClient("url to server", "username", "password");
await ftpClient.ConnectAsync();
await ftpClient.SetWorkingDirectoryAsync("relative path");
var exists = await ftpClient.FileExistsAsync("Bok.xlsx");
```

If I change the last row in the above code example to as below, I can figure out that the file exists by code. Here the exists-variable is true.

```
var ftpClient = new FtpClient("url to server", "username", "password");
await ftpClient.ConnectAsync();
await ftpClient.SetWorkingDirectoryAsync("relative path");
var exists = (await ftpClient.GetListingAsync()).Any(x => x.Name == "Bok.xlsx");
```

LuisSerrano:
Any news about this isuue?
I put a try catch in my code to prevent any error when I copy, rename or delete in the ftp server.

jeremy-addy:
I had the same issue. Here is my workaround:

```c#
if(!ftpClient.Download(memStream, ftpFilePath))
   throw new FtpException($"Failed to download file ftp://{ftpHost}/{ftpFilePath}. Please make sure the file exists.");
```
Another option
```c#
// Verify directory
if(!ftpClient.DirectoryExists(dir))
    throw new FtpException($"FTP directory {dir} does not exist.");

// Get list of files in directory
var listing = ftpClient.GetListing(dir);

// Verify the file is in the directory
if(listing.All(x => x.FullName != ftpFilePath))
    throw new FtpException($"FTP file {ftpFilePath} does not exist.");
```

andresm9:
Problem continues on last version
![image](https://user-images.githubusercontent.com/14965388/52822914-23bcc900-3081-11e9-9f26-5585aac7fea4.png)

KoalaBear84:
Works over here, just took a random FTP. Maybe it is specific to particular FTP server software?

```csharp
using FluentFTP;
using System;

namespace ConsoleApp4
{
    class Program
    {
        static void Main(string[] args)
        {
            Console.WriteLine("Started!");

            FtpClient ftpClient = new FtpClient("ftp.ihe.net");

            Console.WriteLine("Connecting...");
            ftpClient.Connect();
            Console.WriteLine("Connected");

[...]

PR #439, opened 2019-08-05
Let FileExists and FileExistsAsync cater for 550 - SIZE is not allowed in ASCII

Dragonsangel:
Let the methods FileExists- and FileExistsAsync-Methods cater for the Error 550 - "SIZE is not allowed in ASCII" by switching the DataType to Binary.

This could be a potential fix for #328.


robinrodricks:
Sorry to reject this but I applied a variation of your fix.

PR #490, opened 2019-12-02
Fix for non-routable IP regex

andywhitfield:
The 172.2 pattern has an extra \ causing the fallback to host ip not to occur


robinrodricks:
Thank you for spotting this!

andywhitfield:
> 
> 
> Thank you for spotting this!

No probs, thanks for your quick look at this PR - do you have any plan when the next package will be pushed to nuget? We came across this bug connecting to a third-party system and have temporarily used our copy of FluentFTP with this fix to test it confirm it worked - which it did :) But this was just a quick hack and we'd want to use the actual nuget feed version of FluentFTP.

robinrodricks:
Its already published to nuget.

Issue #387, opened 2019-03-27
when upload big file(11g db backup file), timed out trying to read data from the socket stream! 


luchunminglu:
**FTP OS: winserver2016

**FTP Server:filezilla0.9.60

**Computer OS:winserver 2012

it doesn't happen when upload small file.
when upload a big（test on 200mb & 10gb） db backup file, client.UploadFile(localFullFilePath, ftpFullPath) method report a timeout exception.The Error is "Timed out trying to read data from the socket stream! "
the wield thing is the db file is actual uploaded, i check the file size in ftp is right(but not check the file content). 
so i think may be after upload big file, the flutent ftp does something read, which cause this error.

FluentFTP.FtpException: Error while uploading the file to the server. See InnerException for more info. ---> System.TimeoutException: Timed out trying to read data from the socket stream!
   在 FluentFTP.FtpSocketStream.Read(Byte[] buffer, Int32 offset, Int32 count)
   在 FluentFTP.FtpSocketStream.ReadLine(Encoding encoding)
   在 FluentFTP.FtpClient.GetReply()
   在 FluentFTP.FtpClient.UploadFileInternal(Stream fileData, String remotePath, Boolean createRemoteDir, FtpExists existsMode, Boolean fileExists, Boolean fileExistsKnown, IProgress`1 progress)
   --- 内部异常堆栈跟踪的结尾 ---
   在 FluentFTP.FtpClient.UploadFileInternal(Stream fileData, String remotePath, Boolean createRemoteDir, FtpExists existsMode, Boolean fileExists, Boolean fileExistsKnown, IProgress`1 progress)
   在 FluentFTP.FtpClient.UploadFileFromFile(String localPath, String remotePath, Boolean createRemoteDir, FtpExists existsMode, Boolean fileExists, Boolean fileExistsKnown, FtpVerify verifyOptions, IProgress`1 progress)
   在 FluentFTP.FtpClient.UploadFile(String localPath, String remotePath, FtpExists existsMode, Boolean createRemoteDir, FtpVerify verifyOptions, IProgress`1 progress)

(000001)2019/3/28 12:50:03 - scims (192.168.163.1)> OPTS UTF8 ON
(000001)2019/3/28 12:50:03 - scims (192.168.163.1)> 202 UTF8 mode is always enabled. No need to send this command.
(000001)2019/3/28 12:50:03 - scims (192.168.163.1)> SYST
[...]

PR #507, opened 2020-01-06
#387 exhaust NOOP responses also after 226 Transfer complete

aliquid:


robinrodricks:
Awesome! One small question, what if there are no responses? Will it block forever at the ReadStaleData call?

robinrodricks:
Sorry! Just read your comment and you've tested this, which is great. Thanks again.

aliquid:
Yeah no `ReadStaleData/Async` is effectively a no-op if there is no available data. It does not even poll.

Only reason for available data would be superfluous responses, and new code only bothers if it knows it sent any NOOPs.

PR #511, opened 2020-01-16
Added support for the HOST command to be sent after the handshake

dansharpe83:
added a variable in the properties and a command in both Connect and ConnectAsync if the value is set to true


robinrodricks:
Hi, why is this required? Can you explain what issue it fixes or what it
allows you?

dansharpe83:
We need to connect to an FTP site where the HOST command has to be sent directly after the handshake, i think its to get around where the FTP site resolves to an address rather than ip. without this fix we can't connect to the FTP we use

robinrodricks:
Can you send the logs of a successful connection? See the FAQ in how to get
logs.

robinrodricks:
Thanks Dan!

PR #525, opened 2020-02-15
MMD5 Hash Command support

n0ix:
Added support for MMD5 Command.

See https://tools.ietf.org/html/draft-twine-ftpmd5-00 for doc


robinrodricks:
This is great. Thanks!

PR #526, opened 2020-02-15
New FtpRule => FtpRegexRule

n0ix:
I added a simply FtpRule which can check if any of the supplied RegEx pattern matches.

I thought that may be usefull.

### DIFF
--- a/FluentFTP/Rules/FtpFIleNameRegexRule.cs
+++ b/FluentFTP/Rules/FtpFIleNameRegexRule.cs
@@ -10,7 +10,7 @@ namespace FluentFTP.Rules {
 	/// <summary>
 	/// Only accept files that have the given name, or exclude files of a given name.
 	/// </summary>
-	public class FtpRegexRule : FtpRule {
+	public class FtpFIleNameRegexRule : FtpRule {
 
 		/// <summary>
 		/// If true, only items where one of the supplied regex pattern matches are download. If false, items where one of the supplied regex pattern matches are excluded.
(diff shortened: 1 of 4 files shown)


robinrodricks:
Great work! Thanks for this.

robinrodricks:
Hi, some changes:

1. FtpFileNameRegexRule is only meant to check files, correct? So I added a check in `IsAllowed` to ensure only file objects are checked

2. FtpFolderRegexRule is meant to check folder name or folder path? Because the class name suggests you are checking folder name but in the code `item.FullName` is used which is the entire folder path. 

I have fixed both issues, if there is something wrong let me know.

n0ix:
Hi,

thanks for the quick merge and the fixes.

Regarding 3.) - This was intensional, but maybe its better to move this to a sperate Rule (like FtpFolderPathRegexRule  )

robinrodricks:
Released in 31.2.0
  - New: Predefined rules for filtering on file name using regular expressions (thanks [n0ix](/n0ix))
  - New: Predefined rules for filtering on folder name using regular expressions (thanks [n0ix](/n0ix))
  - Fix: Don't calculate ETA and percentage of `FtpProgress` if file size is zero (thanks [Adhara3](/Adhara3))
  - Fix: `GetFilePermissions` should use `GetObjectInfo` instead of `GetListing` to prevent incorrect filepaths

Documented

- https://github.com/robinrodricks/FluentFTP/wiki/Rules#what-kinds-of-rules-are-supported-and-how-do-rules-work
- https://github.com/robinrodricks/FluentFTP/wiki/Class-FtpFileNameRegexRule
- https://github.com/robinrodricks/FluentFTP/wiki/Class-FtpFolderNameRegexRule

Thanks :)

PR #527, opened 2020-02-17
Update FtpClient_FileVerification.cs

Adhara3:
When file size is unknown then leave ETA and Percentage to default values


robinrodricks:
Thanks for the fix!

PR #530, opened 2020-02-18
[WiP] FXP implementation

n0ix:
Hi (again),

i added [File eXchange Protocol (FXP)](https://en.wikipedia.org/wiki/File_eXchange_Protocol) Protocol to the FluentFTP libary.

**Why "Work in Progress" ?**
Cause there is a lot of code i added and i think there may be bugs or things didn't even think of. 
Also i think there is space for optimizations regarding reusing existing methods (cause some methods very similar).  I hope here you can help me, cause you know the whole code of fluent ftp much better then i do.

Also there is missing documentation of the new functions - i will do this later.

- n0ix


robinrodricks:
Ohmigawd! This as awesome! A starting point is really good coz hey people can always contribute fixes as we go along. Its the initial setup that is the hardest to build, good job! I'll review it and let you know.

FluentFTP/Client/FtpClient_FXP.cs
```diff
@@ -0,0 +1,1058 @@
...
+				var transferStarted = DateTime.Now;
+				long lastSize = 0;
+
+				var sourceFXPTransferReply = ftpFxpSession.sourceFtpClient.GetReply();
+				var destinationFXPTransferReply = ftpFxpSession.destinationFtpClient.GetReply();
+
+				while (!sourceFXPTransferReply.Success || !destinationFXPTransferReply.Success)
+				{
+
+					if (remoteClient.EnableThreadSafeDataConnections)
```
robinrodricks: Why is it required to check EnableThreadSafeDataConnections here?

FluentFTP/Client/FtpClient_FXP.cs
```diff
@@ -0,0 +1,1058 @@
...
+					{
+						FtpTrace.Write(FtpTraceLevel.Info, "reporting progress");
+						// send progress reports
+						if (progress != null && fileSize != -1)
+						{
+							offset = remoteClient.GetFileSize(remotePath);
```
robinrodricks: Here you are getting the file size from the remote server how often?

FluentFTP/Client/FtpClient_FXP.cs
```diff
@@ -0,0 +1,1058 @@
...
+
+			// get all the files in the local directory
+			var fileListing = GetListing(sourceFolder, FtpListOption.Recursive).Where(x => x.Type == FtpFileSystemObjectType.File).Select(x => x.FullName).ToArray();
+
+			// loop thru each file and transfer it
+			var filesToUpload = GetFilesToTransfer(sourceFolder, remoteFolder, rules, results, shouldExist, fileListing);
+			TransferServerFiles(filesToUpload, remoteClient, existsMode, verifyOptions, progress, remoteListing);
+
+			// delete the extra remote files if in mirror mode and the directory was pre-existing
+			// DeleteExtraServerFiles(mode, shouldExist, remoteListing);
```
robinrodricks: A bit of code duplication here, but still great considering most of the inner functions are reused.

FluentFTP/Client/FtpClient_FXP.cs
```diff
@@ -0,0 +1,1058 @@
...
+
+				// record that this file should exist
+				shouldExist.Add(remoteFile.ToLowerInvariant(), true);
+
+				// absorb errors
+				filesToTransfer.Add(result);
+			}
+
+			return filesToTransfer;
+		}
```
robinrodricks: Isn't this total code duplication? Is there a way to make a common method so the Upload/Download Directory methods can reuse this with FXP? or is it too many changes?

FluentFTP/Client/FtpClient_FXP.cs
```diff
@@ -0,0 +1,1058 @@
...
+			}
+
+			if (!this.IsConnected)
+			{
+				throw new FluentFTP.FtpException("The source FXP FtpClient must be open and connected before a transfer between servers can be intitiated");
+			}
+
+			if (!await FileExistsAsync(sourcePath, token)){
+				throw new FluentFTP.FtpException(string.Format("Source File {0} cannot be found or does not exists!", sourcePath));
+			}
```
robinrodricks: This is a good, safe check, but is it needed again at this point?

FluentFTP/Client/FtpClient_Properties.cs
```diff
@@ -795,6 +795,17 @@ public partial class FtpClient : IDisposable {
 			set => m_DownloadDataType = value;
 		}
 
+		public FtpDataType m_FXPDataType = FtpDataType.Binary;
+
+		/// <summary>
+		/// Controls if the high-level API transfer (FXP) files in Binary or ASCII mode.
```
robinrodricks: Change comment:

Controls if the FXP server-to-server file transfer API uses Binary or ASCII mode.

FluentFTP/Client/FtpClient_FXP.cs
n0ix: Cause we can't do this when we have only one connection, the one which is open is already used for the FXP Transfer and can't answer any other commands. Maybe i got this wrong, but a ftp client should/can be onlycloned if ThreadSafeDataConnections are enabled. In OpenPassiveFXPConnection i clone the connection if ThreadSafeDataConnections is enabled so we can later track progress.

FluentFTP/Client/FtpClient_FXP.cs
n0ix: Very often, and i know this is ugly - but i did find any other possibility to track progress of the FXP Transfer (cause we have no stream here). May be we can add another Property which enables this?

FluentFTP/Client/FtpClient_FXP.cs
n0ix: Yes i agree here is definitly space to deduplicate code. The problem are the "directions" in FXP. The Upload und Download methods i can't reuse cause the assume there is a a local path which we don't have when we use fxp.

FluentFTP/Client/FtpClient_FXP.cs
n0ix: as already mentioned before, reuse of this methods is difficult cause the assume a local path (as source or as destination) which don't have here.

FluentFTP/Client/FtpClient_FXP.cs
n0ix: Yeah cause it's never have been checked before or i'am wrong?

robinrodricks:
Thank you for the effort!

robinrodricks:
Thanks for this fantastic contribution. I've cleaned up the code to remove duplicate sections and also improved the comments. I noticed how you stayed within the guidelines by using the FolderUpload/FolderDownload logic as a template. I also fixed some logging (LogFunc) and code logic. 

New methods that are reused between Folder Upload and Folder Transfer (FXP):

- `CanUploadFile`
- `CreateSubDirectories`

New methods that are reused in File Transfer (FXP):

- `VerifyTransferFileParams`

robinrodricks:
Released in 31.3.0!


  - New: `TransferFile` and `TransferDirectory` methods to transfer files from server to server (thanks [n0ix](/n0ix))
  - New: FXP (File Transfer Protocol) implementation to support direct server-to-server transfers (thanks [n0ix](/n0ix))

robinrodricks:
Added an icon and a note in the top features on the [readme](https://github.com/robinrodricks/FluentFTP). Will do documentation soon.

n0ix:
Awesome! Thanks for the code cleanup an the quick release!

robinrodricks:
Welcome. When is the FtpFxpSession destroyed? We might need to destroy the cloned connection to the remote server after transfer.

robinrodricks:
Documented! - https://github.com/robinrodricks/FluentFTP/wiki/FXP-Server-To-Server

n0ix:
> 
> 
> Welcome. When is the FtpFxpSession destroyed? We might need to destroy the cloned connection to the remote server after transfer. We probably don't need to destroy the Source server connection, am I right? Because if that is the main connection then why should we destroy it in `FtpFxpSession.Dispose`?

Hm, yeah we should clean up this, but only if we **cloned** the connections. Else we need to leave as them are cause the user created them manually and won't expect that one the connections getting disposed. Is there an property on cloned connections so we can identify them? Then it should be easy to dispose the `FxpSession `in the finally block of `TransferFile `/ `TransferDirectory`

robinrodricks:
I've tried to handle this, can you check the latest trunk and let me know if it works?

Also, can we reuse the same FXP session for an entire folder instead of creating one per file transfer? I'll do the refactoring, just tell me if its possible and recommended.

robinrodricks:
Newly released in 32.2.0:

  - New: `GetChecksum` allows you to specify a hash algorithm to be run on the server if supported
  - New: FXP file transfer now validates the file using the first mutually supported algorithm

Can you check if the FXP logic for checking mutually supported algorithm is correct? It is based on the new feature in GetChecksum that allows you to execute a specific hash on the server rather than executing any random hash.

robinrodricks:
Please see - https://github.com/robinrodricks/FluentFTP/issues/546

PR #548, opened 2020-03-16
Processed bytes progress

Adhara3:
Sometimes filesize information is not available and thus no progress info is actually available. I added the stream position to provide at least an absolute reference in progress


robinrodricks:
This is great. Thank you!

Issue #561, opened 2020-04-12
UploadDirectory should consider the FtpRules in Mirror Mode on Server side


philippjenni:
Motiviation
--------------

I want to mirror application files (WebApp) to a webserver over ftp which has application data on the server that must not be deleted on update.

Description
--------------
If you will mirror a local directory to the server with FtpRules to ignore Directories, ignored directories should not be deleted on the server.

robinrodricks:
I don't understand this. Can you explain. Currently we have full support for Mirror/Update modes on download and upload directory. What are you trying to implement here?

philippjenni:
Here an example as use case

I have an ASP.NET application which has an App_Data Folder. The application on server stores data in this App_Data folder.

I have built a new Release of this application, that have an empty App_Data Folder. Now when i mirror the local folder to the server, the App_Data folder would be deleted, because the local directory it is empty. Now i have create an FtpRule that excludes the App_Data folder to upload. Files in this folder are no more uploaded, but the folder on the server is deleted by the uploading process.

It is a must,that the app_data folder on the server is not changed

robinrodricks:
I think just a simple option that says DeleteFilteredFiles would be enough. It would control if the filtered out files are ignored or deleted.

philippjenni:
> I think just a simple option that says DeleteFilteredFiles would be enough. It would control if the filtered out files are ignored or deleted. So this way you can add a rule to filter out the App_Data folder and it will not be deleted. It will be ignored.

Yes that be an good option ... if the option is not set it should be the current behavoir ... i can implement this, but should i change UploadDirectory's with a new parameter. I can't find a parameter where it makes sense to expand the option.

philippjenni:
![Synchronisation Tasks](https://user-images.githubusercontent.com/579302/79630482-f589bc00-8151-11ea-8cc5-2a242a1ed797.png)

OK. But this should implement all the rules in the graph:

- Excluded files and folders should not be uploaded (App_Data, LogData.mdf)
- Excluded files and folders should not be deleted on remote side (LogData.ldf)
- Additional files and folders on local side should be uploaded (ImageMedium.png)
- Additional files and folders on remote side should be deleted (ImageSmall.png, file1.tmp, file2.tmp)
- Changed files should be overwritten (All yellow marked files)

This is what my pull request implemented

robinrodricks:
[...]

PR #563, opened 2020-04-12
#561 Implementation FtpRules in DeleteExtraServerFiles with Mirror Mode

philippjenni:
- Extend EditorConfig with Bracket Rule for C#
- Implenmentation of FtpRules in DeleteExtraServerFiles to implemenet #561
- Implenmentation of FtpRules in DeleteExtraLocalFiles to implemenet #564

_Breaking Change:_
- Method "UploadDirectory" and "UploadDirectory" has a changed signature because an new property was added
- Method "DownloadDirectory" and "DownloadDirectoryAsync" has a changed signature because an new property was added

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,2 +1,3 @@
 [*.cs]
 indent_style = tab
+csharp_new_line_before_open_brace = none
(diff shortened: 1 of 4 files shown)


robinrodricks:
Thanks for this. I'll check it out and let you know.

robinrodricks:
I have added questions to your issues.

robinrodricks:
I will implement the property to control whether excluded files should be deleted. Thus adding rules on source & target are not required. Thanks for your PR but this is not required.

PR #586, opened 2020-05-25
Clear custom parser when removing parser or clearing all parsers

### LINKED ISSUE #585: Custom parser not cleared correctly
**FluentFTP Version:** 32.4.1

We are using FluentFTP to read from different sources. Those sources come from a config file, each config instance has a property 'useCustomParser'.
When this property is set to ```true``` we add the custom parser as follow: 

```FtpListParser.AddParser(parser);```

When the next item in the config does not use a custom parser we remove it as follow:

```FtpListParser.RemoveParser(parser)```

The method ```RemoveParser``` does remove the custom parser from the list ```FtpListParser.m_parsers```  but it does not remove the current parser from the ```FtpListParser.m_customParser``` field,
which is set in the ```FtpListParser.AddParser``` method. Even ```FtpListParser.ClearParsers``` does not remove the current parser from the ```FtpListParser.m_customParser``` field.

This sounds like a bug to me. If you can confirm that this is not desired behaviour, i will provide a pull request to fix this.

Looking forward to your reaction, thanks!

rubenhuisman:
Closes #585

### DIFF
--- a/FluentFTP/Helpers/FtpListParser.cs
+++ b/FluentFTP/Helpers/FtpListParser.cs
@@ -276,6 +276,7 @@ namespace FluentFTP {
 				}
 
 				m_parsers.Clear();
+				m_customParser = null;
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


robinrodricks:
Thanks!

robinrodricks:
Thanks!

Released in https://www.nuget.org/packages/FluentFTP/32.4.3

Issue #493, opened 2019-12-10
Allow skipping the certificate revocation check


lscorcia:
Hi, I have an application that needs to talk to a FTPS server that is currently using a certificate signed by the Actalis G2 intermediate certificate (https://www.actalis.it/documenti-it/actalisorganizationvalidatedservercag2.aspx). This certificate unfortunately uses LDAP to distribute the CRL and since my application is sitting behind a corporate http/s proxy FluentFTP complains that it cannot validate the certificate. It would be great if I could maintain the certificate chain validation, but skip the certificate revocation check.

I have found issue #300 that seems to have identified the very same problem, and here it is said it has been already addressed with a specific parameter of the FtpClient class, but looking at the source it seems to me that the actual code is still forcing the validation check:

https://github.com/robinrodricks/FluentFTP/blob/2845a26bfb361d5586c0de078e06516b159582eb/FluentFTP/Stream/FtpSocketStream.cs#L1068-L1071

(the problem is the last hardcoded `true` value in the calls). 

Could you please consider adding the `SkipCertificateRevocationCheck` property to `FtpClient`?
Thanks in advance

lscorcia:
If such a property should not be desiderable for any reason, I found that the following code seems to be a suitable workaround:


```
client.ValidateCertificate += OnClientValidateCertificate;
...
private void OnClientValidateCertificate(FtpClient control, FtpSslValidationEventArgs args)
{
    // If the remote certificate has been correctly validated from the system,
    // then it's fine for us.
    if (args.Accept) 
        return;

    // Perform basic validation using the cached CRL entries only and write a note in the log
    args.Chain.ChainPolicy.RevocationMode = X509RevocationMode.Offline;
    args.Chain.ChainPolicy.VerificationFlags =
        X509VerificationFlags.IgnoreEndRevocationUnknown |
        X509VerificationFlags.IgnoreCertificateAuthorityRevocationUnknown |
        X509VerificationFlags.IgnoreRootRevocationUnknown;
    args.Accept = args.Chain.Build(new X509Certificate2(args.Certificate));
    log.WarnFormat("System Certificate Validation failed (Policy Errors: {0}). Trying without checking for revocation...: {1}", 
        args.PolicyErrors, args.Accept ? "OK": "Failed");
}
```

robinrodricks:
Done!

New: `ValidateCertificateRevocation` property to control if certificate revocation is checked.

https://www.nuget.org/packages/FluentFTP/28.0.5

PR #603, opened 2020-07-29
Fix 'ValidateCertificateRevocation' property not being used in async version

kolorotur:
Use of hard-coded 'checkCertificateRevocation' parameter value to SslStream's 'AuthenticateAsClientAsync' method was replaced with the 'ValidateCertificateRevocation' property as a fix for #493, but async version of the code was omitted.


robinrodricks:
Thanks for this!

Issue #283, opened 2018-03-23
calling UploadFile for large file got exception with message "You cannot modify the position of a FtpDataStream. This property is updated as data is read or written to the stream."


amidofu:
**FTP OS:** not sure, probably Unix, the server is provided by GoDaddy: https://en.wikipedia.org/wiki/GoDaddy

**FTP Server:** not sure

**Computer OS:** ? Windows 7

<write details about your bug report / feature request here>
I try to upload a big file (about 1.5GB) by FtpClient.UploadFile, but it always throws exception with the message: "You cannot modify the position of a FtpDataStream. This property is updated as data is read or written to the stream."
Tried to use FileZilla to upload such large file and it worked fine.

Small files can be uploaded (tried with files that are less than 500KB) normally by FtpClient.UploadFile

**Logs :**

[fluent_ftp_log_file.txt](https://github.com/robinrodricks/FluentFTP/files/1840777/fluent_ftp_log_file.txt)

```
<paste logs here but DO NOT delete the lines above and below this line>
```

robinrodricks:
Sounds like a part of our uploading implementation is not supported by your server. We'll have to modify it accordingly.

CHMaagaard:
I have encountered the same problem using a FileZilla server.

klesteb:
I have too, In my case the server had run out of disk space.

sanjitmisra:
Does this library return the 4xx codes in case of transient negative completions? I am encountering an issue where the upstream server has run out of disk but I am getting a generic error message 

"You cannot modify the position of a FtpDataStream. This property is updated as data is read or written to the stream"

Is this bug being actively worked upon?

robinrodricks:
Hi, no this bug is not actively being worked upon. We have insufficient
info to fix something. You are free to submit a PR or help us understand
how to send specialised 4xx code.

Trox085:
I too am encountering this issue while being connected to FileZilla Server, but the error is not limited to just that server.  It doesn't happen all the time, but I have tracked down the top-level reason.

FluentFTP/Client/FtpClient_HighLevelUpload.cs line 1003 shows
```
							upStream = OpenAppend(remotePath, UploadDataType, true);
							upStream.Position = offset;
```

[...]

PR #643, opened 2020-11-09
Fix Resume when the connection drops

pradu71:
While resuming a FTP transfer, we can't rely on our knowledge of the actual offset really written on the remote server.
Thus I inverted the logic, when opening a new connection for resume (in ResumeUpload) we retrieve the actual remote offset, so we can reposition our local stream to the effective position.

this fixes for me issue #283


robinrodricks:
Thank you for this.

robinrodricks:
Released! https://www.nuget.org/packages/FluentFTP/33.0.3

PR #645, opened 2020-11-12
Set default value for restartPosition in Download method on IFtpClient to match FtpClient method

AdamLewisGMSL:
Does what it says on the tin. FtpClient.Download has a default value, IFtpClient.Download does not


robinrodricks:
Thank you for this!

robinrodricks:
Released! https://www.nuget.org/packages/FluentFTP/33.0.3

PR #654, opened 2020-11-30
added fallback to regular passive if the response is not Extended

hornell:
Debugging the following log message:
"Command: EPSV
 Response: 227 Entering Passive Mode( **, **, **, **, 143,225).
"

Which resulted in an exception "Failed to get the ESPV port from ..."

Determined that the regex being used by GetEnhancedPassivePort was not working with that response. Which is no surprise, as the response is not a Enhanced(Expanded) passive response. Unsure why the FTP server responds this way, but having the function use GetPassivePort in this case results in the response port being parsed.


robinrodricks:
Thank you for this!

PR #655, opened 2020-11-30
fix ReadStaleData causing infinite loop

hornell:
Encountered an infinite loop when connecting.

Debugging shows that the ReadStaleData function was closing the connection, resulting in a connection attempt being made (from Execute()), which repeats endlessly.

For the particular FTP server I was connecting to it appears that the only stale data was the newline character.

ReadStaleData actually already trims off the newline characters when building the string to log out, so we can easily check if that string in empty in order to determine if we really need to close the stream for that "stale" data.

Once that change was made, connections to the server were successful.


robinrodricks:
Thank you for this!

PR #679, opened 2021-02-14
Update FtpClient_FileProperties.cs

YSternlicht:
Simple typo (I think)


YSternlicht (commented):
Says Gets the modified time instead of changes

Issue #713, opened 2021-04-22
Missing capability to bind to a specific local interface


daviddenis-stx:
Hello,

It's more a RFE than an issue.

The idea would be to be able to use one of the many local public IP addresses one can have (say bare metal server with several publicly routable IP addresses, either/both v4 or v6) as source address to connect to remote FTP hosts. Meaning that those FTP hosts would see you coming from the interface you choose (instead of the interface holding the default route that is selected when you create a Socket). 

The only thing required in order to achieving that is calling the Bind() method before connecting (the very same way you would do before a Listen() on server side)

For example lftp has parameters for that (see http://manpages.org/lftp)
- ftp:port-ipv4/6
- net:socket-bind-ipv4/6

I have quickly prototyped a change doing just that here:
- https://github.com/daviddenis-stx/FluentFTP/commit/6075f6266456a73979c1a329cc2e62f300cf74e9

Tested on a local proftpd server. Directory listings and downloads seem to work fine (not tested active mode though)

Maybe you would want to consider such a functionality for future releases.

Best

robinrodricks:
It looks good, can you submit a PR?

We already have `client.AddressResolver` which helps to resolve the local address. Is your change related or completely different to that?

daviddenis-stx:
Hello Robin,

Didn’t see about the DNS part already present. I had to use this to test on my side, it seems that if you have IPv6 addresses (ex: a few Zerotier networks) and your default route only has Ipv4 (most of us) asking .NET about you own IP addresses will omit those. Not sure why really.
The patch I made is probably not PR ready (though, I made my own nuget with it and it’s running production, seems to work correctly on two servers each of them having 3 different public IPv4 addresses).

I probably would need to at least remove that part. Also on my setup I modified csprojs to be able to build as the very old net20 and such I can’t build those anymore 😊

Let me know what would make your community happy and I’ll make the PR 😊

D

robinrodricks:
Create a PR with whatever new code changes you have done. If there is any
change in the default behaviour that might affect other users just let me
know and I'll review it.

I don't think I need the changes in csproj within the PR, because you can
use conditional compiling to target certain platforms, like "#if CORE" to
only complete the code on .net core, etc. We have constants for all
platforms and groups.

Thanks.

daviddenis-stx:
Hello again 😊

Tell you what, I’ll make a PR containing only *.cs and we’ll see if the SF manages to build/test it 😊

D

[...]

PR #719, opened 2021-04-30
Locally bind to specific address via new LocalIpAddress property

daviddenis-stx:
As promised in discussion in https://github.com/robinrodricks/FluentFTP/issues/713 

A PR with only *.cs files

Not sure it will build with all targets (net20 and such, who still uses that anyway ?)

### DIFF
--- a/FluentFTP/Client/FtpClient_Properties.cs
+++ b/FluentFTP/Client/FtpClient_Properties.cs
@@ -1042,6 +1042,20 @@ namespace FluentFTP {
 			set => m_SendHostDomain = value;
 		}
 
+		private IPAddress m_LocalIpAddress;
+		/// <summary>
+		/// Allow to bind sockets to a particular local IP/interface.
+		/// Useful if you have several usable public IP addresses and want to use a particular one.
+		/// </summary>
+		public IPAddress LocalIpAddress
+		{
+			get => m_LocalIpAddress;
+			set => m_LocalIpAddress = value;
+		}
+
+		public IPEndPoint LocalEndPoint => this.m_stream.LocalEndPoint;
+		public IPEndPoint RemoteEndPoint => this.m_stream.RemoteEndPoint;
+
 		// ADD PROPERTIES THAT NEED TO BE CLONED INTO
 		// FtpClient.CloneConnection()
 	}
--- a/FluentFTP/Streams/FtpSocketStream.cs
+++ b/FluentFTP/Streams/FtpSocketStream.cs
@@ -9,22 +9,28 @@ using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
 using System.Net;
+using System.Net.NetworkInformation;
+
+using FluentFTP.Helpers;
 using FluentFTP.Exceptions;
 #if CORE || NET45
 using System.Threading.Tasks;
-
 #endif
 
 namespace FluentFTP {
+
+
 	/// <summary>
 	/// Stream class used for talking. Used by FtpClient, extended by FtpDataStream
 	/// </summary>
 	public class FtpSocketStream : Stream, IDisposable {
 		private SslProtocols m_SslProtocols;
+		private IPAddress m_LocalIpAddress;
 		public FtpClient Client;
 
-		public FtpSocketStream(SslProtocols defaultSslProtocols) {
+		public FtpSocketStream(SslProtocols defaultSslProtocols, IPAddress localIpAddress) {
 			m_SslProtocols = defaultSslProtocols;
+			m_LocalIpAddress = localIpAddress;
 		}
 
 		/// <summary>
@@ -815,7 +821,7 @@ namespace FluentFTP {
 		/// <param name="ipVersions">Internet Protocol versions to support during the connection phase</param>
 		public void Connect(string host, int port, FtpIpVersion ipVersions) {
 #if CORE
-			IPAddress[] addresses = Dns.GetHostAddressesAsync(host).Result;
+			IPAddress[] addresses = Dns.GetHostAddresses(host);
 #else
 			IAsyncResult ar = null;
 			var addresses = Dns.GetHostAddresses(host);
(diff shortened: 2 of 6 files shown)


robinrodricks:
Awesome, I'll check it out!

FluentFTP/Streams/FtpSocketStream.cs
```diff
@@ -815,7 +821,7 @@ public class FtpSocketStream : Stream, IDisposable {
 		/// <param name="ipVersions">Internet Protocol versions to support during the connection phase</param>
 		public void Connect(string host, int port, FtpIpVersion ipVersions) {
 #if CORE
-			IPAddress[] addresses = Dns.GetHostAddressesAsync(host).Result;
```
robinrodricks, line 818: Is this change needed?

FluentFTP/Client/FtpClient_Properties.cs
```diff
@@ -1042,6 +1042,20 @@ public partial class FtpClient : IDisposable {
...
+		/// Allow to bind sockets to a particular local IP/interface.
+		/// Useful if you have several usable public IP addresses and want to use a particular one.
+		/// </summary>
+		public IPAddress LocalIpAddress
+		{
+			get => m_LocalIpAddress;
+			set => m_LocalIpAddress = value;
+		}
+
+		public IPEndPoint LocalEndPoint => this.m_stream.LocalEndPoint;
```
robinrodricks, line 1056: Read only or editable?

FluentFTP/Streams/FtpSocketStream.cs
```diff
@@ -861,10 +867,14 @@ public class FtpSocketStream : Stream, IDisposable {
 				}
 
 				m_socket = new Socket(addresses[i].AddressFamily, SocketType.Stream, ProtocolType.Tcp);
+				BindLocalIfNecessary();
+
 #if CORE
+
+
 				var args = new SocketAsyncEventArgs {
-					RemoteEndPoint = new IPEndPoint(addresses[i], port)
```
robinrodricks, line 866: Is this change needed?

robinrodricks (changes requested):
Great work with the example code and helpers. However some changes to the `FtpSocketStream` class could be reverted if possible so that other users are not impacted. If not possible, let me know. Thanks!

FluentFTP/Streams/FtpSocketStream.cs
daviddenis-stx: Strange, i still have this line in my local code, I must have missed something while rebasing

FluentFTP/Streams/FtpSocketStream.cs
daviddenis-stx: Same here, I have that locally, unchanged

daviddenis-stx: looks like whitespace problems

FluentFTP/Client/FtpClient_Properties.cs
daviddenis-stx: It shouldn't be changed if the connection state is already "connected" (it would have no effect since we call BindLocallyIfNeeded() from the Connect() and ConnectAsync() method)

FluentFTP/Streams/FtpSocketStream.cs
daviddenis-stx: Hmmm, I remember, just checked if the async version was more prone to list IPv6 addresses (it wasn't). 

We probably need to use something else here for proper IPv6 resolving support (you told me about some co-routine already present in the code, capable of doing lookups). On my first patch I used DnsClient which resolves correctly.

If GetHostAddresses() won't return IPv6 addresses (that happens when the default route is IPv4). You won't be able to connect to hosts that have also IPv6 addresses. Which is a problem if you bind locally to an IPv6 interface that is capable to reach that IPv6 addresses host also has. It ends up with locally binding to IPv6 interface and trying to connect to an IPv4 host, that can't work.

robinrodricks: So can I revert this or is it needed to make your functionality work? If it is needed I can use a flag (option) to let users enable/disable this? Or are you sure it won't break the existing functionality?

daviddenis-stx:
This does the same stuff [thing.DoSomethingAsync().Result and thing.DoSomething()]
It’s safe to revert

FluentFTP/Streams/FtpSocketStream.cs
robinrodricks: Understood

FluentFTP/Client/FtpClient_Properties.cs
robinrodricks: Thanks

FluentFTP/Streams/FtpSocketStream.cs
robinrodricks: Understood

robinrodricks:
Great, thanks.

robinrodricks:
Thanks for this!

daviddenis-stx:
My pleasure 😊

daviddenis-stx:
It's probable that without the proper DNS resolving for IPv6, people will have to use the IPv6 addresses both for local and remote :) Can you point me to the place where you do DNS resolving in the current code ? I'll try it

robinrodricks:
These are all I can find:
1. https://github.com/robinrodricks/FluentFTP/blob/master/FluentFTP/Streams/FtpSocketStream.cs#L824
2. https://github.com/robinrodricks/FluentFTP/blob/master/FluentFTP/Streams/FtpSocketStream.cs#L939

robinrodricks:
Added to docs.

https://github.com/robinrodricks/FluentFTP/wiki/FTP-Connection#socket-settings

daviddenis-stx:
Thanks a lot Robin 😊 Will give it a try. Keep you posted.

PR #732, opened 2021-06-28
Bugfix for file uploading via IBM zOS FTP.

arafuls:
There is an issue when attempting to upload files to IBM zOS as the remote path does not properly calculate. It ignores the period naming convention for the Partitioned Data Set (PDS) and single quotes surrounding the path.

- Added a conditional before adding a trailing slash to directory paths.
Instead, if the ServerType is IBM zOS then it will add a trailing .' if not found.

- Added a conditional when calculating the remote path on IBM zOS.
The file name will now be appended to the remote dir before the trailing single quote.

### DIFF
--- a/FluentFTP/Client/FtpClient_FileUpload.cs
+++ b/FluentFTP/Client/FtpClient_FileUpload.cs
@@ -19,8 +19,10 @@ using System.Threading.Tasks;
 
 #endif
 
-namespace FluentFTP {
-	public partial class FtpClient : IDisposable {
+namespace FluentFTP
+{
+	public partial class FtpClient : IDisposable
+	{
 		#region Upload Multiple Files
 
 		/// <summary>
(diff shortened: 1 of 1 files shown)


arafuls:
Found more issues to be fixed.

ederpf:
Hi @arafuls!!

We are facing the same issue with IBM and for us is really urgent to have a release with this issue fixed. Do you know when you will be able to upload the changes?

Thanks a lot for your time trying to discover the problem.

arafuls:
@ederpf 

I am actively working on this issue but found multiple parts of the code which might face to be a problem.. Hence why I closed this PR. 

Currently, I am seeing if I can come up with a solution externally from FluentFTP, hopefully using class method overriding, but if not I may change the FluentFTP code directly and hope for an approved merge from the maintainer. Not sure how long a solution will take.

PR #741, opened 2021-07-20
Fixed IBM zOS MVS File uploading

arafuls:
An issue with my previous implementation is that it is assuming you are always trying to upload to a PO (Physical Organization) and would throw the error "Unable to upload, use MVS naming convention" when trying to upload to other locations.. So I only fixed half the problem.

By just sending the filename into the STOR command rather than with the full file path, the STOR command is intelligent enough to determine if the target location is a Physical Organization and append the (filename) convention to upload to a Partitioned Data Set. It also can append the .filename convention if it is a Physical Sequential location and upload it like a normal file rather than a PDS.

### DIFF
--- a/FluentFTP/Client/FtpClient_FileUpload.cs
+++ b/FluentFTP/Client/FtpClient_FileUpload.cs
@@ -97,17 +97,10 @@ namespace FluentFTP {
 				var remotePath = "";
 
 				if (isPDS) {
-					remotePath = remoteDir;
-
-					// make sure path doesn't end with a dot before appending "(filename)"
-					if (remotePath[remotePath.Length - 2] == '.') {
-						remotePath = remotePath.Remove(remotePath.Length - 2, 1);
-					}
-
-					// Append "(filename)" before closing single quote
-					remotePath = remotePath.Insert(remotePath.Length - 1, "(");
-					remotePath = remotePath.Insert(remotePath.Length - 1, fileName);
-					remotePath = remotePath.Insert(remotePath.Length - 1, ")");
+					// STOR cmd is intelligent enough to determine the full path
+					// and if it needs to append the ".filename" or "(filename)"
+					// addition to the remote path internally for Dsorgs.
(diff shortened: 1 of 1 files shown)


fdrobidoux:
I'd love to have this fix merged, we need it for work at my company. Could this be merged soon, please? Thanks!

arafuls:
@robinrodricks If merged, another quick NuGet release would also be greatly appreciated!

arafuls:
Still no insight on issues with downloading, FYI. But I'm still investigating.

fdrobidoux:
@arafuls I'm having something unintended in the generation of the path for the FTP file.
I'm creating the FtpClient like `new FtpClient("A.LONG.HOST.NAME.CA", 21, "USERNAME", "PASSWORD")`
And then I do `await ftpClient.UploadAsync(byteData, "AAAAA.ALLUSER.BBBB.CCCC.INPUT")` and the resulting path by the time I reach `GetAbsolutePathAsync(path.GetFtpPath(), token)` the result ends up being `"'USERNAME.'/AAAAA.ALLUSER.BBBB.CCCC.INPUT"` which might be part of the problem. 

Did I forget to do something?

Thanks!

arafuls:
@fdrobidoux As discussed, I have not modified the other upload methods.

I will look into these when I am able to as they are not my primary efforts. If you decide to look into them yourself, look at how it is calculating the remote path as that is most likely where is issue lies.

robinrodricks:
Okay I will do a release tomorrow.

arafuls:
@robinrodricks Any update?

arafuls:
@robinrodricks I'm sorry for bothering you so much the past week but thank you for your help! Another quick NuGet release would also be greatly appreciated!

robinrodricks:
No problem @arafuls

robinrodricks:
Published.

https://www.nuget.org/packages/FluentFTP/34.0.2

PR #746, opened 2021-08-06
append method was not appending.

cgunaydin:
New chunk's position and offset reset.
Stream length sum with chunk size and uploaded file size.


robinrodricks:
Thanks for this!

Issue #677, opened 2021-02-11
On IBM zOS using a DATASET file system, the FILE command always returns with an empty list


cmdpdl:
An IBM server described by the library as follows
220-**FTPD1 IBM FTP CS V2R3** at XXXXXXXXXXXXX, 16:48:12 on 2021-02-11.
Response: 220 Connection will close if idle for more than 5 minutes.
Status:   **Detected FTP server: IBMzOSFTP**
....
MVS is the operating system of this server. FTP Server is running on z/OS.

When I try to list the contents of a directory I always get an empty list, see the library log for more details:

```
# Connect()
Status:   Connecting to 10.1.2.83:21
Response: 220-FTPD1 IBM FTP CS V2R3 at XXX.XXXX.COM, 17:41:19 on 2021-02-11.
Response: 220 Connection will close if idle for more than 5 minutes.
Status:   Detected FTP server: IBMzOSFTP
Command:  USER DEVCDP
Response: 331 Send password please.
Command:  PASS ***
Response: 230 DEVCDP is logged on.  Working directory is "DEVCDP.".
Command:  FEAT
Response: 211- Extensions supported
Response: AUTH TLS
Response: PBSZ
Response: PROT
Response: 211 End
Status:   Text encoding: System.Text.ASCIIEncoding
Command:  SYST
Response: 215 MVS is the operating system of this server. FTP Server is running on z/OS.
Command:  LIST ['DEVCDP.*']
Response: 550 No data sets found.
Command:  CWD 'DEVCDP.'
Response: 250 "DEVCDP." is the working directory name prefix.
Command:  PWD
Response: 257 "'DEVCDP.'" is working directory.
# GetWorkingDirectory()
Command:  PWD
Response: 257 "'DEVCDP.'" is working directory.
# GetListing("", SizeModify)
# GetWorkingDirectory()
Command:  PWD
Response: 257 "'DEVCDP.'" is working directory.
Command:  TYPE I
Response: 200 Representation type is Image
# OpenActiveDataStream(AutoActive, "LIST 'DEVCDP.'", 0)
Command:  EPRT |1|10.1.3.181|52927|
Response: 200 EPRT request OK
Command:  LIST 'DEVCDP.'
Response: 550 No data sets found.


# Connect()
Status:   Connecting to X.X.X.X:21
Response: 220-FTPD1 IBM FTP CS V2R3 at XXX.XXXX.COM, 17:41:19 on 2021-02-11.
[...]

PR #764, opened 2021-10-12
Support z/OS and OS/400 GetListing Function

FanDjango:
Robin,

Here is my current take on getting GetListing(...) to work with z/OS (and OS/400, which is unchanged and just separated out from the "generic" IBM label).

Some other functions of FluentFTP will still **not** work, especially those that rely on path manipulation etc., unless you are moving in the z/OS USS realm - this is sufficiently unix-like.

I see some places where others have already fixed upload and download for native z/OS dataset names (like "`'HLQ.QUAL1.QUAL2'`"),  those very valuable fixes handle the case of a fully complete dataset specification and I have been happily uploading and downloading thanks to those.

To show the user a list of items to choose from, this PR concerns getting a sensible listing of the available dataset / files under the current directory.

I wanted to put this out there for any that might need this urgently and also would like play with it and report anything that might still need to be done.

Here is a summary of what was changed:
Separate z/OS and OS/400 (two different animals)
- Make new components for IBM z/OS, IBM OS/400 (take the code of "IBM"), remove IBM
- Fix the Welcome Message of OS/400
- Use the Welcome Message of "IBM" in IBM z/OS
- Reflect these changes in the Enums, the Servers, etc.
- Create Directory Listing Parser for zOS native realm. Simply call the standard Unix Parser for the USS Realm,  
- Minor tweaks to System and Parser recognition

As soon as you move to the native z/OS realm, you will need to "CWD" explicitly to the level where you want to do use `GetListing()` and you need to add  `FtpListOption.NoPath`. **Edit:** But I have also made a modification so that this is not needed after all.

Note that `GetListing()` _sort of_ worked already before, as you can see from the discussion in #677.

Now, in the native realm, `isDir` is set depending on **PO**, the date format is parsed and an attempt is made to "guess" the size.

Here is some example code for both realms:

**Native z/OS realm:**
```

string zosPath = "'SYS1'"; // you can also use 'SYS1.', both work

FTP_Sess.Execute("CWD " + zosPath);
FtpListItem[] list = FTP_Sess.GetListing("", FtpListOption.NoPath);
string fullName = zosPath.Trim('\'').Trim('.') + '.' + list[0].FullName.Remove(0, 1);
```
**USS (linux-like) realm:**
```
string zosPath = "/projects";

FtpListItem[] list = FTP_Sess.GetListing(zosPath);
string fullName = list[0].FullName;
```

### DIFF
--- a/FluentFTP/Client/FtpClient_Listing.cs
+++ b/FluentFTP/Client/FtpClient_Listing.cs
@@ -250,8 +250,15 @@ namespace FluentFTP {
 			var isGetModified = options.HasFlag(FtpListOption.Modify);
 			var isGetSize = options.HasFlag(FtpListOption.Size);
 
-			// calc path to request
-			path = GetAbsolutePath(path);
+			// Only disable the GetAbsolutePath(path) if z/OS
+			// Note: "TEST.TST" is a "path" that does not start with a slash
+			// This could be a unix file on z/OS OR a classic CWD relative dataset
+			// Both of these work with the z/OS FTP server LIST command
+			if (path.StartsWith("/") || ServerType != FtpServer.IBMzOSFTP || ServerOS != FtpOperatingSystem.IBMzOS)
+			{
+				// calc the absolute filepath
+				path = GetAbsolutePath(path);
+			}
 
 			// MLSD provides a machine readable format with 100% accurate information
 			// so always prefer MLSD over LIST unless the caller of this method overrides it with the ForceList option
@@ -378,7 +385,9 @@ namespace FluentFTP {
 					//this.LogStatus(FtpTraceLevel.Verbose, "Skipped self or parent item: " + item.Name);
 				}
 			}
-			else {
+			// for z/OS, return of null actually means, just skip with no warning
+			if (ServerType != FtpServer.IBMzOSFTP || ServerOS != FtpOperatingSystem.IBMzOS)
+			{
 				LogStatus(FtpTraceLevel.Warn, "Failed to parse file listing: " + buf);
 			}
 			return true;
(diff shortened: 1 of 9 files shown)


robinrodricks:
Holy crap! Forgive my late reply but this already looks very cool! I'll
check it out soon and merge it. Thanks a ton!! 🥳

robinrodricks:
This is absolutely brilliant work! Thank you so much and sorry for the long delay in accepting this PR as I was caught up with other RL work. Thanks again! Much appreciated!

robinrodricks:
Your documentation is great - I created [this wiki page](https://github.com/robinrodricks/FluentFTP/wiki/IBM-zOS-and-OS-400-Support) based on it, feel free to send more snippets here and I can add it. If you want edits you can let me know.

robinrodricks:
When you get some time, could you implement the methods you described above within the core FluentFTP engine? It would be better than asking developers to implement their own version..

- GetPWD
- IsRoot
- GetRealm
- GetFileSize
- GetLRECL

FanDjango:
Hello Robin,

 

Here is the current status:

 

GetPWD not needed. GetWorkingDirectory works fine for z/OS (as does SetWorkingDirectoy).

Need to change the Wiki text.

 

PR #774 IsRoot API function
This is actually quite useful, see the PR comments
Needs documentation in the Wiki



PR #775 GetzOSListRealm() API function
Can be useful when walking the z/OS filesystem
Needs documentation in the Wiki



PR #777 Catch CWD in Execute to invalidate cwd cache.
While testing the above, I found a glitch in the code. z/OS users would use Execute("CWD….")

and this cause the cached working directory to out of sync with reality.

 

PR #778 Enhance the z/OS listing parser to get LRECL (via XDSS) on behalf of user

The listing parser now does the XDSS (if it is a non-RECFM=U partitioned dataset) and calculates 

the size of each member. The Wiki text needs to reflect that.

 

GetLRECL (see below): Not needed any more, see PR #778

 

GetFileSize (see below): Not needed any more, see PR #778

 

I am using these enhancements / modifications already in my z/OS FTP Client Gui, I was able to remove quite a lot of

kludgy snippets due to them.

 

I will try to take your big wiki text and suggest the modifications in the next days.

 

mit freundlichen Grüßen,

Michael Stiemke ***@***.***)

 

AnSyNova GmbH

Ringstr. 13

85405 Nandlstadt

+49 8756 969 49 79

PR #767, opened 2021-10-27
Small tweak in REAMDE.md for IBM servers System/z and System/i

FanDjango:
System/i runs (among others) OS/400.
System/z runs (among others) z/OS


robinrodricks:
Thanks!

Issue #768, opened 2021-10-27
FtpClient.GetListing() working directory regression bug


tommysor:
**FTP OS:** N/A
**FTP Server:** N/A
**Computer OS:** N/A

**FluentFTP Version:** 35.0.5
First Version with bug: 35.0.0

Commit: "sanitize all input paths" introduces regression bug.
Affects methods where null is a valid value for path parameter.

Known failing methods:
FtpClient.GetListing()
FtpClient.GetListingAsync()
FtpClient.GetNameListing()
FtpClient.GetNameListingAsync()

**Behavior for FtpClient.GetListing()**

- Expected (and actual up to version 34.0.2)
  - GetListing(null, Auto)
  - GetWorkingDirectory()
  - OpenPassiveDataStream(AutoPassive, "MLSD /working/directory", 0)
- Actual (version 35.0.5)
  - GetListing("/", Auto)
  - OpenPassiveDataStream(AutoPassive, "MLSD /", 0)

PR #769, opened 2021-10-27
#768 Listing methods where parameter "path" may be null will now use Working Directory instead of "/"

tommysor:
Fixes regression from "sanitize all input paths" / restores behavior from version 34.0.2.


robinrodricks:
Thanks for this!

PR #775, opened 2021-11-03
GetzOSListRealm() API function

FanDjango:
This is a new API function.

When writing a nice FTP Client GUI for file system navigation in z/OS, one comes across the problem of "What is the z/OS realm of the filesystem path described by the current working directory".

For a unix like filesystem, the realm is of course unix and such a function is not needed.

But for a z/OS, a specialised function can make things easier.

The currrent working directory on z/OS either begins with a `/`, which tells us: realm is HFS.

Otherwise, it must be one of the z/OS (aka MVS) realms. But which one?

Short of doing a LIST command (which would be one way to do this), this makes use of a nice z/OS FTP server feature: When you `CWD` to an arbitrary path, if that path denotes a partitioned dataset, your are told that this is the case. And even more interesting: If it then also happens to be a RECFM=U PDS, your are **also** told that it is (probably) a load library.

This means you can determine the realm of a z/OS path **WITHOUT** performing a LIST.

An alternative in terms of coding this would be to perform a LIST on the path and to analyze the results, perhaps by calling the dirlist parser.  But such a LIST might be huge, so that approach should be avoided.

### DIFF
--- a/FluentFTP/Client/FtpClient_FolderManagement.cs
+++ b/FluentFTP/Client/FtpClient_FolderManagement.cs
@@ -774,5 +774,119 @@ namespace FluentFTP {
 #endif
 
 		#endregion
+
+		#region GetzOSListRealm
+
+		/// <summary>
+		/// If an FTP Server has "different realms", in which realm is the
+		/// current working directory. 
+		/// </summary>
+		/// <returns>The realm</returns>
+		public FtpzOSListRealm GetzOSListRealm()
+		{
+
+			// this case occurs immediately after connection and after the working dir has changed
+			if (_LastWorkingDir == null)
+			{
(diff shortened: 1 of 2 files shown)


FanDjango:
I will add some docs in the next days.

FanDjango:
I have some ideas to make this more "server relevant". Please stand by and don't merge.

robinrodricks:
Outstanding work! Added to wiki under API section.

robinrodricks:
Released!! https://www.nuget.org/packages/FluentFTP/35.1.0

Great work my friend and thank you a metric ton for this excellent work on supporting IBM z/OS for FluentFTP. You have my gratitude and my respect. Have a great day1

PR #776, opened 2021-11-03
Fix unwanted Modified date overwrite.

Dylan-DutchAndBold:
This change prevents the Modified date of a FtpListItem being overwritten if the consecutive MDTM command fails.


robinrodricks:
Excellent work, thank you!

PR #779, opened 2021-11-04
GetzOSFileSize(string path) API function

FanDjango:
This seemed like a good idea after all. Saves a lot of coding in the user code.


robinrodricks:
Very rare to get new APIs from contributors. Thanks a ton!!

PR #780, opened 2021-11-05
Re-use of client instance to server with different capabilities

FanDjango:
Re-using a client instance to connect to a server with different capabilites is possible but the `FEAT` command is not performed on the new connection. This causes problems when the new server has **LESS** features than the previous server. Unless it is a cloned connection.

Simple solution is to clear the capabilites list if it is not a clone.

### DIFF
--- a/FluentFTP/Client/FtpClient_Connection.cs
+++ b/FluentFTP/Client/FtpClient_Connection.cs
@@ -437,6 +437,11 @@ namespace FluentFTP {
 
 				// if this is a clone these values should have already been loaded
 				// so save some bandwidth and CPU time and skip executing this again.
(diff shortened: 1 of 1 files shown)


FanDjango:
**NOTE**
How did I notice this problem? Two cases: 
1. Connect to Pure-FTPd and get a huge FEAT list. Then disconnect. Then connect to IBM z/OS FTP and this one has no FEAT's at all. But the previous FEAT list is still active causing wrong list parser selection in this case, because of FEAT MLST being wrongly assumed.
2. Some servers are configured differently and report different lists of FEATs.

robinrodricks:
Great work! Generally I don't recommend reusing the same client to connect to different servers, but it appears to be a valid use-case. So accepted.

PR #781, opened 2021-11-05
Fix GetListing() regression caused by PR #764 - many untrue parse fail warnings

FanDjango:
**urgent**

GetListing() works but PR 764 causes very many unneccessary parse fail warnings.

An `else` keyword somehow disappeared on the commit/push sequence.


robinrodricks:
Sorry for the long delay, just got time and your entire listing of PRs looks like some truly amazing work. Keep it up!

PR #782, opened 2021-11-05
FtpListOption.ForceList is not being honored

FanDjango:
When connected to a FEAT MLST / MLSD server, the parser is set to "machine".

If you now specify FtpListOptions.ForceList, one would expect this to **take precedence** over the "machine" setting of the parser.

You absolutely need this if you are interested in seeing entries that are _links_ - these are not supported by the MLSD "machine" parser, for understandable reasons (see [this](https://www.rfc-editor.org/errata/eid1500) to realise that MLST might be very nice and fast, easier to parse, but sadly it currently does not support link entries in a standardized way.


robinrodricks:
Excellent find, thnk

PR #784, opened 2021-11-15
Better handling of not-accepted certificates on autoconnect

FanDjango:
This one is hard to explain:

If:

You use autoconnect

and

You decide to use the `OnValidateCertificate` delegate

and

You decide to **REJECT** the certificate...

the current logic will notice the failure and attempt a disconnect, which will attempt a "QUIT" execution, which will percolate an outstanding exception, which makes everything following this very unclean.

I feel that a successful connection that fails due to a rejected certificate should terminate the sequence of connections attempts, and furthermore, do this in a more graceful fashion.

My 2 cents on this, so I submit this for review and discussion.

As with all other PRs I am using these ASIS for a z/OS - Unix FTP navigator.

### DIFF
--- a/FluentFTP/Client/FtpClient_AutoConnection.cs
+++ b/FluentFTP/Client/FtpClient_AutoConnection.cs
@@ -7,6 +7,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Net;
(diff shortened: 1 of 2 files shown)


robinrodricks:
Great work, thank you!

Issue #82, opened 2017-05-04
Do you support SOCKS5 proxies?


pwallner:
is working with socks5?

I used this code, but it doesn't work
```
Dim fluproxy As FluentFTP.Proxy.ProxyInfo = New FluentFTP.Proxy.ProxyInfo
            fluproxy.Host = "127.0.0.1"
            fluproxy.Port = 3000
            fluproxy.Credentials = New NetworkCredential("", "")

            Dim client22 = New FluentFTP.Proxy.FtpClientUserAtHostProxy(fluproxy) ' FtpClient()
            client22.Host = "192.168.0.246"
            client22.Port = 21
            'client22.SslProtocols = System.Security.Authentication.SslProtocols.Tls12
            'if you don't specify login credentials, we use the "anonymous" user account
            client22.Credentials = New NetworkCredential("xxx", "yyyy")


            'begin connecting to the server
            client22.Connect()
```

without proxy everything is working...

hgupta9:
I don't think SOCKS proxy support has been implemented. Only HTTP 1.1 and User@Host AFAIK. Maybe the authors of the proxy can confirm otherwise. @Zoltan666 @elmar69 @L3Z4

I suggest using [EDT FTP](http://enterprisedt.com/products/edtftpnet/) as it has support for SOCKS proxy, but NOT in free version. As per [docs](http://enterprisedt.com/products/edtftpnet/doc/manual/index.html):

> Proxy support is not included in edtFTPnet/Free.

pwallner:
I found [this](https://github.com/poma/ProxySocket), maybe it helps?? I can connect via Socks5. But with this, I'm far away from any ftp solution...

pwallner:
And [here](https://github.com/bentonstark/starksoft-aspen) I found this which works, but I wish if you can implement this...

[Here](http://stackoverflow.com/questions/73544/ftp-client-class-for-net) I found the code for starksoft ftp:
```
Socks4ProxyClient socks = new Socks4ProxyClient("socksproxyhost",1010);
FtpClient ftp = new FtpClient("ftpshost",2010,FtpSecurityProtocol.Tls1Explicit);
ftp.Proxy = socks;
ftp.Open("userid", "******");
ftp.PutFile(@"C:\519ec30a-ae15-4bd5-8bcd-94ef3ca49165.xml");
Console.WriteLine(ftp.GetDirListAsText());
ftp.Close();
```

gyma11:
[...]

PR #785, opened 2021-11-17
#82 - Implement No Auth SOCkS proxy 

bjth:
#82 Adding unauthenticated SOCKS5 proxy support.

Changes include adding an implementation of a SOCKS proxy following the RFC : https://datatracker.ietf.org/doc/html/rfc1928

Tested against VSFTPD using Dante as a SOCKS proxy, both locally and against an Azure installation, where the FTP was running on a local address.

Commands to spin up a local copy for testing:

`docker run -d --restart=always -p 9020:20 -p 9021:21 -p 9100-9110:9100-9110 -e FTP_USER=ftp -e FTP_PASS=ftp -e PASV_ADDRESS=172.18.144.1 -e PASV_MIN_PORT=9100 -e PASV_MAX_PORT=9110 -v /data/ftp:/home/vsftpd fauria/vsftpd`

`docker run -d -p 1080:1080 wernight/dante`

Example usage can be found in the test project.

### DIFF
--- a/FluentFTP.Tests/Tests.cs
+++ b/FluentFTP.Tests/Tests.cs
@@ -28,6 +28,9 @@ namespace Tests {
 		private const string m_user = "";
 		private const string m_pass = "";
 
+		private const string m_proxy_host = "";
+		private const int m_proxy_port = 0;
+
 		private static readonly int[] connectionTypes = new[] {
 			(int) FtpDataConnectionType.EPSV,
 			(int) FtpDataConnectionType.EPRT,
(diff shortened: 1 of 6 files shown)


robinrodricks:
Absolutely fantastic work! This has been a long requested feature and I'm so grateful that you implemented it. Kudos and have a great day!

robinrodricks:
Added some FAQs and proxy usage - https://github.com/robinrodricks/FluentFTP/wiki/FTPS-Proxies

robinrodricks:
Will release soon.

robinrodricks:
Released.

https://www.nuget.org/packages/FluentFTP/35.2.0

bjth:
Much appreciated, and thank you for the kind words.

Have a great weekend.

ilc-ilopez:
FYI for others trying to use FluentFTP with a authenticated SOCKS5 proxy. It will not work (as of today).

The negotiation packet is sending a "No Authentication" method to the SOCKS5 proxy. Which in our use-case it is authenticated.

https://github.com/robinrodricks/FluentFTP/blob/4281b6f049193ac79d7a773498069beeabd6b3a4/FluentFTP/Proxy/Socks/SocksProxy.cs#L39

In reviewing how this happened, launching the danted package from docker is preconfigured for no-authentication.
https://hub.docker.com/r/wernight/dante

`The default config in this image allows everyone to use the proxy. You can add a simple authentication (which will send data unencrypted) by setting up a Dockerfile like:`

Which explains how we have SOCKS5 proxy support, but only if its behind a firewall because I doubt anyone would have a wide-open proxy on the public internet.

I will do what I can to improve this, and submit a pull request, if time allows.

PR #788, opened 2021-11-17
CalculateFullFtpPath final Code + Wiki update

FanDjango:
### **Consolidates the IBM z/OS fixes for `.FullName` with the preparatory (don't merge) PR's #786, #787 + Wiki change (see comment below)**

Together with the (small) code changes for #786 and #787, there is a quite large block of code needed for IBM z/OS, to calculate `.FullName` in `GetListing(....)` entries for the different scenarios of using `GetListing(...)` with or without `FtpListOption.NoPath` in the different `zOSListingRealm`s.

So this PR contains the needed code for the **whole** fix. The previous PR's (#786, #787) are solely to explain the "collateral" damage when trying to  fixup `CalculateFullFtpPath()`.

It contains the smaller amounts of code for #786, #787 **plus** this large block of IBM z/OS code, which 
 looks like this:
```
			if (client.ServerType == FtpServer.IBMzOSFTP &&
				client.ServerOS == FtpOperatingSystem.IBMzOS &&
				client.zOSListingRealm != FtpZOSListRealm.Unix)
			{
				// The user might be using GetListing("", FtpListOption.NoPath)
				// or he might be using    GetListing("not_fully_qualified_zOS path")
				// or he might be using    GetListing("'fully_qualified_zOS path'") (note the single quotes)

				// The following examples in the comments assume a current working
				// directory of 'GEEK.'.

				// If it is not a FtpZOSListRealm.Dataset, it must be FtpZOSListRealm.Member*

				// Is caller using FtpListOption.NoPath and CWD to the right place?
				if (path.Length == 0)
				{
					if (client.zOSListingRealm == FtpZOSListRealm.Dataset)
					{
						// Path: ""
						// Fullname: 'GEEK.PROJECTS.LOADLIB'
						item.FullName = client.GetWorkingDirectory().TrimEnd('\'') + item.Name + "\'";
					}
					else
					{
						// Path: ""
						// Fullname: 'GEEK.PROJECTS.LOADLIB(MYPROG)'
						item.FullName = client.GetWorkingDirectory().TrimEnd('\'') + "(" + item.Name + ")\'";
					}
				}
				// Caller is not using FtpListOption.NoPath, so the fullname can be built
				// depending on the listing realm
				else if (path[0] == '\'')
				{
					if (client.zOSListingRealm == FtpZOSListRealm.Dataset)
					{
						// Path: "'GEEK.PROJECTS.LOADLIB'"
						// Fullname: 'GEEK.PROJECTS.LOADLIB'
[...]

PR #790, opened 2021-11-18
SITE DATASETMODE and QUOTESOVERRIDE commands on connect to IBM z/OS FTP server

FanDjango:
See the new wiki text in PR #788


robinrodricks:
This is brilliant! I was about to suggest it but its like you read my mind!

Issue #751, opened 2021-08-26
GetObjectInfo falling back to GetListing fails when not setting path to FTP path


JosHuybrighs:
**FTP OS:** Don't know

**FTP Server:** Don't know

**Computer OS:** Windows 10

**FluentFTP Version:** 34.0.2

When using GetObjectInfo(path) towards a server that doesn't support MLST then the check to find a match with the returned items of GetListing assumes that 'path' is set in FTP format, i.e. "/root/content/sync".

Code in GetObjectInfo, which does a wrong check for 'path':
```
var dirPath = path.GetFtpDirectoryName();
var dirItems = GetListing(dirPath);
foreach (var dirItem in dirItems) {
	if (dirItem.FullName == path) {
		result = dirItem;
		break;
	}
}
```

The overcome this error one has to call GetObjectInfo as follows:
```
var ftpPath = path.GetFtpPath();
var item = _ftpClient.GetObjectInfo(ftpPath);
```

Is this really an error or do we have to use the FTP path format everywhere in our code? I though that FluentFTP takes care of mixing forward and backward slashes.

robinrodricks:
I have fixed it for GetObjectInfo. I will ensure all other methods are also fixed and cleanup FTP paths internally.

Please try this build:

[FluentFTP.35.0.0_BETA5.zip](https://github.com/robinrodricks/FluentFTP/files/7070451/FluentFTP.35.0.0_BETA5.zip)

robinrodricks:
Released.

https://www.nuget.org/packages/FluentFTP/35.0.0

robinrodricks:
Hi @JosHuybrighs, can you try the latest build from nuget and tell me if it resolves your issue? I need feedback on this please.

AngryShaDoW:
At v35.2.0 client.GetListing() still throw null reference exception. Probably because this:
```
public FtpListItem[] GetListing()
{
    return GetListing(null);
}

public FtpListItem[] GetListing(string path, FtpListOption options)
{
...
    if (path.StartsWith("/") || ServerType != FtpServer.IBMzOSFTP || ServerOS != FtpOperatingSystem.IBMzOS)
    {
        path = GetAbsolutePath(path);
    }
...
}
```

FanDjango:
I'll check that

FanDjango:
Ok, I can reproduce that. Thx for the heads up, my recent z/OS changes are at fault. I'll put in a PR to fix that right away.

[...]

PR #793, opened 2021-11-22
Fix Regression #751 due to IBM z/OS changes

FanDjango:
Recently updated issue #751 describes the case where `path` can be `null`, thus causing an exception to be thrown.

So, rearranging the OR clause and adding a `null` check to fix this.


robinrodricks:
Great fix, thanks!

robinrodricks:
Released.

https://www.nuget.org/packages/FluentFTP/35.2.1

PR #794, opened 2021-11-24
z/OS Parser: Big files, do not want negative file sizes

FanDjango:
Discovered this one listing files under "SYS1.".

It is very minor, has no huge consequences. But for the sake of getting it right... very big files will overflow on the size calculation resulting in negative file sizes. We are talking about `ntracks * 56664` overflowing `int`. But, can happen, as I saw.


robinrodricks:
Awesome, thanks!

robinrodricks:
Released.

https://www.nuget.org/packages/FluentFTP/35.2.1

PR #796, opened 2021-11-25
Getting some infinite transferSpeeds on progress delegate

FanDjango:
I was getting some sporadic errors (value is infinity, so no assign to `long` possible) in my download progress delegate.

Thus, the first thing was to use (in my own progress delegate code):

`double.IsInfinity(p.TransferSpeed) ? long.MaxValue : (long)p.TransferSpeed)`

Then I had a look at `FtpProgress.cs` and saw:

```
			if (double.IsNaN(transferSpeed) && double.IsInfinity(transferSpeed)) {
				transferSpeed = 0;
			}
```

Trouble is, I was getting `PositiveInfinity`, once in a while.

Should it not be:

```
			if (double.IsNaN(transferSpeed) || double.IsInfinity(transferSpeed)) {
				transferSpeed = 0;
			}
```

But I decided to make the checks more specific to the possible division by zero errors possible in the try section.

I also believe that `progressValue` is safe because of `if (fileSize > 0)`.

I also corrected the spelling mistake: `estimatedRemaingTime`

Please look the suggest code and review.

### DIFF
--- a/FluentFTP/Model/FtpProgress.cs
+++ b/FluentFTP/Model/FtpProgress.cs
@@ -102,40 +102,36 @@ namespace FluentFTP {
 
 			// default values to send
 			double progressValue = -1;
-			double transferSpeed = 0;
-			var estimatedRemaingTime = TimeSpan.Zero;
+			double transferSpeed;
+			var estimatedRemainingTime = TimeSpan.Zero;
 
 			// catch any divide-by-zero errors
 			try {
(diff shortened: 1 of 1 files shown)


FanDjango:
I am closing this one. Casting double to long was as additional problem.
Still believe we need an "or" in there and not an "and", but I will do further research if I ever have time.

PR #797, opened 2021-11-29
z/OS: Consolidation and some further (small) changes

FanDjango:
Thanks to the previous merges, here are (more or less) the last changes:

----
`FtpClient_FileProperties.cs`
If a z/OS server reports "SIZE" capability, things go wrong in case of z/OS realm datasets. Because other than in the HFS realm, no size is available. On the other hand, if it does not report the "SIZE" capability, we can **still** obtain a size due to the previous changes we made. So let's do it ourselves... ==> ignore SIZE capability, even if the z/OS FTP server is thus configured.

...but this then means:

`FtpClient_FileDownload.cs`
Since we have a size now (however we acquired it) on z/OS files (whichever realm), but this size is **not** accurate (see wiki), we need to `readToEnd` = `true` ==> this means **act** like `fileLen` is zero.

----
`FtpClient_IBMzOS.cs`
Since the last change, the z/OS parser himself knows how to handle setting the path for a XDSS command on a partitioned dataset (by removing the parens and their content). So we can really simplify this API function. The `GetZOSFileSize()` API function could even be removed and its code moved into `GetFileSize()` and the wiki updated. Please indicate if you would prefer that. Right now, the `GetFileSize()` API function will internally call `GetZOSFileSize()` on zOS servers, making the code look cleaner. The `ListingParser = FtpParser.IBMzOS` prevents automatic parser detection switching to unix on z/OS HFS files. 
```
			FtpListItem[] entries = GetListing(path);
			if (entries.Length != 1) return -1;
			FtpListItem entry = entries[0];
```
This snippet "misuses" `GetListing(`path)` to provide a single entry and to get the size from that. 

----
`FtpProgress.cs`
`Nan` and `Infinity`: A (double) value can be **either** `NaN` **or** `PositiveInfinity/NegativeInfinity` but not both at the same time. I also would have at first thought that an infinity representation should also be flagged as "not a number", but duh... [see this](https://docs.microsoft.com/en-us/dotnet/api/system.double.nan?redirectedfrom=MSDN&view=net-6.0)
I think the logic of this progress function can be simplified/improved. I will see if I find time...

### DIFF
--- a/FluentFTP/Client/FtpClient_FileDownload.cs
+++ b/FluentFTP/Client/FtpClient_FileDownload.cs
@@ -678,7 +678,7 @@ namespace FluentFTP {
 
 				// if the server has not provided a length for this file
 				// we read until EOF instead of reading a specific number of bytes
-				var readToEnd = fileLen <= 0;
+				var readToEnd = fileLen <= 0 || (ServerType == FtpServer.IBMzOSFTP && ServerOS == FtpOperatingSystem.IBMzOS);
 
 				const int rateControlResolution = 100;
 				var rateLimitBytes = DownloadRateLimit != 0 ? (long)DownloadRateLimit * 1024 : 0;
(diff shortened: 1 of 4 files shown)


robinrodricks:
Thanks for the great work as usual.

robinrodricks:
Published.
https://www.nuget.org/packages/FluentFTP/35.2.2

PR #805, opened 2021-12-17
BUG FIX: Set Data Type DIRECTLY before the RETR, z/OS ASCII transfer fails

FanDjango:
`length = checkIfFileExists ? client.GetFileSize(path) : 0;` can, in some cases, cause a `SetDataType(...)` for BINARY. This can happen for z/OS, where the `FileExists` and the `GetFileSize` functionality relies on getting a file listing, which in turn sets BINARY.

I propose:

If the `SetDataType` asks for `ASCII`, this will be overturned be a `BINARY` from the `GetListing(...)`, **unless the `SetDataType` is placed DIRECTLY before the RETR command. So this is the safest place for the `SetDataType`**.

Question: So much size-getting. Was gotten before, why again (see #806)? Needs to be investigated...

### DIFF
--- a/FluentFTP/Client/FtpClient_Stream.cs
+++ b/FluentFTP/Client/FtpClient_Stream.cs
@@ -1231,9 +1231,8 @@ namespace FluentFTP {
 					client = this;
 				}
 
-				client.SetDataType(type);
-
 				length = checkIfFileExists ? client.GetFileSize(path) : 0;
+				client.SetDataType(type);
 				stream = client.OpenDataStream("RETR " + path, restart);
 #if !CORE14
 			}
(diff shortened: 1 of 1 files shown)


robinrodricks:
We keep getting the file size because its the fastest way to check if a file exists. The older implement was to start a file download but that is even worse.

robinrodricks:
I didn't understand the need for this PR, could you elaborate?

FanDjango:
If you want a z/OS ASCII file transer - the transfer mode is set to ASCII, but due to the way the files size is retrieved (GetListing), transfer mode is set to IMAGE, which sets the transfer back to image.

FanDjango:
Re "Size getting".

> We keep getting the file size because its the fastest way to check if a file exists. The older implement was to start a file download but that is even worse.

Sadly, "Size Getting" to check for existence is VERY expensive on z/OS, because that needs a GetListing AND in some cases a XDSS command.

One should delegate the existance check to "FileExists" which will choose the best way to do this instead of doing a "Getsize".

In many cases "FileExist" will choose to do a SIZE, or a MDTM or whatever, but in the case of z/OS it WILL NOT.

Actually that makes me feel I should amend this PR to do just that: Use FileExist.

But still, any actions that might change the transfer type back from ASCII to IMAGE should be done BEFORE the transfer type command. The ordering is important-

PR #806, opened 2021-12-17
Clean up FileDownload read to EOF / progress / zOS

FanDjango:
Unclean progress when z/OS FileDownload with ASCII,  because size is not retrieved (even if it is unneeded due to read to EOF)

When progress is requested, you should ALWAYs get the file size.

Then, when reading to EOF is to be decided, one can use:

Read to EOF when:

Server has not provided a length
or
Transfer is ASCII
or
Transfer is to z/OS

This issue of "Size getting" needs to be cleaned up. See #805

### DIFF
--- a/FluentFTP/Client/FtpClient_FileDownload.cs
+++ b/FluentFTP/Client/FtpClient_FileDownload.cs
@@ -667,18 +667,20 @@ namespace FluentFTP {
 			var disposeOutStream = false;
 
 			try {
-				// get file size if downloading in binary mode (in ASCII mode we read until EOF)
+				// if progress requested get file size (if not known already)
 				long fileLen = 0;
-				if (DownloadDataType == FtpDataType.Binary && progress != null) {
+				if (progress != null) {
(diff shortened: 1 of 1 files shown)


robinrodricks:
Again, I did not understand the need for this. Apart from z-OS, in all other cases we cannot / should not get the file size for progress reports on ASCII types because ASCII reading means we are reading till EOF. So wasting time fetching the size does nothing, as many times the server fails to get the file size. File size only accurately reads in binary mode (iirc). If this is a z-OS specific change, please add appropriate if/else logic.

FanDjango:
It is really just a refactor plus a z/OS specific change.

Apart from this:

"When progress is requested, you should ALWAYs (try to) get the file size".

robinrodricks:
Ok, lets add this in, if there is an issue, we will have to fix it later.

Sadly as this is an integration library and I don't have dozens of ready to use FTP servers for testing, integration testing is impossible. It would be nice to have, perhaps using docker. So that way we can run automated tests at least for critical behaviours.

robinrodricks:
Oh wait its closed...

PR #810, opened 2021-12-28
z/OS FTP Server SYST command replies differently depending on realm

FanDjango:
This PR concerns `DetectFtpOSBySyst(...)`:

Someone reported that after connecting to his z/OS CS FTP server, he cannot access the z/OS realm.

It turns out FluentFTP recognizes the system as being "UNIX".

Research showed up the following:

There is a parameter for the IBM z/OS CS FTP server called "STARTUPDIRECTORY" (see [here](https://www.ibm.com/docs/en/zos/2.4.0?topic=protocol-startdirectory-ftp-server-statement#strtdir)).

It was set to "HFS", which causes the `SYST` command to respond like this:

`215 UNIX is the operating system of this server. FTP Server is running on z/OS.`

Up to now we have only been expecting:

`215 MVS is the operating system of this server. FTP Server is running on z/OS.`

Both of these texts contain the needed "`Z/OS`" (ToUppered...), but the `if` clauses trigger on the "`UNIX`" some lines further up. Therefore, a re-ordering of these statements is needed to solve that problem.

### DIFF
--- a/FluentFTP/Servers/FtpServerSpecificHandler.cs
+++ b/FluentFTP/Servers/FtpServerSpecificHandler.cs
@@ -134,6 +134,13 @@ namespace FluentFTP.Servers {
 				// Windows OS
 				serverOS = FtpOperatingSystem.Windows;
 			}
+			else if (system.Contains("Z/OS"))
+			{
+				// IBM z/OS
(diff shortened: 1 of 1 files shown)


robinrodricks:
Perfect thanks!

robinrodricks:
Published.

PR #811, opened 2021-12-28
Stop some callers from bypassing the CurrentDataType cache

FanDjango:
This PR concerns `SetDataType(...)`:

In the log, I was surprised to see so many unneeded "`TYPE I`" commands, even though there is a caching mechanism (similar to the one for "`CWD`" / "`PWD`").

It turns out that some callers (rightly) go in at `...NoLock(...)`, thereby bypassing the cache.

This PR moves the caching code down into the no-lock part of the code. The diff looks crazy, I know, maybe lay the two versions side by side :-)

### DIFF
--- a/FluentFTP/Client/FtpClient_Stream.cs
+++ b/FluentFTP/Client/FtpClient_Stream.cs
@@ -1643,13 +1643,7 @@ namespace FluentFTP {
 			lock (m_lock) {
 #endif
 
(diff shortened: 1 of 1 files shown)


robinrodricks:
Thanks for finding this and fixing it!

robinrodricks:
Published.

PR #814, opened 2021-12-28
SetDataType DIRECTLY before the RETR

FanDjango:
This PR is a reiteration of #805, concerns `OpenRead(...)`

I closed that to research some other problems and to isolate a number of other things to look at, but am putting this back in now.

SInce the changes for z/OS, the `GetFileSize(...)` which is performed prior to the `RETR` will (in case of z/OS) perform a `GetListing(...)` to determine the file size. But this will issue a `SetDataType(...)` for `TYPE I`. This will counter the `SetDataType(...)` for `ASCII` that a user might have requested.

This is not an old problem. It was introduced by the z/OS changes (and only happens for z/OS servers).

It means that a z/OS server cannot be told to perform an `ASCII` transfer unless this is merged.

### DIFF
--- a/FluentFTP/Client/FtpClient_Stream.cs
+++ b/FluentFTP/Client/FtpClient_Stream.cs
@@ -1231,9 +1231,9 @@ namespace FluentFTP {
 					client = this;
 				}
 
-				client.SetDataType(type);
-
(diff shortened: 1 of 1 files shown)


robinrodricks:
Thanks!

robinrodricks:
Published.

PR #817, opened 2022-01-02
File_Download / OpenRead Refactor

FanDjango:
In order to get progress information on ASCII transfers and to reduce the number of redundant calls to `GetFileSize(...)` I needed to do the following:

Oh, and I think you are not going to like this, maybe. Big bad ugly PR's are more a thing for the maintainer than for contributer :-)) BTW: This is not IBM z/OS specific.

In `DownloadFileInternal(...)`:
Separate the decision to get the file size from the decision of when to read to the end.

Thus: 
- get the file length (unless already known = supplied by caller) if progress is required, regardless of ASCII or Binary. If file length is "already known", just use that. If no progress is requested, and there is no file length, `OpenRead(...)` will do it internally.
- pass the file length (which might be zero) to `OpenRead(....)` and modify that function to understand that information.
- set the read to end decision, when there is no file length, or when ASCII transfer or when IBM z/OS.

In `OpenRead(...)`:
Refactor and remove uneeded function overloads.
Reflect all relevant removals and changes to `IFtpClient.cs`.
Add new _master_ function prototype which takes the file length (and remove `checkIfFileExists`)

Since all callers of `OpenRead(...)` other than `DownloadFileInternal(...)` use `true` for the now redundant parameter `checkIfFileExists`, it can be removed. The needed information is already self-contained in the new parameter `fileLen`. This combines the previous information `fileLen > 0` - the entire `fileLen` is now passed to `OpenRead(...)`. 

The previous logic was strange anyway: Why should a file length greater that zero cause **ANOTHER** `GetFileSize(...)`. The known file size was already gotten once, so why again? It is only done in the code to set the stream length.

### DIFF
--- a/FluentFTP/Client/FtpClient_FileDownload.cs
+++ b/FluentFTP/Client/FtpClient_FileDownload.cs
@@ -667,18 +667,23 @@ namespace FluentFTP {
 			var disposeOutStream = false;
 
 			try {
-				// get file size if downloading in binary mode (in ASCII mode we read until EOF)
+				// get file size if progress requested
 				long fileLen = 0;
-				if (DownloadDataType == FtpDataType.Binary && progress != null) {
+
+				if (progress != null) {
 					fileLen = knownFileSize > 0 ? knownFileSize : GetFileSize(remotePath);
 				}
 
 				// open the file for reading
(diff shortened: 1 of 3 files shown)


FanDjango:
Some more refactoring needed, closing. Re-open some time in the future...

PR #819, opened 2022-01-02
Download/Upload/Stream refactor

FanDjango:
I wondered how to PR this, it is impossible to separate into smaller pieces, so please have a good look at the following and let's discuss. It is the last of my changes, nothing more in the pipeline, but I have left the worst one for the last. Big bad ugly PR's are more a thing for the maintainer than for a contributer :-)) BTW: This is not IBM z/OS specific, but I am currently actually using it to improve the users experience with my FTP client.

This replaces PR #817 which I closed in order to create a "tidier" and more complete PR.

### In order to get progress information **even on ASCII download transfers** and **to reduce the number of redundant calls to `GetFileSize(...)`** I needed to do the following:

**In `DownloadFileInternal(...)`:**

Separate the decision to _get the file size_ from the decision of when to _read to the end_.

Thus:

get the file length (unless already known = supplied by caller) if progress is required, regardless of ASCII or Binary. If file length is "already known", just use that. If no progress is requested, and there is no file length, `OpenRead(...)` will do it internally.
Pass the file length (which might be zero, this replaces the original bool true) to `OpenRead(....)` and modify that function to understand that information.

set the _read to end_ decision, when there is no file length, or when ASCII transfer or when IBM z/OS.

**In `OpenRead(...)`:**

Refactor and remove uneeded function overloads.

Reflect all relevant removals and changes to `IFtpClient.cs`.

Add new master function prototype which takes the file length (and remove `bool checkIfFileExists`)

Since all callers of `OpenRead(...)` other than `DownloadFileInternal(...)` use true for the now redundant parameter `checkIfFileExists`, it can be removed. The needed information is already self-contained in the new parameter `fileLen`. This combines the previous information `fileLen > 0` - the entire file length is now passed to `OpenRead(...)`.

The previous logic was strange anyway: Why should a file length greater that zero cause ANOTHER `GetFileSize(...)`. The known file size was already gotten once, so why again? It is only done in the code to set the stream length.

Essentialy the same logic was applied to `OpenWrite(...)` and `OpenAppend(...)` but the needed changes were much less complex. I also expanded PR #814 to cover these two functions - although in theory this problem can really only happen in `OpenAppend(...)`.

### DIFF
--- a/FluentFTP/Client/FtpClient_FileDownload.cs
+++ b/FluentFTP/Client/FtpClient_FileDownload.cs
@@ -667,18 +667,23 @@ namespace FluentFTP {
 			var disposeOutStream = false;
 
 			try {
-				// get file size if downloading in binary mode (in ASCII mode we read until EOF)
+				// get file size if progress requested
 				long fileLen = 0;
-				if (DownloadDataType == FtpDataType.Binary && progress != null) {
+
+				if (progress != null) {
 					fileLen = knownFileSize > 0 ? knownFileSize : GetFileSize(remotePath);
 				}
 
 				// open the file for reading
-				downStream = OpenRead(remotePath, DownloadDataType, restartPosition, fileLen > 0);
+				downStream = OpenRead(remotePath, DownloadDataType, restartPosition, fileLen);
 
-				// if the server has not provided a length for this file
+				// if the server has not provided a length for this file or
+				// if the mode is ASCII or
+				// if the server is IBM z/OS
 				// we read until EOF instead of reading a specific number of bytes
-				var readToEnd = fileLen <= 0 || (ServerType == FtpServer.IBMzOSFTP);
+				var readToEnd = (fileLen <= 0) || 
+								(DownloadDataType == FtpDataType.ASCII) || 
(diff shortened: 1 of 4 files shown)


robinrodricks:
This looks much more mature than your previous changes for the file length issue. Thanks for the excellent work as usual!

robinrodricks:
Released.

https://www.nuget.org/packages/FluentFTP/36.0.0

FanDjango:
Thanks. I will monitor for any problems that arise.

FanDjango:
That's a good idea, I will do that and submit a PR to doc that.

I will also check the code to make sure this works in all cases.

FanDjango:
@MrJul see #826 and thanks for the heads up

FanDjango:
I actually took the decision (in the latest incarnation of #826), to bring back the old methods and to add the new one - so @MrJul you can decide which call type you prefer. Depends on what Robin decides, though, too.

MrJul:
Updated to v36.1.0 and used the new-old bool overload, everything works fine. Thanks for taking my feedback into account and the quick turnaround! :)

PR #822, opened 2022-01-05
Fix #821 FtpSocketStream mishandles timeout and CancellationToken

### LINKED ISSUE #821: FtpSocketStream mishandles timeout and CancellationToken
**FTP OS:** Irrelevant

**FTP Server:** Irrelevant

**Computer OS:** Windows

**FluentFTP Version:** 35.2.3

Currently `ConnectTimeout` is not working for `ConnectAsync` (at least for .NET Core). While debugging the source of the issue, I found out an (unrelated) issue is that a `CancellationToken` does not actually severe the `Socket` created as well.

```c#
const string Host = "10.255.255.1"; // To ensure it timeouts
const int Timeout = 1000;
const int TokenTimeout = 3000;

using var timeoutClient = new FtpClient(Host)
{
	ValidateAnyCertificate = true,
	EncryptionMode = FtpEncryptionMode.Explicit,

	ConnectTimeout = Timeout,
	DataConnectionConnectTimeout = Timeout,
	DataConnectionReadTimeout = Timeout,
	ReadTimeout = Timeout,
};

var start = DateTime.UtcNow;
Console.WriteLine("Started: " + start);
try
{
	var src = new CancellationTokenSource();
	src.CancelAfter(TokenTimeout);

	// Neither timeout nor token expiration works
	await timeoutClient.ConnectAsync(src.Token);

	// Connect() works correctly
	//timeoutClient.Connect();
}
catch (SocketException)
{
	Console.WriteLine((DateTime.UtcNow - start).TotalMilliseconds + "ms");
	return;
}
```

Without interference, it takes about 20 seconds for the Socket to timeout.

**Logs :**

No Log since no connection is made.

datvm:
Fix #821 FtpSocketStream mishandles timeout and CancellationToken.

Note that this pull request only fixes .NET Core code. The issue may still exists in other .NET version.

### DIFF
--- a/FluentFTP/Streams/FtpSocketStream.cs
+++ b/FluentFTP/Streams/FtpSocketStream.cs
@@ -977,6 +977,16 @@ namespace FluentFTP {
 				m_socket = new Socket(addresses[i].AddressFamily, SocketType.Stream, ProtocolType.Tcp);
 				BindSocketToLocalIp();
 #if CORE
+				if (this.ConnectTimeout > 0)
+				{
+					var timeoutSrc = new CancellationTokenSource();
+					timeoutSrc.CancelAfter(this.ConnectTimeout);
+
+					var tokenSrc = CancellationTokenSource.CreateLinkedTokenSource(
+						token, timeoutSrc.Token);
+					token = tokenSrc.Token;
+				}
+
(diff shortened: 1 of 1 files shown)


robinrodricks:
Thanks a lot!

robinrodricks:
Released.

https://www.nuget.org/packages/FluentFTP/36.0.0

PR #826, opened 2022-01-14
Bring back original OpenXXX methods as documented in Wiki and ADD the new ones instead

FanDjango:
**Concerns the OpenRead, OpenAppend, OpenWrite group of methods**

In response to [this](https://github.com/robinrodricks/FluentFTP/pull/819#issuecomment-1013114436).

The changes to the last version caused the following problem:

For these methods, the param `checkIfFileExists` no longer exists. It is replaced by `fileLen` with a similar logic, but enhanced by _one more_ possibility.

One user immediatly responded, he was able to circumvent, but there may be more out there who **do not intend** to change their code to replace a `true` / `false` by either `0`or `-1`, respectively.

Perhaps it would be a good idea to bring back the old methods so any users out there will not be hurt. And to have **one new** method with the `fileLength` instead of the `checkIfFileExists`, which can be called by the old ones. Discussion maybe?

You need to regenerate the Wiki, by the way: In Async OpenRead, there was a double "token" parm in the /// comments.

### DIFF
--- a/FluentFTP/Client/FtpClient_Stream.cs
+++ b/FluentFTP/Client/FtpClient_Stream.cs
@@ -1378,7 +1378,6 @@ namespace FluentFTP {
 		/// </summary>
 		/// <param name="path">The full or relative path of the file</param>
 		/// <param name="token">The token that can be used to cancel the entire process</param>
-		/// <param name="token">The token that can be used to cancel the entire process</param>
 		/// <returns>A readable stream of the remote file</returns>
 		public Task<Stream> OpenReadAsync(string path, CancellationToken token = default(CancellationToken)) {
 			return OpenReadAsync(path, FtpDataType.Binary, 0, 0, token);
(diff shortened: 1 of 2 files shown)


robinrodricks:
I agree. This restores the old behavior and users can optionally upgrade if they want to.

robinrodricks:
Regenerating the wiki is a bit a hard; it was a one time process to generate it from CSDoc comments, but I'll see what I can do.

robinrodricks:
Every time we have breaking changes in the API I am forced to bump the major version which I did last time, but since this update does not break anything (as per my understanding of your comments), I will publish a minor instead.

robinrodricks:
Published.

https://www.nuget.org/packages/FluentFTP/36.1.0

FanDjango:
Ok, thx for your patience.

I will further monitor for any problems on this.

> I will publish a minor instead

Yes, definitely. Actually the code is back to where it was before, but since you can't go back, all is good.

> Regenerating the wiki is a bit a hard

Ahh, I see. No problem, I will put together the _single_ little correction to the old text and the two new ones for the new functions, to make life easier for you.

> I agree. This restores the old behavior...

I am sorry I did not see the low level funcs in the WIki and blatantly removed them.

But I do have a question re. their defs: **Some of them**, but only a few, are decorated as _virtual_. Any reason not all or none? Seems strange?

PR #837, opened 2022-02-07
Some ArgumentExceptions take `paramName` as the first parameter

jnyrup:
Some places the arguments to `ArgumentNullException` and `ArgumentOutOfRangeException` where incorrect.

The constructors are defined as follow:
* `public ArgumentNullException (string paramName)`
* `public ArgumentNullException (string paramName, string message)`
* `public ArgumentOutOfRangeException(string paramName)`
* `public ArgumentOutOfRangeException(string paramName, string message)`


robinrodricks:
Good work, thanks for the fixes!

robinrodricks:
Released.

PR #838, opened 2022-02-07
Dispose CancellationTokenSource

jnyrup:
This PR disposes the `CancellationTokenSource` added in #822.
It is also only necessary to create a single `CancellationTokenSource`.


robinrodricks:
Great work, thanks!

robinrodricks:
Released.

PR #840, opened 2022-02-07
Use Task factory methods instead of going `async`

jnyrup:
The `async` keyword makes the compiler add a bunch of code to create the underlying state machine that drives asynchronous methods in C#.

You can see [here](https://sharplab.io/#v2:D4AQTAjAsAUCAMACEECsBuWsQGZlkQGFYBvWRC5PEADmQDYAeAIwHtWAbAPkQEEBnAJ4A7AMYBpAKaCA7qwBOAEwAUASnKUyMSjuQB2RADMAhh36TM2ygF8NFO1QYt23BgDF5rALYAlSfwBXDgAXNQctXUoQAxB6ADoPbz9AkOUTM0lVSx1bGFzYIA==) what the compiler generated code is.

For a longer explanation of the difference, see e.g https://stackoverflow.com/questions/52146203/async-method-with-no-await-vs-task-fromresult

### DIFF
--- a/FluentFTP/Client/FtpClient_Hash.cs
+++ b/FluentFTP/Client/FtpClient_Hash.cs
@@ -561,40 +561,41 @@ namespace FluentFTP {
 
 #if ASYNC
 		[ObsoleteAttribute("Use GetChecksum instead and pass the algorithm type that you need. Or use CompareFile.", true)]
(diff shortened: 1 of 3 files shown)


robinrodricks:
Released.

jnyrup:
Sorry about the follow-up commit you had to made. 
I cannot compile the net20, net35 and net45 locally, so I disabled those while making the changes.

PR #843, opened 2022-02-09
Invalidate PWD cache in ASYNC code also, this was missing

### LINKED ISSUE #842: SetWorkingDirectoryAsync doesn't set working directory
**FluentFTP Version:** 37.0.0

I've recently updated from version 33.2.0 to 37.0.0 and noticed that `SetWorkingDirectoryAsync` no longer sets the working directory, for example when calling `GetListingAsync()` after calling `SetWorkingDirectoryAsync(path)` returns the files and folders from the root directory, not the files from the path passed into `SetWorkingDirectoryAsync`. The non-async method `SetWorkingDirectory` works as expected.

FanDjango:
This should fix #842. Once again fell into the trap of a missing ASYNC duplicate of a change.

### DIFF
--- a/FluentFTP/Client/FtpClient_Stream.cs
+++ b/FluentFTP/Client/FtpClient_Stream.cs
@@ -162,6 +162,11 @@ namespace FluentFTP {
 				commandTxt = "PASS ***";
 			}
 
+			// A CWD will invalidate the cached value.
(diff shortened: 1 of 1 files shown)


robinrodricks:
Released.

PR #847, opened 2022-02-17
Fix #845

### LINKED ISSUE #845: Unexpected download "success" when internet cuts out
**FluentFTP Version:** 37.0.0

I'm using the method "DownloadFile" to download a file over FTP. It is working just fine in the usual situations so I'm trying a few odd cases. To know whether or not the download finished properly, I'm checking the "FtpStatus" object returned by the method, as indicated in the documentation of the method.

The issue comes when the download starts correctly but the internet cuts out in the middle of it. When this happens and the timeout is reached (in my case, I set it to 60 seconds) the method returns "FtpStatus.Success" just like if the download finished properly. The downloaded file is obviously not complete when checking its size and contents. This returned value is consistent no matter the amount of attempts, the target file or the target server.

Is this intended behaviour? If it's the case, am I missing the proper way to check if the download finished successfully? I cannot do a checksum as the server I'm connecting to doesn't support it, and I need to know whether the download finished correctly to log any incident properly.

FanDjango:
See #845

### DIFF
--- a/FluentFTP/Client/FtpClient_FileDownload.cs
+++ b/FluentFTP/Client/FtpClient_FileDownload.cs
@@ -775,7 +775,7 @@ namespace FluentFTP {
 
 						// fix: attempting to download data after we reached the end of the stream
 						// often throws a timeout exception, so we silently absorb that here
-						if (offset >= fileLen) {
+						if (offset >= fileLen && !readToEnd) {
 							break;
 						}
 						else {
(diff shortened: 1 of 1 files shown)


robinrodricks:
I just want to say a big thank you for debugging this issue so extensively and collaborating with other users to do so. You have been very helpful of late and both me and the community is grateful.

robinrodricks:
Released.
https://www.nuget.org/packages/FluentFTP/37.0.2

PR #848, opened 2022-02-17
Fix z/OS IsRoot() and some comment typos

FanDjango:
Since moving this to the server specific handler, needs to handle z/OS unix realm as well. Also the z/OS realm was not handled correctly in all cases.


robinrodricks:
Thanks for constantly improving this!

robinrodricks:
Released.
https://www.nuget.org/packages/FluentFTP/37.0.2

PR #849, opened 2022-02-18
z/OS GetListing(): Handle "+++++" in Used collumn

FanDjango:
When the number of tracks used by a z/OS Dataset exceeds "99999" (i.e. 5 digits), the size is depicted as "+++++". This caused the size calculation to fail for obvious reasons. (Exception). In this case we use the absolute maximum size that is possible for a single volume dataset, which of course is an abominable overestimation, thanks IBM for such a stupid limit of 5 chars for the "Used" collumn.

```
//Volume Unit    Referred Ext Used Recfm Lrecl BlkSz Dsorg Dsname   
//ANSYBG 3390   2020/01/03  1   15  VB   32756 32760  PS  $.ADATA.XAA
//ANSYBH 3390   2022/02/18  1+++++  VBS  32767 27966  PS  $.BDATA.XBB
```


robinrodricks:
Great work, thanks!

Issue #771, opened 2021-10-28
FileZilla FTPS - TLS fails on data upload.


TrevorLinkgreen:
**FTP OS:** Windows

**FTP Server:** FileZilla (latest  (1.0.1)

**Computer OS:** Windows

**FluentFTP Version:** 35.0.5

Attempting to upload a file to FileZilla from C# (.net 3.1) using explicit TLS fails to perform TLS handshake (note "Connect" is successful)

This can be replicated every time (I've done it on 3 machines) by debugging the following code after installing FileZilla on a windows machine.

FileZilla settings - Under FTP Server
 - Protocol: "Require explicit FTP over TLS"

I have also tried downloading the certificates via openssl and added them in to the code below, but with the same results.

```
 using var client = new FtpClient("127.0.0.1", 21, "orderease", "xxx")
            {
                SocketKeepAlive = true,
                ValidateAnyCertificate = true,
                SslProtocols = SslProtocols.Tls12,
                EncryptionMode = FtpEncryptionMode.Explicit,
                DataConnectionType = FtpDataConnectionType.PASV,
            };


            var ftpLog = new StringBuilder();
            client.OnLogEvent = (level, s) => { ftpLog.AppendLine(level + " - " + s); };
            client.Connect();
            client.SetWorkingDirectory("/GMS-EDI/GMS_Order");
            
            var response = client.Upload(Encoding.ASCII.GetBytes(content), $"{fileName}.txt");
            if (response == FtpStatus.Failed)
            {
                log.LogError("Failed to upload ");
                log.LogError(ftpLog.ToString());
            }

            client.Disconnect();

```

**Logs :**
Filezilla Logs
```
2021-10-28T18:08:42.299Z >> [FTP Session 26 127.0.0.1 orderease] SYST
2021-10-28T18:08:42.299Z << [FTP Session 26 127.0.0.1 orderease] 215 UNIX emulated by FileZilla.
2021-10-28T18:08:49.043Z >> [FTP Session 26 127.0.0.1 orderease] CWD /GMS-EDI/GMS_Order
2021-10-28T18:08:49.043Z << [FTP Session 26 127.0.0.1 orderease] 250 CWD command successful
2021-10-28T18:08:52.655Z >> [FTP Session 26 127.0.0.1 orderease] PWD
2021-10-28T18:08:52.655Z << [FTP Session 26 127.0.0.1 orderease] 257 "/GMS-EDI/GMS_Order" is current directory.
[...]

PR #854, opened 2022-03-09
#771 Fixed an issue where the TLS socket would be incorrectly closed,…

michael-hoedl:
... which led to an error with a FileZilla FTP Server

In this PR, the SSL Stream gets closed first. Also, if the method is available, the SSL Stream gets shutdown and will not be disposed in a hard way, which can lead to an error on FileZilla servers. 
The socket will also send a shutdown signal, so that the other endpoint knows, that the connection will be closed.

### DIFF
--- a/FluentFTP/Streams/FtpSocketStream.cs
+++ b/FluentFTP/Streams/FtpSocketStream.cs
@@ -731,11 +731,31 @@ namespace FluentFTP {
 			}
 
 #if !NO_SSL
+			if (m_sslStream != null)
+			{
+				try
+				{
+#if NET5_0_OR_GREATER || NETSTANDARD2_1_OR_GREATER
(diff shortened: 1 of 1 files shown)


guyulmaz:
is this merged and released?

robinrodricks:
@guyulmaz It has been merged as you can see on the top. You can always verify that the fix is present by opening the codebase :

https://github.com/robinrodricks/FluentFTP/blob/master/FluentFTP/Streams/FtpSocketStream.cs#L738

robinrodricks:
@Noxxys and @guyulmaz 

Forgot to release this. Released now!

https://www.nuget.org/packages/FluentFTP/37.0.4

Noxxys:
@robinrodricks thank you, it works in the new version.

robinrodricks:
Awesome!

PR #862, opened 2022-04-13
Added localPath to log message

tryallthethings:
I've had a couple of uploads fail today with the error "File does not exist". I'm using the UploadFiles() function to upload a bunch of files at once. Since I didn't see any way of retrieving a list of failed files, my approach is to filter the log for messages and create my own list.

Regarding the problem itself: I suppose it's an issue with a path being too long (running the code on a Windows system). Since I can only access the system during certain times, I can't provide any more details at the moment. I'll open a separate issue if necessary.

Feel free to adjust my changes if necessary.


robinrodricks:
Sorry for the long delay and thanks a lot for your contribution!

PR #871, opened 2022-05-25
Fix regex for files with whitespaces

simonefil:
This fixes a bug that won't allow a correct checksum for files with whitespaces in their name (or in parent directories)


robinrodricks:
Thanks for this!

robinrodricks:
Released now!

https://www.nuget.org/packages/FluentFTP/37.0.4

PR #893, opened 2022-07-19
Fixes error with downloading while using SOCKS4/SOCKS4a proxy

fire-lizard:
Fixes error while downloading file through SOCKS4 / SOCKS4a proxy


robinrodricks:
Can you explain how this fix works or what it does?

fire-lizard:
It seems that when downloading works through SOCKS4 proxy, you library tries to restart download with the restart parameter **after** length of the _**already downloaded**_ file (i.e. if we download, say, test.xml with 1000 bytes, it restarts with restart parameter 1000 or 1001 after file is successfully downloaded and it causes error. Tested on FileZilla and WindowsIIS. Happens of TitanFTP too, but TitanFTP is smart enough to handle that thing so download completes normally.

robinrodricks:
Please check if this works for you:

https://www.nuget.org/packages/FluentFTP/37.1.2

PR #897, opened 2022-07-25
SOCKS5 Support with Username/Password Authentication

ilc-ilopez:
Our use case needed username/password authentication against a SOCKS5 proxy.

We realized that the previous pull request did not implement RFC1929.
https://datatracker.ietf.org/doc/html/rfc1929

We saw where/how it was not doing it correctly, and implemented RFC1929 as best as we could.

Then we tested in a separate project (in our dev-environment, we only care about .net core 3.1) and tested the private resources we were working with. Then confirmed everything was working, and prepared this pull request.

Let us know if you have any questions.

Thanks.


robinrodricks:
This is fantastic! Thanks so much for this huge contribution!

robinrodricks:
FYI your website is down.

https://israellopezconsulting.com/

image

ilc-ilopez:
@robinrodricks - for stupid reasons, we have some region blocking with our domain because of some abuse, hence why we are on cloudflare. We are definitely up :P.

Thanks for taking the contribution, I hope the community can use the contribution.

PR #902, opened 2022-08-02
Configure containers.

tommysor:
Correct namespace for SkippableAttributes.
Correct chaining of TestcontainerBuilder.
Change Fixture to CollectionFixture.
Refactor timeout test.
Add SocketException to accept criteria (occurs when running in GitHub Actions).

This PR is a few fixes to get the TestcontainersBuilder back to working in new structure.
Change Fixture to CollectionFixture, from ClassFixture, to keep the same container instance running for all test classes.

### DIFF
--- a/.github/workflows_WIP/pull_request.yml
+++ b/.github/workflows_WIP/pull_request.yml
@@ -13,9 +13,11 @@ jobs:
     strategy:
       matrix:
         include:
-        - ftp: pure-ftpd
+        - ftp: pureftpd
           os: ubuntu-latest
         - ftp: vsftpd
+          os: ubuntu-latest
+        - ftp: pyftpdlib
           os: ubuntu-latest
           
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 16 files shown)


tommysor:
>  Instead of using the available dockers I made the choice of copying the source so that we are decoupled from any breaking changes in those repos or any images that go defunct (may not be the best decision)
Being able to change the Dockerfile would be nice.
I think it would be easier to get the servers up with customized Dockerfiles.
Simple example is the proftp that requires a volume mounted to work, would be nice to remove that requirement.
I haven't experimented with it yet, but seems there is a `ImageFromDockerfileBuilder` in Testcontainers-dotnet that might work to build image from Dockerfile.

tommysor:
> \> This method will run 1 single type of ftp-server when run on developer computer.
I have one question. The whole integration test system has been designed to be run on Github Actions. What if I wanted to run it locally? How do I create the "matrix" of tests within Visual Studio so it can be run in such a way?

I don't have a solution for that.
I'm fairly certain that it would be possible to extend xUnit behavior to do it, though I'm not sure how.
Though that would likely remove the ability to have GitHub Actions run each server in parallel.

I tried https://github.com/nektos/act to run GitHub Actions on my PC.
My quick experiment failed. Possibly it tried to start 2 workers that both wanted to start docker ftp-server and colliding.
There might be a solution involving this or similar tool.

robinrodricks:
Wow. Just saw this. Thanks a ton!!

robinrodricks:
> This PR is a few fixes to get the TestcontainersBuilder back to working in new structure.

Thanks very much!!

robinrodricks:
> Being able to change the Dockerfile would be nice.
> I think it would be easier to get the servers up with customized Dockerfiles.

Then let me bring back my forks.

robinrodricks:
Added [customized dockerfiles](https://github.com/robinrodricks/FluentFTP/tree/master/FluentFTP.Dockers), please feel free to edit them as you wish!

Updated [steps](https://github.com/robinrodricks/FluentFTP/wiki/Automated-Testing#steps).

PR #905, opened 2022-08-04
Catch errors in async tests.

tommysor:
Bugfix observe exceptions from async tests.
Since the Task.Run was not awaited the tests would ignore exceptions.
Changed tests to async Task to remove need for Task.Run and instead await the tests directly.


robinrodricks:
Brilliant, thanks!

PR #913, opened 2022-08-12
Make vsftpd Dockerfile work

tommysor:
Make Dispose wait for completion to avoid starting next container before the previous is shut down.
Add finalizer to call Dispose if an exception is thrown in constructor.


robinrodricks:
Hi, thanks for this. I'm ok with the docker file changes, but I see you have removed the system command. So how will the docker be shut down?

tommysor:
> how will the docker be shut down?

The docker is shut down when  the container is disposed `_container?.DisposeAsync()`.
I have tested this running "all tests" and observed that the container for each server was removed before the next server was started.

robinrodricks:
Ok, thank you.

PR #917, opened 2022-08-19
Fix fix #909

### LINKED ISSUE #909: Inconsistent implementations of GetListing for z/OS
Hi @FanDjango, 

I noticed this difference in behavior. Please tell me which one is correct.

A: GetListing
```
			// Only disable the GetAbsolutePath(path) if z/OS
			// Note: "TEST.TST" is a "path" that does not start with a slash
			// This could be a unix file on z/OS OR a classic CWD relative dataset
			// Both of these work with the z/OS FTP server LIST command
			if (ServerType != FtpServer.IBMzOSFTP || path == null || path.StartsWith("/"))
			{
				// calc the absolute filepath
				path = GetAbsolutePath(path);
			}
```
B: GetListingAsync
```
			// calc path to request
			path = await GetAbsolutePathAsync(path, token);
```

FanDjango:
Refactoring of `IbmZosFtpServer.cs`:

Small oversight/typo, discovered on further testing.

`ConvertListingPath` was never being called. Thus, file sizes were only correct for z/OS unix files.

### DIFF
--- a/FluentFTP/Servers/Handlers/IbmZosFtpServer.cs
+++ b/FluentFTP/Servers/Handlers/IbmZosFtpServer.cs
@@ -168,7 +168,7 @@ namespace FluentFTP.Servers.Handlers {
 		/// <summary>
 		/// Return true if the path should be converted to an absolute path.
 		/// </summary>
-		public virtual bool ConvertListingPath(string path) {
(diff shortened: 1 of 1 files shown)


robinrodricks:
Thank you!

PR #919, opened 2022-08-20
RemotePaths: Add some special cases for item.Fullname

FanDjango:
Some special cases were missing when calculation item.Fullname


robinrodricks:
Thanks!

PR #925, opened 2022-08-21
Eliminate redundant ServerHandler queries

FanDjango:
This gives us a `IsCustomGetAbsolutePath()` query **INSIDE** the regular `GetAbsolutePath()` and thus eliminates a lot of `if (ServerHandler...` code outside.

**Note:** The regular `GetAbsolutePath(...)` **still** contains a second ServerHandler function, `IsAbsolutePath(...)` which was created for **OpenVMS**. This can probably be migrated to the new custom mechanism, but since I cannot test it, I have left it in place for the time being.

The functionality of `IsAbsolutePath(...)` sadly was not sufficient to also use for z/OS, although it looked promising at first, early simple tests actually worked, but not all special cases.

There was also some "old" z/OS (MVS) code in there that has been redundant for some time.

An additional benefit of this change, is that **all** callers of `GetAbsolutePath(...)` will now go through the ServerHandler and not only those that I changed for z/OS some time ago.

### DIFF
--- a/FluentFTP/Client/FtpClient_FileManagement.cs
+++ b/FluentFTP/Client/FtpClient_FileManagement.cs
@@ -105,13 +105,7 @@ namespace FluentFTP {
 
 				LogFunc(nameof(FileExists), new object[] { path });
 
-				// A check for path.StartsWith("/") tells us, even if it is z/OS, we can use the normal unix logic
-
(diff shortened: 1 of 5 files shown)


robinrodricks:
Awesome, thanks!

robinrodricks:
Released!

https://www.nuget.org/packages/FluentFTP/39.4.0

Issue #903, opened 2022-08-04
Some test dockers fail to run


robinrodricks:
**OS:** Windows 10 x64

Docker ID mentioned along with logs taken from the docker output from within the container.

------------

pureftpd:fluentftp
```
/bin/sh: 1: /run.sh: not found
```

crushftp:fluentftp
```
/var/opt/setup.sh: line 2: $'\r': command not found

/var/opt/setup.sh: line 4: $'\r': command not found

/var/opt/setup.sh: line 17: syntax error near unexpected token `elif'

'var/opt/setup.sh: line 17: `elif [ -z ${CRUSH_ADMIN_PASSWORD} ] && [ ! -f ${CRUSH_FTP_BASE_DIR}/admin_user_set ]; then
```

proftpd:fluentftp
```
standard_init_linux.go:228: exec user process caused: no such file or directory
```

filezilla:fluentftp

```
[s6-init] making user provided files available at /var/run/s6/etc...exited 0.

[s6-init] ensuring user provided files have correct perms...exited 0.

[fix-attrs.d] applying ownership & permissions fixes...

[fix-attrs.d] done.

[cont-init.d] executing container initialization scripts...

[cont-init.d] 00-app-niceness.sh: executing...

[cont-init.d] 00-app-niceness.sh: exited 0.

[cont-init.d] 00-app-script.sh: executing...

[cont-init.d] 00-app-script.sh: exited 0.

[cont-init.d] 00-app-user-map.sh: executing...

[cont-init.d] 00-app-user-map.sh: exited 0.

[cont-init.d] 00-clean-logmonitor-states.sh: executing...

[cont-init.d] 00-clean-logmonitor-states.sh: exited 0.

[cont-init.d] 00-clean-tmp-dir.sh: executing...

[cont-init.d] 00-clean-tmp-dir.sh: exited 0.

[cont-init.d] 00-set-app-deps.sh: executing...

[cont-init.d] 00-set-app-deps.sh: exited 0.

[cont-init.d] 00-set-home.sh: executing...

[cont-init.d] 00-set-home.sh: exited 0.

[cont-init.d] 00-take-config-ownership.sh: executing...

[cont-init.d] 00-take-config-ownership.sh: exited 0.

[cont-init.d] 00-xdg-runtime-dir.sh: executing...

[cont-init.d] 00-xdg-runtime-dir.sh: exited 0.

[cont-init.d] 10-certs.sh: executing...

[...]

Issue #816, opened 2022-01-02
How to test FluentFTP against all supported FTP servers?


robinrodricks:
**UPDATE: We now have extensive automated integration testing for many FTP servers. See the [Automated Testing](https://github.com/robinrodricks/FluentFTP/wiki/Automated-Testing) page for more info. You can help by adding more FTP server dockers!**

-----

We now support over [two dozen servers](https://github.com/robinrodricks/FluentFTP/wiki/Server-Information#faqs), but our tests have not kept up the pace. We don't have a way to test FTP functionality against all types of FTP servers. Each server has different handling, so its not possible to setup one server alone, we need to setup most/all of these servers in some way and create automated tests.

I ask you, community, do you have any suggestions?

**Some ideas:**
1. Create docker containers for individual FTP servers and create setup scripts to help setup a full test-suite locally
2. Create steps to setup servers on local machines so advanced developers can set it up manually
3. Create automated tests for all main methods, against all FTP servers that we support (or at least the major ones)
4. Create a few test files that test various cases of upload and download (small files, large files, binary/text)

**Methods/workflows that need to be tested**
- Connection/disconnection
- File upload/download in binary and ASCII mode
- Folder upload/download
- Get listing in recursive and non-recursive modes
- Getting file exists, modified time, chmod, permissions
- File transfer with weird file names (including special chars, spaces, unicode chars)
- Get hash/checksum
- Renaming/moving files and folders

FanDjango:
Hmmm - in the case of IBM z/OS and OS/400, these are licensed software and need specialised hardware. There actually IS an emulator for IBM System/z, which one could run in a docker container but there is no operating system available that needs no license and provides a modern z/OS CS FTP server. So no dice for these two server types. I currently run a test program against my own z/OS system, which is not publicly available.

robinrodricks:
Wonderful, thanks for this info @FanDjango !

datvm:
I am also looking into this though at a smaller scale. For e.g. I want to test on a server that support TLS only, and one that do not have any; one that allow anonymous and one the requires login.

Right now I simply do manual test by switching settings from a FileZilla Server. Will just leave a comment here to learn any useful trick.

tommysor:
I would suggest a somewhat parallel path with more proper unit tests. Which might be the most practical way to (mostly) test some implementations (like OS/400, as FanDjango points out).

[...]

PR #926, opened 2022-08-21
Cleanup docker tests

tommysor:
Remove docker files that does not work.
All the remaining integration tests now works.

Add a few listing tests.

Handles all issues in #903 (of course, still missing some of those servers, and others from #816).


FanDjango:
Super, thanks -

I also needed to remove those 3 fails from
`IntegrationTests.cs` and for completeness from `DockerFtpContainerIndex.cs` to get a green screen:

image

I am impressed to no end.

robinrodricks:
Sad to lose 3 dockers, but if it does not work, ok. Thanks for the efforts @tommysor

robinrodricks:
Released!

https://www.nuget.org/packages/FluentFTP/39.4.0

PR #928, opened 2022-08-23
Rest of the changes

FanDjango:
This change finalizes my attempts to move the z/OS logic to a ServerHandler. Once you get some practice (on the previous ones), you can move faster... Success!

Added ServerHandler(s):

For `FtpClient_FileManagement`

**FileExists(..)**
`ServerHandler.DontUseSizeEvenIfCapable`
`ServerHandler.DontUseMdtmEvenIfCapable`

`ServerHandler.IsCustomFileExists`
`ServerHandler.FileExists`

For `FtpClient_FileUpload`

**UploadFiles(..)**
More or less complete rewrite of the path/directory handling logic so to use the new ServerHandlers and the following new ones:

`GetAbsoluteDir()`
`GetAbsoluteFilePath()'

For `FtpListParser`

`ServerHandler.IsCustomCalculateFullFtpPath`
`ServerHandler.CalculateFullFtpPath`

and so on and on.. it is a lot, I realize, but quite systematic:

1. A number of new serverhandler routines, some are boolean, some return values, some are even conditional - using either nullable value types as indicators or as in the case of a string return value, a value of null to indicate: Let the base handler do it, I refuse to handle it. Oh, and wherever needed, also for ASYNC.

2. Replace every usage of the server type enum with the appropriate logic.

3. Recode UploadFiles from the changes of #741 to a more consistant approach to path/filename handling.

image

Looking OK. 

In the next period of time, I will endeavour to improve the comments and docs for FtpBaseServer.

### DIFF
--- a/FluentFTP/Client/FtpClient_FileManagement.cs
+++ b/FluentFTP/Client/FtpClient_FileManagement.cs
@@ -109,9 +109,7 @@ namespace FluentFTP {
 
 				// since FTP does not include a specific command to check if a file exists
 				// here we check if file exists by attempting to get its filesize (SIZE)
-				// If z/OS: Do not do SIZE, unless we have a leading slash
-				if (HasFeature(FtpCapability.SIZE) && (ServerType != FtpServer.IBMzOSFTP || path.StartsWith("/"))) {
-
+				if (HasFeature(FtpCapability.SIZE) && ServerHandler != null && !ServerHandler.DontUseSizeEvenIfCapable(path)) {
 					// Fix #328: get filesize in ASCII or Binary mode as required by server
 					var sizeReply = new FtpSizeReply();
 					GetFileSizeInternal(path, sizeReply, -1);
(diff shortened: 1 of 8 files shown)


robinrodricks:
Looks pretty intensive, thanks for all the work. Sure there are no breaks in the behavior for VMS/zOS?

FanDjango:
**major** ??? Well, I suppose I had it coming to me.

> Sure there are no breaks in the behavior for VMS/zOS?

VMS: No. No way to test it. But I have a few more comments on that later on.
z/OS: Looks very good to me, I have tested it extensively with my local test suite and z/OS System. The only slightly "iffy" place is UploadFiles, where I have tested it in different ways but cannot be sure if the changes handle everything correctly.

That having been said, the changes in "UploadFiles" actually hit **ALL** servers, not only VMS or z/OS. **So do** the changes for "FileExists" and for the listing parser analysis stuff.

Whereas, for VMS, I can actually see one place where I made an unsafe assumption and will correct that later on. After that, I think I am quite confident about VMS, too.

Beyond that, please humour me when I say that maybe I was also hoping for some more eyes to inspect and to review, beyond msyself reviewing it over and over again. I pledge to be available to sort out any error reports ASAP for the next months and you have been quite responsive in the past to handle "sudden" crises. I had the feeling that the users out there are also, if not "forigiving", generally prepared to accept glitches for advances in the software features.

What I **can** do, I suppose, is to separate this change into three smaller ones. The concerns about each of them breaking something remain the same, though.

Or, we can just leave this PR sitting here and it can be used at some point during the refactoring you are planning - because it shows the places where `if ServerType = xxx` needs to be touched.

Call it "Do Not Merge - Collection of necessary changes", maybe.

Mind you: I will be using this version myself from now on.

robinrodricks:
Ok, yolo, thanks!

FanDjango:
As a James Bond fan, I would say yoltomo.

PR #931, opened 2022-08-25
z/OS Listing FullName sometimes incorrect

FanDjango:
#928 is fine but I discovered this type that needs to be fixed. 

It causes the `item.FullName` to be incorrect for any z/OS GetListing.


robinrodricks:
Thanks!

PR #932, opened 2022-08-25
Update FtpClient_Stream.cs

elmar69:


elmar69:
This is an fix to Pull Request #567. As documented there it is not really necessary to close AsyncWaitHandle on unsucessfull connect. Indeed this seems to be harmful, I do get an "System.ObjectDisposedException: Safe handle has been closed" when trying to connect to an offline host.

PR #936, opened 2022-08-27
Handle z/OS Listing single line outputs

FanDjango:
Further tests for `UploadDirectory`, `UploadFiles` and `DownloadFiles` which produced some errrors or failed have shown up a quirk of z/OS LIST command output in the Unix (HFS) Realm:

If listing a directory all is well. But if listing a path that represents a single file (as `GetFileSize` or `FileEsists` will do for z/OS), a single line is output - the title line `total: 1` is missing.

These functions are heavily used by the `UploadDirectory`, `UploadFiles` and `DownloadFiles` methods and if you mix these for z/OS datasets and then for Unix (HFS) files, you will get hard to understand errors.

It is caused by a switch to unix parser and a missing z/OS realm detection. Doesn`t happen often and only if multiple operations are sequenced to the same connection without "resetting/renewing" the client. 

It is interesting to note that this has been observed before and there were "primitive" quick and dirty solutions to this problem ((z/OS)`GetFileSize` first action to reset the current listing parser), no longer needed.


FanDjango:
**This is needed on V40 too!**

robinrodricks:
Please recheck this PR, I have merged the v40 branch into master.

FanDjango:
Ok, you can use it now for the new master

robinrodricks:
Awesome thanks!

PR #938, opened 2022-09-03
Properly override HandshakeAsync in async proxies

jnyrup:


robinrodricks:
My bad, thanks for the fix.

PR #940, opened 2022-09-03
Styling

jnyrup:
As my Visual Studio is set to auto-format on save, it would reformat the braces of this project all the time as FluentFTP uses K&R brace styling as opposed to the default Allman style.
This made it a bit of a pain point for me to contribute, as I had to disable auto-formatter when editing this project.

This PR adds some more styling rules to `.editorconfig` to prevent using IDE defaults.

None of the changes should alter the behavior of FluentFTP.
I've made a commit per type of change to make reviewing easier.


robinrodricks:
Very large PR, but seems to be excellently done. Thanks!

Issue #763, opened 2021-10-08
FTPS Passive Explicit closing stream err 426


SteveB360:
**FTP OS:** Unisys OS 2200 Mainframe

**FTP Server:** Proprietary

**Computer OS:** ? Win 10

**FluentFTP Version:** ? 35.0.5

I'm using FTPS Passive Explicit to connect to the Unisys mainframe. FluentFTP download PASV Explicit on port 21 works. Upload PASV on port 21 doesn't work, Explicit doesn't seem to matter.

                ftpsclient = new FtpClient(host, port, userID, pass);
                ftpsclient.EncryptionMode = FtpEncryptionMode.Explicit;
                ftpsclient.DataConnectionType = FtpDataConnectionType.PASV;
                ftpsclient.ValidateAnyCertificate = true;
                ftpsclient.Connect();
                ftpsclient.Execute("ACCT SPSE");
. . . 
                ftpsclient.UploadDataType = FtpDataType.ASCII;
                xferResult = ftpsclient.UploadFile(local, host, FtpRemoteExists.NoCheck, false, FtpVerify.None);

The xferResult is a fail. So I stepped through the execution. It looks like several buffers are written successfully. Then after the file has been sent, it gets to a chunk of code in FtpClient_FileUpload.UploadFileInternal() around line 895 (below) and gets a reply with an error code 426, that the stream has been closed.

                               upStream.Dispose();

				// FIX : if this is not added, there appears to be "stale data" on the socket
				// listen for a success/failure reply
				try {
					while (!m_threadSafeDataChannels) {
						var status = GetReply();

						// Fix #387: exhaust any NOOP responses (not guaranteed during file transfers)
						if (anyNoop && status.Message != null && status.Message.Contains("NOOP")) {
							continue;
						}

						// Fix #353: if server sends 550 or 5xx the transfer was received but could not be confirmed by the server
						// Fix #509: if server sends 450 or 4xx the transfer was aborted or failed midway
						if (status.Code != null && !status.Success) {
							return FtpStatus.Failed;

From the mainframe log (attached Mainframe Log 2021-10-08.docx) it appears the client closed the stream. 

If I comment out the upStream.Dispose() immediately above on line 894 I stop getting the 426 code from the mainframe. But then I found I am getting an unhandled System.IO.IOException with an Inner exception
{"Unable to read data from the transport connection: A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond.."}

[...]

PR #944, opened 2022-09-08
Fixed issue with 426 error for vsftpd servers

vasyaxparfenov:
Issue: https://github.com/robinrodricks/FluentFTP/issues/763

Accrording to this [acrticle](https://www.suse.com/support/kb/doc/?id=000020678) in order to correctly receive response on upload status from vsftpd servers data connection should be closed using SSL_Shutdown command. This functionality has always been present, but it never kicked in due to dispose order in FtpSocketStream dispose logic. When ssl stream given DisconnectWithShutdown is set to true tried to shutdown SSL Stream underlying stream was already disposed which throwed exception swollowed inside dispose method.

### DIFF
--- a/FluentFTP/Streams/FtpSocketStream.cs
+++ b/FluentFTP/Streams/FtpSocketStream.cs
@@ -31,7 +31,7 @@ namespace FluentFTP {
 		/// to determine if Poll() should be used to test for
 		/// socket connectivity. The socket in this class will
 		/// not know it has been disconnected if the remote host
-		/// closes the connection first. Using Poll() avoids 
+		/// closes the connection first. Using Poll() avoids
 		/// the exception that would be thrown when trying to
 		/// read or write to the disconnected socket.
 		/// </summary>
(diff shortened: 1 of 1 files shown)


VanyaBurdeyniy:
Awesome

vasyaxparfenov:
@robinrodricks Could you please look at this PR when you have free time?

robinrodricks:
Hi, I already looked at it. Thanks for this PR. There have been many changes to the stream handling, almost all of which had to be reverted as it was causing issues for others. Can you run the integration test suite on your machine and attach a screenshot of it working, or, we will have to verify it and then it can be accepted. Also after v40 I am hesitant to make major changes to the core otherwise more people will revert to v39x.

robinrodricks:
If you are sure it applies to all servers, test it. If you are not, then add logic to only trigger if `stream.client.ServerType` is vsftp, and leave the old behaviour for the other servers.

vasyaxparfenov:
@robinrodricks Thank you! I’ve run it against vsftpd version and one that IIS provides. We could also apply this as a patch to older version. This logic has always been there, just wrong order. I’ll run integration test and attach results here!

robinrodricks:
Thanks for your efforts.

vasyaxparfenov:
Here is a screenshot of integration tests run. I've build containers using your docker files (special thanks for fixing ProFtpd docker-entrypoint.sh, it was failing to recognise FTP_LIST env variable when using image from docker hub 🤣). It works for all types of servers.

image

robinrodricks:
Thanks a lot!

robinrodricks:
Released!

FanDjango:
See #996

PR #953, opened 2022-09-15
** DO NOT MERGE** Parser tests problem.

FanDjango:
**DO NOT MERGE**

Hi Robin,

while finagling some tests for the OS400 issue recently, I noticed that the parser tests were not calling any custom server handler routines. This didn't hurt up to now, success was still reported. Only for my special tests did this show up.

So I needed to change the small `new parser` and `new client` setup slightly, to make the new client get a specific server handler.

**But**: This meant changing 'ServerModule` to public (from internal). I don't think that is really what you would like. **Is there any other way to do this?**:

```
			var client = new FtpClient();
			client.ServerHandler = ServerModule.GetServerHandler(FtpServer.IBMOS400FTP);
			var parser = new FtpListParser(client);
			parser.Init(FtpOperatingSystem.IBMOS400, FtpParser.IBMOS400);
```

The only other place where the server handler is actually set, is within a (successful) connect.

** NOTE: Since this is not really needed, perhaps only a comment would suffice, to warn future coders in this test code that the ServerHandlers are not set, so not to surprised when using the test facility to debug a parser.

### DIFF
--- a/FluentFTP.Tests/Unit/ParserTests.cs
+++ b/FluentFTP.Tests/Unit/ParserTests.cs
@@ -1,3 +1,4 @@
+using FluentFTP.Client.Modules;
 using FluentFTP.Helpers;
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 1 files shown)


robinrodricks:
Interesting problem, I'll think on it.

robinrodricks:
@FanDjango Pls send me your email ID at robinrodricks7@gmail.com

FanDjango:
I will insert comments for the time being and put this on the Todo list.

Issue #954, opened 2022-09-16
issue with Active Mode, cannot reuse the ActivePorts, then throw exception


DoraemonYu:
**FluentFTP Version:** reproduced on **v40/v39/v38**
**Framework:** reproduced on **.NET4** and .**NET6**
**FTP Server:** reproduced on **build-in FTP From Microsoft IIS**  and **FileZilla Server**
**FTP Client OS:** reproduced on **Microsoft Window 7**  and  **Ubuntu 18.04**
**FTP Server OS:** reproduced on **Microsoft Windows Server 2016**

---
Dear team,    
Hello. **Passive Mode** is used in many situations, as far as we know. 
However, in the recent projects, we have to use the **Acive Mode** and limit the usage with ports rang. 

Simple source code to reproduce this issue:
```
var ftpClient = new FluentFTP.FtpClient();
ftpClient.Host = "[[replace your ftp server ip here pls]]";
ftpClient.Port = 21;
ftpClient.Config.DataConnectionType = FluentFTP.FtpDataConnectionType.PORT;
ftpClient.Config.ActivePorts = new int[] { 8082 };

ftpClient.UploadFile(@"D:\0.jpg", "/0.jpg", FluentFTP.FtpRemoteExists.Overwrite, true);
ftpClient.UploadFile(@"D:\1.jpg", "/1.jpg", FluentFTP.FtpRemoteExists.Overwrite, true);
ftpClient.UploadFile(@"D:\2.jpg", "/2.jpg", FluentFTP.FtpRemoteExists.Overwrite, true);
```
  Only the `0.jpg` would be uploaded successfully.  The subsequent files would throw the excpetion:
```
No valid active data port available!

This exception was originally thrown at this call stack:
    FluentFTP.FtpClient.StartListeningOnPort(FluentFTP.FtpDataStream)
    FluentFTP.FtpClient.OpenActiveDataStream(FluentFTP.FtpDataConnectionType, string, long)
    FluentFTP.FtpClient.OpenDataStream(string, long)
    FluentFTP.FtpClient.OpenWrite(string, FluentFTP.FtpDataType, long)
```

Logs from FluentFTP:
```
# Connect()
Status:   Connecting to ***:21
Response: 220-FileZilla Server 0.9.60 beta
Response: 220-written by Tim Kosse (tim.kosse@filezilla-project.org)
Response: 220 Please visit https://filezilla-project.org/
Status:   Detected FTP server: FileZilla
Command:  USER ***
[...]

PR #957, opened 2022-09-19
Fix issue 954

FanDjango:
** Active Mode, cannot reuse the ActivePorts **

Please refer to the excellent problem description on #954

Accepting on a listening socket creates a **new** socket. The previous code simply took over the new socket, literally leaving the listening socket sitting there, still listening, for eternity, thus it was no longer eligible when the next request tried to select an available socket.

The new approach is to close this listening socket after accepting from it.

Surprisingly, a shallow copy of the socket suffices to be able to do this.

Passes the intergration tests.


robinrodricks:
Looks great, thanks a lot.

robinrodricks:
Released!

Issue #858, opened 2022-03-22
UploadFile(Stream) calls SIZE when overwrite behavior is set to NoCheck


j-childers:
**FTP OS:** Any

**FTP Server:** Any

**Computer OS:** Any

**FluentFTP Version:** 37.0.2

We have a use case where we send files via FTP that are guaranteed to have unique names, and want to send them as quickly as possible.  The destination only supports FTP.  So we desire as few round-trips as possible.  While optimizing the transfer we noticed that even when we tell it not to check for any existing file, the SIZE command is still sent before starting the transfer.

It appears that `OpenWrite` checks the remote file size because `UploadFileInternal` doesn't set the `fileLen` parameter to a non-zero value.  It would be nice to have a way to prevent this by exposing `fileExistsKnown` as a parameter to `UploadFile`, passing this to UploadFileInternal, and setting the `remoteFileLen` to the local stream length if `existsMode` is `NoCheck`, `fileExists` is `false`, and `fileExistsKnown` is `true`.

robinrodricks:
I guess you are asking for a low level method when you know the filesize. We have such methods (`UploadFileInternal`) which you can access using reflection or by subclassing the FtpClient class.

j-childers:
Not exactly.  It's not about knowing the size or not.  We know the file does not exist at all (i.e. there is no size).  Subclassing is an option, and when the project picks up again next week, I'll probably go that route in order to shave a few milliseconds off.

Just thought it might benefit others (however slightly) since it seems inefficient to send a SIZE command when the caller explicitly says not to check for the file's existence.

robinrodricks:
> Just thought it might benefit others (however slightly) since it seems inefficient to send a SIZE command when the caller explicitly says not to check for the file's existence.

Then its probably a bug. Which enum value were you using for the RemoteExists parameter?

j-childers:
Here's how we've calling it:

`_FtpClient.Upload(stream, $"{remoteFolderName}/{remoteFileName}", FtpRemoteExists.NoCheck, false, uploadProgressAction)`

FanDjango:
> It would be nice to have a way to prevent this by exposing fileExistsKnown as a parameter to UploadFile, passing this to UploadFileInternal, and setting the remoteFileLen to the local stream length if existsMode is NoCheck, fileExists is false, and fileExistsKnown is true.

Yes

FanDjango:
[...]

PR #959, opened 2022-09-20
Optimize SIZE command usage for UploadFile NoCheck AND OverWrite (Issue #858)

FanDjango:
Please see the tail end of Issue #858 

After adding the needed code, I realised that a refactor of `UploadFileInternal` would optimise the use of the SIZE command **not only** for the case when `NoCheck` is requested, but **also** for the case when `OverWrite` is active.

Here is what it now looks like if `Overwrite` is requested (only one SIZE command, previously there were **two!**):
```
>         UploadFile("D:\temp\test1", "/home/mike/test1", Overwrite, False, None)
>         FileExists("/home/mike/test1")
Command:  SIZE /home/mike/test1
Response: 550 /home/mike/test1: No such file or directory
>         OpenWrite("/home/mike/test1", Binary)
>         OpenActiveDataStream(PORT, "STOR /home/mike/test1", 0)
Command:  PORT 192,168,1,147,31,146
Response: 200 PORT command successful
Command:  STOR /home/mike/test1
Response: 150 Opening BINARY mode data connection for /home/mike/test1
Response: 226 Transfer complete
```
And for `NoCheck` (no SIZE command, previously there was **one!**):
```
>         UploadFile("D:\temp\test1", "/home/mike/test1", NoCheck, False, None)
>         OpenWrite("/home/mike/test1", Binary)
Command:  TYPE I
Response: 200 Type set to I
>         OpenActiveDataStream(PORT, "STOR /home/mike/test1", 0)
Command:  PORT 192,168,1,147,31,146
Response: 200 PORT command successful
Command:  STOR /home/mike/test1
Response: 150 Opening BINARY mode data connection for /home/mike/test1
Response: 226 Transfer complete
```

### DIFF
--- a/FluentFTP/Client/SyncClient/UploadFileInternal.cs
+++ b/FluentFTP/Client/SyncClient/UploadFileInternal.cs
@@ -27,7 +27,7 @@ namespace FluentFTP {
 			}
 
 			try {
-				long localPosition = 0, remotePosition = 0, remoteFileLen = 0;
+				long localPosition = 0, remotePosition = 0, remoteFileLen = -1;
 
 				// check if the file exists, and skip, overwrite or append
 				if (existsMode == FtpRemoteExists.NoCheck) {
--- a/FluentFTP/Client/AsyncClient/UploadFileInternal.cs
+++ b/FluentFTP/Client/AsyncClient/UploadFileInternal.cs
@@ -27,7 +27,7 @@ namespace FluentFTP {
 			}
 
 			try {
-				long localPosition = 0, remotePosition = 0, remoteFileLen = 0;
+				long localPosition = 0, remotePosition = 0, remoteFileLen = -1;
 
 				// check if the file exists, and skip, overwrite or append
 				if (existsMode == FtpRemoteExists.NoCheck) {


FluentFTP/Client/SyncClient/UploadFileInternal.cs
```diff
@@ -27,7 +27,14 @@ public partial class FtpClient {
...
 			try {
-				long localPosition = 0, remotePosition = 0, remoteFileLen = 0;
+				long localPosition = 0, remotePosition = 0, localFileLen = 0, remoteFileLen = 0;
+
+				// calc local file len
+				localFileLen = fileData.Length;
+
+				// for most of the below cases,
+				// apart from ResumeNoCheck, AddToEndNoCheck, Resume and AddToEnd
+				remoteFileLen = localFileLen;
```
robinrodricks, line 37: Looks suspect...

robinrodricks:
This looks excellent, However the idea of assuming the remote file length seems suspect to me. Can you elaborate, I mean, it won't mess up any of the upload scenarios, will it?

FluentFTP/Client/SyncClient/UploadFileInternal.cs
FanDjango: Ok.
```
				// for most of the below cases,
				// apart from ResumeNoCheck, AddToEndNoCheck, Resume and AddToEnd
```
Lets look at each case, and also at the code:

**NoCheck**
For all FTP servers that I know of, not checking for the existence of the file on the server means you are risking a failure if it exists. It is for those people who **know** that the check is unnecessary and therefore, by implication, the file **does not exist**. Therefore there is no `remoteLength` to be gotten by the `GetSize` in `OpenWrite`, it would have returned -1 and no `SetLength` would have occurred at all. But since we know the local file size, isn't this the size of the stream that we need to write this file from afresh?

**Overwrite**
Same as NoCheck: Initially, a `FileExists` (sadly, boiling down to a `SIZE` command) is done, solely to determine if a `DeleteFile` is needed. After that delete, the file does not exist, or did not exist in the first place. We are back in the same logic as with **NoCheck**

All other cases: `remoteFileLen` is recalculated, the initial value from the top ( = `localFileLen` ) is discarded. Effectively, this is the old code, no change.

**Therefore**
The only real change is that a `SetLength` is now done on the stream where the **whole** local file is uploaded against a server where there is no exisiting file. Before, that `SetLength` was not done at all.

This is the point of discussion: Is it harmful to do this `SetLength` in this case.

FanDjango:
See the above conversation. And ASYNC is still missing, of course.

robinrodricks:
Ok, you seem to have thought this through. Have you tested the case where we are trying to upload a file but:

**Overwrite:**
1. Case 1 - The remote file does not exist (full upload)
2. Case 2 - The remote file does exist (so it should overwrite the file, preferably using the delete command)

**AddToEnd:**
1. Case 1 - The remote file does not exist (full upload)
2. Case 2 - The remote file does exist (so it should add the remaining parts)

Apologies if this is too much over checking on my part, but since we don't have any decent tests for these cases (we probably should create them), I need to check manually.

FanDjango:
> Apologies if this is too much over checking on my part, but since we don't have any decent tests for these cases (we probably should create them), I need to check manually.

No need to apologize.

> Overwrite, AddToEnd

Looking at the code:

All modes that do **not** have a "NoCheck" in their name will initially do this:
```
					// check if the remote file exists
					if (!fileExistsKnown) {
						fileExists = FileExists(remotePath);
					}
```
Skip will skip if it exists.

OverWrite will then do:
```
						if (fileExists) {
							DeleteFile(remotePath);
						}
```
and go on to do a full transfer.

Resume and AddToEnd will **only** modify the stream length and position if the file exists, otherwise it will be a full transfer.

When it next rains I will try to test AddToEnd/Resume to confirm. Leave this hanging here until then. Maybe we could offer this PR change to the issue owner as a fix for him to test it too?

FanDjango:
I actually did find a bug, thanks to doing some testing on `AddToEnd`. Come on - own up: You tested it!

robinrodricks:
@FanDjango  thanks for the testing, let me know when you are done and I should merge.

FanDjango:
I am going to rebase and bring all the changes together, then I will go over the changes in detail and also compare sync and async.

FanDjango:
Ready now.

robinrodricks:
Its just showing 1 line changed, variable value from 0 to -1, is this it?

FanDjango:
Yes. Such irony, 😃

robinrodricks:
Released!

PR #960, opened 2022-09-20
DeleteDirectoryContents: New API to leave top level dir intact

FanDjango:
New function overload (Wiki API will need update) to allow users of `DeleteDirectory` to optionally make this function leave the top level directory intact.

Using this will prevent use of any highly optimised server side RMDIR functionality on the uppermost level, but that's the price.

Usage:
```
		/// <summary>
		/// Deletes the specified directory and all its contents.
		/// </summary>
		/// <param name="path">The full or relative path of the directory to delete</param>
		/// <param name="options">Useful to delete hidden files or dot-files.</param>
		/// <param name="deleteFinalDir">Delete the top level dir too</param>
		public void DeleteDirectory(string path, FtpListOption options, bool deleteFinalDir = true)
```

The default for `deleteFinalDir` is `true`, which is the normal behaviour right now.

### DIFF
--- a/FluentFTP/Client/AsyncClient/DeleteDirectory.cs
+++ b/FluentFTP/Client/AsyncClient/DeleteDirectory.cs
@@ -22,7 +22,7 @@ namespace FluentFTP {
 			path = path.GetFtpPath();
 
 			LogFunction(nameof(DeleteDirectory), new object[] { path });
-			return DeleteDirInternalAsync(path, true, FtpListOption.Recursive, token);
+			return DeleteDirInternalAsync(path, true, FtpListOption.Recursive, true, true, token);
 		}
 
 		/// <summary>
@@ -40,7 +40,25 @@ namespace FluentFTP {
 			path = path.GetFtpPath();
 
 			LogFunction(nameof(DeleteDirectory), new object[] { path, options });
-			return DeleteDirInternalAsync(path, true, options, token);
+			return DeleteDirInternalAsync(path, true, options, true, true, token);
+		}
+
+		/// <summary>
+		/// Asynchronously deletes the contents of the specified directory only.
+		/// </summary>
+		/// <param name="path">The full or relative path of the directorys contents to delete</param>
+		/// <param name="options">Useful to delete hidden files or dot-files.</param>
+		/// <param name="token">The token that can be used to cancel the entire process</param>
+		public Task DeleteDirectoryContents(string path, FtpListOption options, CancellationToken token = default(CancellationToken)) {
+			// verify args
+			if (path.IsBlank()) {
+				throw new ArgumentException("Required parameter is null or blank.", nameof(path));
+			}
+
+			path = path.GetFtpPath();
+
+			LogFunction(nameof(DeleteDirectoryContents), new object[] { path, options });
+			return DeleteDirInternalAsync(path, true, options, false, true, token);
 		}
 
 		/// <summary>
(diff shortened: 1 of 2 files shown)


FluentFTP/Client/SyncClient/DeleteDirectory.cs
```diff
@@ -38,7 +38,25 @@ public partial class FtpClient {
...
+			DeleteDirInternal(path, true, options, true, true);
+		}
+
+		/// <summary>
+		/// Deletes the specified directory and all its contents.
+		/// </summary>
+		/// <param name="path">The full or relative path of the directory to delete</param>
+		/// <param name="options">Useful to delete hidden files or dot-files.</param>
+		/// <param name="deleteFinalDir">Delete the top level dir too</param>
+		public void DeleteDirectory(string path, FtpListOption options, bool deleteFinalDir) {
```
robinrodricks, line 50: Can you pls rename this to `EmptyDirectory`. It will be a new API and be nice, and also you can remove `deleteFinalDir`.

FluentFTP/Client/AsyncClient/DeleteDirectory.cs
```diff
@@ -40,7 +40,26 @@ public partial class AsyncFtpClient {
...
+		}
+
+		/// <summary>
+		/// Asynchronously removes a directory and all its contents.
+		/// </summary>
+		/// <param name="path">The full or relative path of the directory to delete</param>
+		/// <param name="options">Useful to delete hidden files or dot-files.</param>
+		/// <param name="token">The token that can be used to cancel the entire process</param>
+		/// <param name="deleteFinalDir">Delete the top level dir too</param>
+		public Task DeleteDirectory(string path, FtpListOption options, bool deleteFinalDir, CancellationToken token = default(CancellationToken)) {
```
robinrodricks, line 53: Can you pls rename this to `EmptyDirectory`. It will be a new API and be nice, and also you can remove `deleteFinalDir`.

robinrodricks:
Looks fantastic! Thanks again for your wonderful work. Added some comments.

FluentFTP/Client/SyncClient/DeleteDirectory.cs
FanDjango: That's why I requested review. I was unhappy with the order of the parameters, but also did not want to change it and then break existing users implementation. I was hoping for a second opinion on this. If only the creator of `DeleteDirectory` had included this option in the beginning, but that's not how it happens, I know.

But creating a separate API is a super idea, but I will then call into the enhanced `DeleteDirectoryInternal`. I like that and will do it.

FanDjango: Ah, I just learned to make this a draft.

FanDjango: What about `ClearDirectory` or `DeleteDirectoryContents`. This last one would then at least be lexically located close to `DeleteDirectory`?

FanDjango:
I will rebase so we can see the net changes

FanDjango:
Yes, the rebase removed all the reference to the serverhandlers and so on, because now with the new API function, things are a whole lot less complicated (instead of the insane overloading). Thanks @robinrodricks

FanDjango:
If you want this to be a separate .cs, can you do that?

robinrodricks:
I would have merged this PR and changed it myself, but since its a draft I cannot.

Anyway, can you please rename the API to `EmptyDirectory` as `DeleteDirectoryContents` is much too long, and I don't bother much with API discoverability. I prefer short names ("TransferFile" rather than "FXPTransferFile" and so on). I like `ClearDirectory` too, however it is very similar in meaning to `DeleteDirectory` so I can see some users getting confused about it.

-------------------

As you can see I try to use the shortest name possible, ideally in this format `<verb> <term>`

- DirectoryExists
- CreateDirectory
- DeleteDirectory
- EmptyDirectory
- MoveDirectory

robinrodricks:
Oh wait I got it.

FanDjango:
I have just tested it, and it still works. Sorry I forgot about adding the interface definitions. I shall do the Wiki addition, in the next few days and I ask you to review that then...

robinrodricks:
Released!

PR #962, opened 2022-09-22
Add null checks when disposing

### LINKED ISSUE #961: Annoying NullReferenceException in DownloadStream for non-existent file
Hi there,

when debugging an application, which utilizes FluentFtp for ftp communication, it breaks due to some internal NullReferenceException when trying to download some non-existent file:

> System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=FluentFTP
  StackTrace:
   at FluentFTP.AsyncFtpClient.<DownloadFileInternalAsync>d__26.MoveNext()

Is there any chance to avoid this, so we do not have to uncheck NullReferenceException in "Exception Settings"?

Best
Andreas

jnyrup:
I believe this fixes #961, but I haven't tried it out.

When `OpenRead` throws an exception `downStream` isn't assigned to a non-null value.
This causes a `NullReferenceException` when trying to dispose it.
The exception is immediately caught but that has two drawbacks.
* It's much more expensive to throw and catch an exception than null checking it.
* As reported in the issue it breaks the IDE when a debugger is attached and `NullReferenceException` is checked in "Exception Settings".

### DIFF
--- a/FluentFTP/Client/AsyncClient/DownloadFileInternal.cs
+++ b/FluentFTP/Client/AsyncClient/DownloadFileInternal.cs
@@ -35,10 +35,8 @@ namespace FluentFTP {
 				// open the file for reading
 				downStream = await OpenRead(remotePath, Config.DownloadDataType, restartPosition, fileLen, token);
 				// workaround for SOCKS4 and SOCKS4a proxies
-				if (restartPosition == 0)
-				{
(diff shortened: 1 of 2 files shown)


robinrodricks:
Looks like a clean fix, thanks!

Issue #704, opened 2021-04-08
Consider including Code in FtpReply Message


jnyrup:
**FluentFTP Version:** 33.1.3

I hit a case in our logs yesterday where it seems the information included in the `FtpCommandException` could be improved.
From how I read the code, it indicates to me that the the `FtpReply` was unsuccessful, but `InfoMessages` was insufficient to generate a non-empty `ErrorMessage`. 
In that case you could consider a fallback where the `Code` property is included in the `ErrorMessage`.
I don't know enough about the correlation between `InfoMessages` and `Code` to tell if it would be duplicate information, if `Code` was always included in the `ErrorMessage`.

**Logs :**

```
FluentFTP.FtpCommandException
   at async Task FluentFTP.FtpClient.SetDataTypeNoLockAsync(FtpDataType type, CancellationToken token)
   at async Task<List<string>> FluentFTP.FtpClient.GetListingInternalAsync(string listcmd, FtpListOption options, bool retry, CancellationToken token)
   at async Task<FtpListItem[]> FluentFTP.FtpClient.GetListingAsync(string path, FtpListOption options, CancellationToken token)
```

FanDjango:
You are saying there (probably) was a 5xx/4xx or so, but no error messages, although there might have been info messages?

jnyrup:
This partly a question, partly a suggestion.
The case was I had an exception with no information about what was wrong, except that I can deduce that it was _some_ 4xx/5xx.
It would have been useful to have the _exact_ completion code as part of the exception message if it was available.

So I suggest we include `FtpCommandException.CompletionCode` in the exception message.

See e.g. how [here](https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/ArgumentException.cs,77) how `_paramName` is included in the thrown exception message.

FanDjango:
```
public override string [Message](https://source.dot.net/System.Private.CoreLib/R/71cfcd61786aa2d8.html)
        {
            get
            {
                [SetMessageField](https://source.dot.net/System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/ArgumentException.cs.html#88bcc0b08f1c28b8)();
 
[...]

PR #963, opened 2022-09-23
Fix/Enhance #704 Consider including Code in FtpReply Message

FanDjango:
Thanks @jnyrup for your patience with this. Dunno why you didn't PR this yourself - as I had seen empty FtpCommandExceptions myself I decided I would like this too.

Adding a test is of course totally the icing on the cake and I copy/pasted that from you too.

@robinrodricks Depends on you to feel happy with this new message format:

image


jnyrup:
One reason I didn't just open the PR myself was that I like to separate the discussion of_whether_ something should change and _how_ it should change. And issue vs PR is a good way to keep that distinction.

robinrodricks:
Good work guys, thanks a lot!

Issue #823, opened 2022-01-13
NOOP should be counted and waited correctly


Adhara3:
Hi,

I am reusing the same connection to download multiple files. I have NOOP enabled because I sometimes experience timeouts.

Look at this log
```
01-13 14:44:02.060 INFO  144 FtpLogger Response: 227 Entering Passive Mode (172,30,14,136,173,237) 
01-13 14:44:02.062 INFO  144 FtpLogger Status:   Connecting to 172.30.14.136:44525 
01-13 14:44:02.067 INFO  144 FtpLogger Command:  RETR 20211215_16.35.27-WD.log 
01-13 14:44:02.124 INFO  144 FtpLogger Response: 125 Transfer starting BINARY mode for 20211215_16.35.27-WD.log. 
01-13 14:44:03.580 DEBUG 144 FtpLogger Command:  NOOP 
01-13 14:44:03.585 DEBUG 144 FtpLogger Status:   Disposing FtpSocketStream... 
01-13 14:44:03.592 INFO  144 FtpLogger Response: 226 Transfer complete. 
01-13 14:44:03.596 INFO  144 FtpLogger File Verification: PASS 
01-13 14:44:03.598 DEBUG 144 FtpLogger >         DownloadFile("C:\temp\logs\20210428_13.42.14-PO.log", "20210428_13.42.14-PO.log", Overwrite, OnlyChecksum) 
01-13 14:44:03.607 DEBUG 144 FtpLogger >         GetFileSize("20210428_13.42.14-PO.log") 
01-13 14:44:03.613 DEBUG 144 FtpLogger >         OpenRead("20210428_13.42.14-PO.log", Binary, 0, -1) 
01-13 14:44:03.625 DEBUG 144 FtpLogger >         OpenPassiveDataStream(PASV, "RETR 20210428_13.42.14-PO.log", 0) 
01-13 14:44:03.636 INFO  144 FtpLogger Command:  PASV 
01-13 14:44:03.700 INFO  144 FtpLogger Response: 200 NOOP command successful. 
01-13 14:44:03.899 DEBUG 144 FtpLogger >         DownloadFile("C:\temp\logs\20211011_13.47.08-PO.log", "20211011_13.47.08-PO.log", Overwrite, OnlyChecksum) 
01-13 14:44:03.914 DEBUG 144 FtpLogger >         GetFileSize("20211011_13.47.08-PO.log") 
[...]

PR #966, opened 2022-09-26
Fix823 noop

FanDjango:
See issue #823 

NOOPs work fine for non-SSL connections.
NOOPs work fine for SSL connections that are not buffered.

NOOP handling currently fails miserably when SSL connections are buffered.

Since a user who activates NOOPs (by setting `Config.NoopInterval` to some value `> 0`) very probably is very much in need of the "keep-alive" functionality of this feature, turn off buffering when creating the stream, even if `Config.SslBuffering` has been requested (`On` or `Auto`).

Since there are other reasons also, when SSL Buffering is disabled, I have added warning messages telling the user that this has happened. Previously, this was done silently, possibly causing confusion.

Example:
```
Response: 150 Opening BINARY mode data connection for /home/mike/test2 (1073741824 bytes)
Warning:  SSL Buffering force disabled, NOOPs requested
Status:   FTPS Authentication Successful
```

### DIFF
--- a/FluentFTP/Streams/FtpSocketStream.cs
+++ b/FluentFTP/Streams/FtpSocketStream.cs
@@ -1116,16 +1116,32 @@ namespace FluentFTP {
 		/// Conditionally create a SSL BufferStream based on the configuration in FtpClient.SslBuffering.
 		/// </summary>
 		private void CreateBufferStream() {
-			// Fix: SSL BufferStream is automatically disabled when using FTP proxies, and enabled in all other cases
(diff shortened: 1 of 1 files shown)


robinrodricks:
Great work debugging and fixing this. Heartiest thanks!

robinrodricks:
I like the extra comments added into the code to give the user perspective into the issue.

FanDjango:
Thanks. It's not really totally fixed, only mitigated. I am working on it still, step by step.

PR #967, opened 2022-09-26
Show protocol in Auth Success message

FanDjango:
I find this very useful when debugging connects.

Show protocol in the FTPS authentication successful message. Change some upper case to lower case at the same time.

```
Command:  AUTH TLS
Status:   Waiting for response to: AUTH TLS
Response: 234 AUTH TLS successful
Status:   FTPS Authentication Successful, protocol = Tls12
Status:   Time to activate encryption: 0h 0m 0s.  Total Seconds: 0,0030331.
```


robinrodricks:
Awesome, thanks!

PR #971, opened 2022-09-27
Next step: Read stale data

FanDjango:
This is a simple addition to `ReadStaleData` which does not change very much.

Each call to this method is garnished with an additional string parameter so that in verbose logging, you can see from where it has been called.

More importantly, it decides to not use the raw socket for cleaning any stale data if encryption is turned on. 

With encryption turned on, this would be destructive in random (but quite often) occurring cases.

This change is not risky. It needs to be in place before I can put in the next one, for `GetReply`, because there is a dependancy.

### DIFF
--- a/FluentFTP/Client/AsyncClient/Authenticate.cs
+++ b/FluentFTP/Client/AsyncClient/Authenticate.cs
@@ -44,7 +44,7 @@ namespace FluentFTP {
 
 				// fix for #620: some servers send multiple responses that must be read and decoded,
 				// otherwise the connection is aborted and remade and it goes into an infinite loop
-				var staleData = await ReadStaleData(false, true, true, token);
+				var staleData = await ReadStaleDataAsync(false, true, "in authentication", token);
 				if (staleData != null) {
 					var staleReply = new FtpReply();
 					if (DecodeStringToReply(staleData, ref staleReply) && !staleReply.Success) {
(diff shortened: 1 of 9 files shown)


FanDjango:
This is the new `ReadStaleData`:
```
		/// <summary>
		/// Data shouldn't be on the socket, if it is it probably means we've been disconnected.
		/// Read and discard whatever is there and optionally close the connection.
		/// Returns the stale data as text, if any, or null if none was found.
		/// </summary>
		/// <param name="closeStream">close the connection?</param>
		/// <param name="logData">copy stale data information to logs?</param>
		/// <param name="logFrom">for the log information</param>
		protected string ReadStaleData(bool closeStream, bool logData, string logFrom) {
			string staleData = null;

			if (logData) {
				LogWithPrefix(FtpTraceLevel.Verbose, "Checking for stale data: " + logFrom);
			}

			if (m_stream != null) {
				while (m_stream.SocketDataAvailable > 0) {
					if (logData) {
						LogWithPrefix(FtpTraceLevel.Info, "Socket has stale data");
					}
					byte[] buf = new byte[m_stream.SocketDataAvailable];
					if (m_stream.IsEncrypted) {
						m_stream.Read(buf, 0, buf.Length);
					}
					else {
						m_stream.RawSocketRead(buf);
					}
					staleData = Encoding.GetString(buf).TrimEnd('\0', '\r', '\n');
					if (logData) {
						LogWithPrefix(FtpTraceLevel.Verbose, "The stale data was: " + staleData);
					}
				}
			}

			if (string.IsNullOrEmpty(staleData)) {
				closeStream = false;
			}

			if (closeStream) {
				LogWithPrefix(FtpTraceLevel.Info, "Closing stream because of stale data");
				m_stream.Close();
			}

			return staleData;
		}
```

robinrodricks:
Looks ok, I am merging.

robinrodricks:
Keep in mind your changes are highly technical and I don't understand 100% of what it does, but I trust you at this point!

PR #972, opened 2022-09-27
Redundant If

FanDjango:
Maybe I am blind, so someone else also have a lool at this. I just stumbled accross it.

```
					reply = Execute("AUTH TLS");
					if (!reply.Success) {
						Status.ConnectionFTPSFailure = true;
						if (Config.EncryptionMode == FtpEncryptionMode.Explicit) {
							throw new FtpSecurityNotAvailableException("AUTH TLS command failed.");
						}
					}
					else if (reply.Success) {
						m_stream.ActivateEncryption(Host, Config.ClientCertificates.Count > 0 ? Config.ClientCertificates : null, Config.SslProtocols);
					}
```


robinrodricks:
Nice, thanks.

PR #975, opened 2022-09-28
Disable buffering ONLY on the control connection

FanDjango:
Disable of SSL Buffering helped a lot for NOOP handling. But the current implementation disable the buffering **also** for the data connection.

That is not needed. It is enough the disable buffering **only** on the control connection.


robinrodricks:
Very innovative, thanks!

robinrodricks:
Released!

PR #980, opened 2022-09-29
Causer server hickup, release pending responses

FanDjango:
This is a very very dirty quick fix for those servers (.NET 6.0, vsftpd) that can't seem to want to relinquish their pent up NOOP responses **until the next command** comes along, which brings commands and responses out of sync.

It fixes the NOOP problems of vsftpd with .NET 6.0+


robinrodricks:
Looks good, thanks a lot!

PR #981, opened 2022-09-30
Switch to using LISTLEVEL 2, filesize more exact

FanDjango:
Try to use LISTLEVEL 2. Parser understands all LISTLEVELs, so no mind if the z/OS is too old and stays on LISTLEVEL 0. 

This will give a better size resolution on huge files. Remember the '+++++' problem, because the Size column was not wide enough?

214-LISTLEVEL={0|1|2} Specifies the format of the LIST reply.  LISTLEVEL 0
214-                specifies that PDS, PDSE and HFS data sets are displayed
214-                with a DSORG value of PO.  LISTLEVEL 1 specifies that PDS
214-                data sets are displayed with a DSORG value of PO, PDSE
214-                data sets are displayed with a DSORG value of PO-E, and
214-                HFS data sets are displayed with a DSORG value of HFS.
214-                LISTLEVEL 2 specifies LISTLEVEL 1 display options, and
214-                also fewer but wider columns of output to accommodate
214-                larger physical sequential data sets.

### DIFF
--- a/FluentFTP/Helpers/Parsers/IBMzOSParser.cs
+++ b/FluentFTP/Helpers/Parsers/IBMzOSParser.cs
@@ -9,13 +9,15 @@ namespace FluentFTP.Helpers.Parsers {
 		/// </summary>
 		public static bool IsValid(BaseFtpClient client, string[] listing) {
 			// Check validity by using the title line
-			// Dataset       : "Volume Unit    Referred Ext Used Recfm Lrecl BlkSz Dsorg Dsname"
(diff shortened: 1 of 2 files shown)


robinrodricks:
You are the Z/OS expert so I will just accept this!

FanDjango:
I will remove the unused variable "test" (see Codacity) in some future other PR.

robinrodricks:
Don't worry about Codacy. Its crazy!

robinrodricks:
Released!

PR #986, opened 2022-10-02
Move CurrentDataType to Status and remove ForceSetDataType

FanDjango:
`Status` was created to keep status information for the session, so why not move `CurrentDataType` to that substructure.

At the same time, if it inits to `Unknown` we can do away with the `ForceSetDataType`.

Also, a bug fixed: If a user executes a `TYPE` command, `CurrentDataType` goes out of sync. Happily, no one really does this, right?

Sometime in the future, issue a STAT command after connect and parse per server, set the initial value from there


robinrodricks:
Looks good, approved.

Issue #747, opened 2021-08-10
z/OS JES2: Cannot seem to access command responses for RETR and STOR


MihaIvan:
**FTP OS:** Windows

**FTP Server:** IBM FTP (z/OS)

**Computer OS:** Client or server? z/OS for server.

**FluentFTP Version:** 34.0.0

I got an error when trying to upload a file. It ran out of space. However I cannot find a way to see what the server response is which contains the details of the problem. I can see it when I am in debug mode in the Immediate Window in Visual Studio, but I cannot seem to find how to get at that from my program.

Another reason I need this is that I submit jobs to run on the mainframe using FTP. I need to access the response in order to get the job number assigned so I can download the output of the job from the output queue.

**Logs :** Don't know how to do logs. Another issue with finding stuff in the documentation. I can see some things regarding logs, but have no idea how to set it up and produce them. VB.NET example would help.

**Additional note**
With zOSFTP, you can run a job by doing a SITE command telling JES to submit a job when you do a GET for the file that contains the JCL to run the job. What it returns is the output of that job, not the JCL.  When I do the Get, this is what I see in the Immediate Window in Visual Studio:
_Command:  RETR cmpa.pvt.source(jlistcat)
Response: 125-Submitting job cmpa.pvt.source(jlistcat) FIXrecfm 80
Response: 125 Unable to send CMPA.CMPA.PVT.SOURCE(JLISTCAT)
Status:   Disposing FtpSocketStream...
Response: 550 Transfer aborted_
What I get in the FTPReply object is just that last line:
    _Code:550
    ErrorMessage:Transfer aborted
    InfoMessages:
    Message:Transfer aborted
    Success:False
    Type:PermanentNegativeCompletion_

fdrobidoux:
I'd like that too, this is preventing us from accessing the error code from a production log.

fdrobidoux:
I found a way for myself to get results from CLI, which works for my use-case; I just enable this :

```csharp
FtpTrace.LogToConsole = true;
```

A bit verbose on the CLI but it works for us.

MihaIvan:
Mine is not a CLI app, and the purpose will be to interpret the raw responses into something the user will understand who knows nothing about FTP. So I need to get at it programmatically.

robinrodricks:
Firstly, when any command fails it will throw an FtpException and the
details will be in that object .

About logs, example code is there in the examples folder for vb and c#.
Unless you want to access logs without it going to CLI - even that is
[...]

PR #987, opened 2022-10-02
Keep a (short, n=5) list of previous replies. 

FanDjango:
Fixes (as an enhancement) #747 

There is now a `List<FtpReply> LastReplies `with max 5 elements.

The previous `LastReply` still exists - it is fed from `LastReplies[0]`. If you want the next reply previous to that, you need `LastReplies[1]` and so on.

Especially useful to access intermediate messages on download (`RETR`) and upload (`STOR`) commands

### DIFF
--- a/FluentFTP/Client/BaseClient/Properties.cs
+++ b/FluentFTP/Client/BaseClient/Properties.cs
@@ -382,10 +382,18 @@ namespace FluentFTP.Client.BaseClient {
 
 		/// <summary> Gets the last reply received from the server</summary>
 		public FtpReply LastReply {
-			get => m_lastReply;
-			protected set => m_lastReply = value;
+			get {
+				return m_lastReply = m_lastReplies[0];
+			}
 		}
 
+		protected List<FtpReply> m_lastReplies = null;
+
+		/// <summary> Gets the last reply received from the server</summary>
+		public List<FtpReply> LastReplies {
+			get => m_lastReplies;
+			set => m_lastReplies = value;
+		}
 
 		/// <summary>
 		/// Callback format to implement your custom FTP listing line parser.
--- a/FluentFTP/Client/BaseClient/Properties.cs
+++ b/FluentFTP/Client/BaseClient/Properties.cs
@@ -387,7 +387,7 @@ namespace FluentFTP.Client.BaseClient {
 			}
 		}
 
-		protected List<FtpReply> m_lastReplies = null;
+		protected List<FtpReply> m_lastReplies;
 
 		/// <summary> Gets the last reply received from the server</summary>
 		public List<FtpReply> LastReplies {
--- a/FluentFTP/Client/BaseClient/GetReply.cs
+++ b/FluentFTP/Client/BaseClient/GetReply.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using FluentFTP.Helpers;
 using System.Text.RegularExpressions;
 using FluentFTP.Client.Modules;
+using System.Collections.Generic;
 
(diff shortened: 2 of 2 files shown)


FluentFTP/Client/BaseClient/Properties.cs
```diff
@@ -382,10 +382,18 @@ public partial class BaseFtpClient {
 
 		/// <summary> Gets the last reply received from the server</summary>
 		public FtpReply LastReply {
-			get => m_lastReply;
-			protected set => m_lastReply = value;
+			get {
+				return m_lastReply = m_lastReplies[0];
```
jnyrup, line 386: If there haven't been any replies yet, this now throws an IndexOutOfRangeException.

FluentFTP/Client/BaseClient/Properties.cs
```diff
@@ -382,10 +382,18 @@ public partial class BaseFtpClient {
...
+			}
 		}
 
+		protected List<FtpReply> m_lastReplies;
+
+		/// <summary> Gets the last reply received from the server</summary>
```
jnyrup, line 392: Reply -> replies

FluentFTP/Client/BaseClient/Properties.cs
```diff
@@ -382,10 +382,18 @@ public partial class BaseFtpClient {
...
 		/// <summary> Gets the last reply received from the server</summary>
 		public FtpReply LastReply {
-			get => m_lastReply;
-			protected set => m_lastReply = value;
+			get {
+				return m_lastReply = m_lastReplies[0];
+			}
 		}
 
+		protected List<FtpReply> m_lastReplies;
```
jnyrup, line 390: How about using an auto property instead of manually adding a backing field for the property?

FluentFTP/Client/BaseClient/Properties.cs
```diff
@@ -382,10 +382,18 @@ public partial class BaseFtpClient {
...
+
+		/// <summary> Gets the last reply received from the server</summary>
+		public List<FtpReply> LastReplies {
```
jnyrup, line 393: Suggestion, initialize this so the default value is an empty list.

FanDjango: I did, Codacity was unhappy and said, better practice to accept what the compiler initializes as default. (null)

FluentFTP/Client/BaseClient/Properties.cs
FanDjango: I'll check that

FluentFTP/Client/BaseClient/Properties.cs
FanDjango: Not sure I understand

FluentFTP/Client/BaseClient/Properties.cs
FanDjango: Yep, I'll fix that tomorrow.

FluentFTP/Client/BaseClient/Properties.cs
jnyrup: The word "reply" in the summary should be in plural.

FanDjango:
My actual intent was: Do we do this like this, and keep `LastReply` as well, so as not to break code for existing users?

FluentFTP/Client/BaseClient/Properties.cs
FanDjango: Oof. Yes, Thx. Willco.

FanDjango:
Closed in order to re-open a cleaner version.

PR #988, opened 2022-10-03
Keep a (short, n=5) list of previous replies.

FanDjango:
Fixes (as an enhancement) https://github.com/robinrodricks/FluentFTP/issues/747

There is now a List<FtpReply> LastReplies with max 5 elements.

The previous LastReply still exists - it is fed from LastReplies[0]. If you want the next reply previous to that, you need LastReplies[1] and so on.

Especially useful to access intermediate messages on download (RETR) and upload (STOR) commands


robinrodricks:
Great!!

Issue #994, opened 2022-10-04
FtpCommandException: Code: 550 Message:  Unable to find file


noxe:
**FTP OS:** OS/400

**FTP Server:**  IBMOS400FTP

**Computer OS:** Windows 11

**FluentFTP Version:** 40 / 41

**Framework:** .NET 6

Hi - it seems there is an problem when an FTP Server contains files which contains whitespaces. See screenshot:

![image](https://user-images.githubusercontent.com/158478/193844106-ec62ce04-2af5-41e9-9f4e-d8ef86504864.png)

Donwloading the files results in an exception, while the real problem is that the ListItem does not contain the correct name, see screen:

![image](https://user-images.githubusercontent.com/158478/193844464-0f67dff7-4868-4ec3-8658-a6666b477655.png)

var listing = client.GetListing();
        var listItem = listing.FirstOrDefault(l => l.Name.StartsWith("B0001417017"));   // but real name is B0001417017 J
        if (listItem != null)
        {
            client.DownloadBytes(out byte[] data, listItem.Name);

FtpCommandException: Code: 550 Message:  Unable to find file /kundenftp/ku0021250/download/B0001417017

**Logs :**

# AutoConnect()

# AutoDetect(True, False)

# Connect()
Status:   Connecting to ***:21
Response: 220-QTCP at holter.ad.holter.at.
Response: 220 Connection will close if idle more than 5 minutes.
Status:   Detected FTP server: IBMOS400FTP
Command:  AUTH TLS
Response: 431 Requested security mechanism not available at this time.
Command:  USER ***
Response: 331 Enter password.
Command:  PASS ***
Response: 230 KU0021250 logged on.
Command:  FEAT
Response: 211-Feature listing follows:
Response: AUTH TLS
Response: AUTH GSS
Response: CCC
Response: PBSZ
Response: PROT
Response: ADAT
Response: ENC
Response: MIC
Response: CONF
Response: 211 End of feature listing.
Status:   Text encoding: System.Text.UTF8Encoding+UTF8EncodingSealed
Command:  OPTS UTF8 ON
Response: 501 OPTS unsuccessful; specified subcommand not recognized.
Command:  SYST
Response: 215  OS/400 is the remote operating system. The TCP/IP version is "V7R4M0".
Status:   Listing parser set to: IBMOS400

# SetWorkingDirectory("/kundenftp/ku0021250/download")
Command:  CWD /kundenftp/ku0021250/download
[...]

PR #995, opened 2022-10-04
IBM OS400: Support blanks in filename, add unit test for that

FanDjango:
IBM OS400 filenames can contain blanks.

Motivated by #994


robinrodricks:
Great work!

Issue #996, opened 2022-10-05
UploadFile: 425 Error while transfering data: ECONNABORTED - Connection aborted


FanDjango:
**FTP OS:** Any

**FTP Server:** FileZille 1.5.x

**Computer OS:** WIndows 10

**FluentFTP Version:** V 41.1.0

**Framework:** .NET 6 AND NETFramework 4.8

```
			FTP_Sess.AutoConnect();

			FTP_Sess.UploadDirectory("D:\\temp\\issue311", "temp/issue311/upload", FtpFolderSyncMode.Update, FtpRemoteExists.Overwrite, FtpVerify.None, null, progress);

			FTP_Sess.Disconnect();
```

Files are uploaded, turn up fine on the server, but the transfer is reported as failed.

**Logs :**
```
[Error] GnuTLS error -110 in gnutls_record_recv: The TLS connection was non-properly terminated.
[Status] Client did not properly shut down TLS connection
[Trace] tls_layer_impl::shutdown()
[Trace] data_socket_->shutdown() = 126
[Response] 425 Error while transfering data: ECONNABORTED - Connection aborted
```

```
>         AutoConnect()
>         AutoDetect(True, False)
>         Connect()
Status:   FluentFTP 41.0.0.0
Status:   Connecting to ***:21
Status:   Waiting for a response
Response: 220-FileZilla Server 1.5.1
Response: 220-Please visit https://filezilla-project.org/
Response: 220 Welcome to AnSyNova FileZilla Server
Status:   Detected FTP server: FileZilla
Command:  AUTH TLS
Status:   Waiting for response to: AUTH TLS
Response: 234 Using authentication type TLS.
Warning:  SSL Buffering force disabled, is .NET 5.0 and later
Status:   FTPS authentication successful, protocol = Tls12
Status:   Time to activate encryption: 0h 0m 0s.  Total Seconds: 0,1204242.
Command:  USER ***
Status:   Waiting for response to: USER ***
Response: 331 Please, specify the password.
Command:  PASS ***
Status:   Waiting for response to: PASS ***
Response: 230 Login successful.
Command:  PBSZ 0
Status:   Waiting for response to: PBSZ 0
Response: 200 PBSZ=0
Command:  PROT P
Status:   Waiting for response to: PROT P
Response: 200 Protection level set to P
Command:  FEAT
Status:   Waiting for response to: FEAT
Response: 211-Features:
Response: MDTM
Response: REST STREAM
Response: SIZE
[...]

PR #997, opened 2022-10-08
FluentFTP wrapper for SslStream

FanDjango:
Send your own close notify. See #996.

This wraps SslStream, takes Dispose and goes into marshalled unsafe code to produce an encrypted message to send on the socket that will be decrypted on the server that then reads correctly as an alert to close notify.

Note: When (someday) TLS 1.3 is taken a step further and this might have been fixed by MS, one might need to programmatically disable the sending of the close notify unless TLS 1.2 is being used.

### DIFF
--- a/FluentFTP/Streams/FluentSslStream.cs
+++ b/FluentFTP/Streams/FluentSslStream.cs
@@ -1,0 +1,223 @@
+﻿using System.Collections.Generic;
+using System;
+using System.IO;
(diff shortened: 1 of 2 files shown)


FanDjango:
Although this already works fine, I will add to this, by making the use of this configurable as an option in the client Config structure.

robinrodricks:
Massive development! Thanks a lot!

PR #999, opened 2022-10-09
Ongoing changes for SSL Shutdown

FanDjango:
1. Go back to a better naming of the SSL wrapper (`FtpSslStrem` etc.)
2. Disregard the setting of `Config.DisconnectWithShutdown`. The parameter remains in `Config` but is ignored. An API change is deferred until some more have accumulated. Wiki will reflect this
3. Remove .NET 5+ code in FtpSslStream, but keep it as comments in case it is ever needed.


FanDjango:
Codacy doesn't like commented out code (see #3 above) and one empty statement ";" which is entirely on purpose.

PR #1002, opened 2022-10-10
Change comment

FanDjango:
Although `DisconnectWithShutdown` is no longer used, keep until V42 so that users builds continue to compile.


robinrodricks:
If its unused, delete it. I will publish v42 as the next release.

FanDjango:
Roger wilco. Still busy with docker and our tests. 

 

mit freundlichen Grüßen,

Michael Stiemke ***@***.***)

 

AnSyNova GmbH

Ringstr. 13

85405 Nandlstadt

+49 8756 969 49 79

PR #1005, opened 2022-10-11
Fix/update vsftpd docker, attempt TLS for all

FanDjango:
**`IntegrationeTestSuite.cs`:**

Turn on SSL/TLS and accept snakeoil certificate on all connections the tests are using.

**dockers/vsftpd**

Replace the old centos docker by **our own** one, based on debian 11 (bullseye). It has the newest vsftpd (3.0.3) and supports TLS.

It is a bit larger, but duh...


robinrodricks:
Great work.

PR #1007, opened 2022-10-11
proftpd ssl

FanDjango:
Here is the integration test server proftpd.

Supports passive and TLS, just like the vsftpd.

These servers are totally standardized, same base image debian, same port ranges, same user and password, same everything (nearly). No more pandering to the different environment variable concepts of the image provider out on the web. Updates easy to get simply by removing and rebuilding.

Switched to an OpenSsl self signed certificate with long duration for the servers.

Fixed some unclean stuff in vsftpd, also.


FanDjango:
Closed because I messed up git on the branch. Easier to close and to re-open a new PR

PR #1009, opened 2022-10-12
Ability to run test server containers as FTP or FTPS servers

FanDjango:
Those servers capable of this (currently vsftpd and proftpd) can be run from a single docker image, a single docker file etc. as either an FTP or an FTPS server. This is controlled by an environment variable (`$USE_SSL`).


robinrodricks:
Brillante Arbeit! Mir fehlen die Worte!

PR #1010, opened 2022-10-12
Add bftpd test server

FanDjango:
Lets add another test server.

This one was interesting: Due to a bug it has (it is not an impressive server, to be honest), it uses up its passive port range and then fails. This is why in this docker container the passive port range is **a little bit larger**, otherwise the tests fail.


robinrodricks:
Absolutely amazing work! I love adding new servers. Thanks a ton!!

PR #1011, opened 2022-10-13
Drastically reduce some test server image sizes

FanDjango:
...by using a two stage build debian bullseye-slim

bftpd **450MB** --> **150MB**, others save 150MB or so.


robinrodricks:
Wow, looks great!

PR #1012, opened 2022-10-18
Add glftpd integration test server

FanDjango:
This one was a little more involved. It is slightly against the docker philosophy to do it this way. 

> absolutely need to run the convoluted install script.
> absolutely need to run xinetd in the container
> absolutely need to connect to the running server to add fluentuser/fluentpass etc. before using it

It is rather big ( = 434MB ), as time permits I will try to get a two-stage build to work. Up to now, I was not able to do that.


robinrodricks:
Wow, impressive work! I always welcome new test servers. Thanks a lot!

PR #1022, opened 2022-10-23
Docker pureftpd as first party server, with and without SSL

FanDjango:
Put pure-ftp on the same base as the previous FluentFTP first party integration test servers. 

No need for source compile, much smaller Dockerfile and less failure potential, just allow capability switch in the Docker kernel on startup.

```
			builder = builder.WithCreateContainerParametersModifier(x => {
				x.HostConfig.CapAdd = new List<string> {
					"ALL"
				};
```

Allow outside config of SSL or no-SSL via environment variable.

Acceptably fast build time.


robinrodricks:
Awesome!!

PR #1028, opened 2022-10-26
Better handling of TLS Auth failures

FanDjango:
Apart from bona fide AuthenticationExceptions, failed TLS session resumptions turn up as:

* AggregateException in .NET 5.0+
* IOException with innerexception in .NET Framework

So, these will be transformed into AuthenticationExceptions.

But: AuthenticationExceptions weren't being handled, just thrown, therefore hiding futher information perhaps available as stale data on the control connection before closing it.

So, it became necessary to identify all users of data streams at the highest level, to where the AuthenticationException should percolate upwards and handle it.

Handling it consists of throwing the exception, which of course could have been done far down in the nested code, **but** up in the callers code, one can use the `ReadStaleData( ... )` function to pick up any leftover information the server deigned to give us before closing down.

A more convoluted solution would be to check for data on the socket and to call `GetReply( ... )`, this would then result in a command exception. I will think about the merits of going that far to handle these kind of failures.

```
# GetListing(null, Auto)
Command:  PWD
Status:   Waiting for response to: PWD
Response: 257 "/" is current directory.
Command:  TYPE I
Status:   Waiting for response to: TYPE I
Response: 200 Type set to I

# OpenPassiveDataStream(AutoPassive, "MLSD /", 0)
Command:  EPSV
Status:   Waiting for response to: EPSV
Response: 229 Entering Extended Passive Mode (|||21133|)
Status:   Connecting to ***:21133
Command:  MLSD /
Status:   Waiting for response to: MLSD /
Response: 150 Opening data channel for directory listing of "/"
Warning:  SSL Buffering force disabled, is .NET 5.0 and later
Error:    FTPS Authentication Failed:
Error:     Received an unexpected EOF or 0 bytes from the transport stream.
Status:   Socket has stale data - after GetListing
Status:   The stale data was: 
Stale:    450 TLS session of data connection has not resumed or the session does not match the control connection
Status:   Closing stream because of stale data
Exception thrown: 'System.Security.Authentication.AuthenticationException' in FluentFTP.dll
An unhandled exception of type 'System.Security.Authentication.AuthenticationException' occurred in FluentFTP.dll
 Received an unexpected EOF or 0 bytes from the transport stream.
```


robinrodricks:
Looks good, keep going! Merge when you feel ready..

FanDjango:
Ok, this looks even more promising:
```
			catch (AuthenticationException ex) {
				FtpReply reply = await GetReplyAsyncInternal(token, "GetListing", false, -1); // no exhaustNoop, but non-blocking
				if (!reply.Success) {
					throw new FtpCommandException(reply);
				}
				throw;
			}
```
On an `AuthorizationException` call (**a new (3rd) mode**) of `GetReply( ... )`. It does not `exhaustNoops`, it does a normal `GetReply( ... )`, **but** in a non-blocking mode: If there's nothing there right away, come back empty, otherwise: process like a normal reply, success/failure and the whole set of stuff.

And then, if there is a bona fide failure message, throw a `FtpCommandException`, otherwise just throw the original `AuthenticationException`.

The 3rd mode of GetReply? Small changes: Add a timeOut Parameter, and if it is set to -1, go non-blocking.

PR #1036, opened 2022-11-08
Async Connect Timeouts

### LINKED ISSUE #869: ConnectTimeout not working for ConnectAsync
**FTP OS:** Windows

**Computer OS:** Windows 10 21H2

**FluentFTP Version:** 37.0.0

Target framework: .NET Framework 4.7.2

When calling ConnectAsync() and the host doesn't exist it takes 21 seconds before there is a timeout.
Below is the code I use.
 ```
using (var conn = new FtpClient())
{
    conn.Host = "192.168.0.14";
    conn.Credentials = new NetworkCredential("ftptest", "ftptest");
    conn.ConnectTimeout = 5000;
    await conn.ConnectAsync();
}
```
When calling Connect() the timeout works fine.

**Logs :**

```

# ConnectAsync()
# ConnectAsync()
Status:   Connecting to 192.168.0.14:21
Status:   Connecting to 192.168.0.14:21
Exception thrown: 'System.Net.Sockets.SocketException' in System.dll
Exception thrown: 'System.Net.Sockets.SocketException' in mscorlib.dll
Exception thrown: 'System.Net.Sockets.SocketException' in mscorlib.dll
Exception thrown: 'System.Net.Sockets.SocketException' in mscorlib.dll
Exception thrown: 'System.Net.Sockets.SocketException' in mscorlib.dll
Exception thrown: 'System.Net.Sockets.SocketException' in mscorlib.dll


# Dispose()
# Dispose()
Status:   Disposing FtpClient object...
Status:   Disposing FtpClient object...

```

### LINKED ISSUE #1035: Connect timeout
**FTP OS:** N/A

**FTP Server:** N/A

**Computer OS:** ? Windows 10

**FluentFTP Version:** ? 42.0.1

**Framework:** .NetFramework 4.8 & .Net 5.0

Different behavior of Connect method (of AsyncFtpClient class) between .NetFramework 4.8 and .Net 5.0:
In .Net 5.0 when trying to connect to a server that doesn't exist, I get a System.TimeoutException and the timeout corresponds to the Config.ConnectTimeout I set.
In .NetFramework 4.8 (using the same code as in .Net 5.0) I get a System.Net.Sockets.SocketException and the timeout is 15 seconds no matter what I put inConfig.ConnectTimeout.

FanDjango:
Fixes #1035.

Also, it brings in some other missing stuff in the async section of Connect.

In the sync section, I tried to make the behaviour for a timeout value of 0 behave the same way as in the async section, which means take the internal socket timeout (about 21 seconds). Otherwise a timeout of 0 would return immediately w/o a connection.

Also, the code to fix #869 was somewhat flaky. Generally, I would refrain from using error texts (in possibly hundreds of different language variants) and try to use socket error codes. Testing for the error text **and** the error code is also quite redundant.

### DIFF
--- a/FluentFTP/Streams/FtpSocketStream.cs
+++ b/FluentFTP/Streams/FtpSocketStream.cs
@@ -256,10 +256,22 @@ namespace FluentFTP {
 		/// Gets or sets the length of time milliseconds to wait
 		/// for a connection succeed before giving up. The default
 		/// is 30000 (30 seconds).
+		/// Note: This can be used to force a lower timeout than the
+		/// system default, but if you want to increase the timeout
+		/// (i.e. to more than 21 seconds as is usually set in MS Windows)
+		/// you have to change the TCP settings of the underlying
+		/// operating system. To change this, you can use the following
+		/// PowerShell command (running the PowerShell prompt as admin):
+		/// "Set-NetTCPSetting -SettingName InternetCustom -MaxSynRetransmissions 4"
+		/// The default value is 2, which yields an effective TCP connection timeout
+		/// of about 21 seconds. A value of 3 increases the timeout to about 40 seconds,
+		/// and 4 to about 90 seconds.
 		/// </summary>
 		public int ConnectTimeout {
 			get => m_connectTimeout;
-			set => m_connectTimeout = value;
+			set {
+				m_connectTimeout = value > 0 ? value : 30000;
+			}
 		}
 
 		/// <summary>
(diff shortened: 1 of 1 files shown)


robinrodricks:
Ok but please keep in mind the required behavior - if there are multiple `addresses[i]`, we are supposed to try each one before breaking with a timeout exception.

FanDjango:
Good catch - I will work on this some more.

robinrodricks:
A suggestion - what if we moved the main connection part into a new subroutine, and then called that from within the "addresses" loop. That way the addresses loop can catch this error and either re-throw it if its the last address, or move to the next address if there are more. So that way it might reduce complication?

FanDjango:
I was thinking in the same direction, only we would need a separate subroutine for sync and for async. But I think it would make the code a lot clearer. Working on it, I am on the road today until the WE and using mobile internet, please stand by.

FanDjango:
SYNC, original code - does not iterate over IPs in the NETSTANDARD case.

 

#if NETSTANDARD

 

                        var args = new SocketAsyncEventArgs {

                             RemoteEndPoint = new IPEndPoint(addresses[i], port)

                        };

                        var connectEvent = new ManualResetEvent(false);

                        args.Completed += (s, e) => { connectEvent.Set(); };

 

                        if (m_socket.ConnectAsync(args)) {

                             if (!connectEvent.WaitOne(connectTimeout)) {

                                   Close();

                             }

                        }

 

                        if (args.SocketError != SocketError.Success) {

                             throw new SocketException((int)args.SocketError);

                        }

 

                        // only try the first address

                        break;

#else

                        ar = m_socket.BeginConnect(addresses[i], port, null, null);

                        bool success = ar.AsyncWaitHandle.WaitOne(connectTimeout, true);

                        if (!success) {

                             Close();

 

                             // check to see if we're out of addresses, and throw a TimeoutException

                             if (i + 1 == addresses.Length) {

                             }

                        }

                        else {

                             ar.AsyncWaitHandle.Close();

                             m_socket.EndConnect(ar);

 

                             // we got a connection, break out

                             // of the loop.

                             break;

                        }

 

 

ASYNC was even worse.

 

So you can see that a slight redesign is in order.

 

mit freundlichen Grüßen,

Michael Stiemke ***@***.***)

 

AnSyNova GmbH

Ringstr. 13

85405 Nandlstadt

+49 8756 969 49 79

PR #1037, opened 2022-11-12
Complete overhaul of connect code

### LINKED ISSUE #869: ConnectTimeout not working for ConnectAsync

### LINKED ISSUE #1035: Connect timeout

FanDjango:
Fixes #1035 

Also, it brings in some other missing stuff in the async section of Connect.

In the sync section, I tried to make the behaviour for a timeout value of 0 behave the same way as in the async section, which means take the internal socket timeout (about 21 seconds). Otherwise a timeout of 0 would return immediately w/o a connection.

Also, the code to fix #869 was somewhat flaky. Generally, I would refrain from using error texts (in possibly hundreds of different language variants) and try to use socket error codes. Testing for the error text and the error code is also quite redundant.

Also implemented @robinrodricks suggestion:

> A suggestion - what if we moved the main connection part into a new subroutine, and then called that from within the "addresses" loop. That way the addresses loop can catch this error and either re-throw it if its the last address, or move to the next address if there are more. So that way it might reduce complication?

I added connect timeout documentation to the Wiki.

### DIFF
--- a/FluentFTP/Model/FtpConfig.cs
+++ b/FluentFTP/Model/FtpConfig.cs
@@ -172,9 +172,11 @@ namespace FluentFTP {
 
 		/// <summary>
 		/// Gets or sets the length of time in milliseconds to wait for a connection 
-		/// attempt to succeed before giving up. Default is 15000 (15 seconds).
+		/// attempt to succeed before giving up. Default is 0 (Use OS default timeout)
+		/// See: https://github.com/robinrodricks/FluentFTP/wiki/FTP-Connection#connection-timeout-settings
+		/// and: https://github.com/robinrodricks/FluentFTP/wiki/FTP-Connection#faq_timeoutwindows
 		/// </summary>
-		public int ConnectTimeout { get; set; } = 15000;
+		public int ConnectTimeout { get; set; } = 0;
 
 		/// <summary>
 		/// Gets or sets the length of time wait in milliseconds for data to be
--- a/FluentFTP/Streams/FtpSocketStream.cs
+++ b/FluentFTP/Streams/FtpSocketStream.cs
@@ -832,7 +832,6 @@ namespace FluentFTP {
 				IPAddress ipad = addresses[i];
 
 				string logIp = Client.Config.LogHost ? ipad.ToString() : "***";
-				string addrFamily = ipad.AddressFamily.ToString();
 
 				if (!IsIpVersionAllowed(ipad, ipVersions, out string logFamily)) {
 					((IInternalFtpClient)Client).LogStatus(FtpTraceLevel.Verbose, "Skipped " + logFamily + " address: " + logIp);
@@ -910,7 +909,7 @@ namespace FluentFTP {
 			return args.SocketError == SocketError.Success;
 #else
 			IAsyncResult iar = m_socket.BeginConnect(ipad, port, null, null);
-			bool success = iar.AsyncWaitHandle.WaitOne(ctmo, true);
+			_ = iar.AsyncWaitHandle.WaitOne(ctmo, true);
 			if (!m_socket.Connected) {
 				Close();
 				throw new TimeoutException("Timed out trying to connect!");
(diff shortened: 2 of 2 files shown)


robinrodricks:
Looks awesome, but I don't like using `int.MaxValue` as it can cause issues if people are storing this in a file or whatever. Can we keep `0` as the default value meaning "no timeout" and then let the user fill it if he wishes?

robinrodricks:
Otherwise it looks great, thanks!

FanDjango:
I am very thankful that we are still discussing and looking at the details. 

 

Actually, I overlooked the config part of this logic. Need to change the docs to reflect the following ( = 15 seconds ).

 

Look in Config,

 

            /// <summary>

            /// Gets or sets the length of time in milliseconds to wait for a connection 

            /// attempt to succeed before giving up. Default is 15000 (15 seconds).

            /// </summary>

            public int ConnectTimeout { get; set; } = 15000;

 

the default is still 15 seconds, which will be honored.

 

When a new socket stream is created:

 

The m_stream connect timeout is defaulted to maxint.

 

            private int m_connectTimeout = int.MaxValue;

 

            /// <summary>

            /// Gets or sets the length of time milliseconds to wait

            /// for a connection succeed before giving up. The default

            /// is 0 = disable, use system default timeout.

            /// </summary>

            public int ConnectTimeout {

                  get => m_connectTimeout;

                  set {

                        m_connectTimeout = value > 0 ? value : int.MaxValue;

                  }

            }

 

Then the config connect timeout is plugged into the m_stream connect timeout, for example, in connect():

 

                  m_hashAlgorithms = FtpHashAlgorithm.NONE;

                  m_stream.ConnectTimeout = Config.ConnectTimeout;

                  m_stream.SocketPollInterval = Config.SocketPollInterval;

                  await ConnectAsync(m_stream, token);

 

If someone wants to disable, he can set to 0, but internally in the m_stream that translates to maxint. This then means that our timeout won't fire. The internal systems timeout will win.

 

The m_stream Connect_Timeout is not the one that the user sees.

 

mit freundlichen Grüßen,

Michael Stiemke ***@***.***)

 

AnSyNova GmbH

Ringstr. 13

85405 Nandlstadt

+49 8756 969 49 79

FanDjango:
So in Config, I have changed the default to 0, which then means the user will get the OS default timeout.

PR #1047, opened 2022-11-16
DNS caching

FanDjango:
This solves two problems:

1. A large amount of DNS request can be issued on connect in some situations. This can cause a rejection by the DNS server as protection against a DOS attack.

It can happen when you issue a lot of connects using the same client, 

or 

when you issue commands that cause a lot of data connections using PASV against local non-routable addresses.

2. All connections to the server should use the same firstly issued DNS query answer, even if the server is multi-interfaced and different DNS replies are issued.

The cache lives as long as the client instance.


FanDjango:
Closed because of Git out of sync

PR #1048, opened 2022-11-16
DNS Caching

FanDjango:
This solves two problems:

1. A large amount of DNS request can be issued on connect in some situations. This can cause a rejection by the DNS server as protection against a DOS attack.

It can happen when you issue a lot of connects using the same client,

or

when you issue commands that cause a lot of data connections using PASV against local non-routable addresses.

2. All connections to the server should use the same firstly issued DNS query answer, even if the server is multi-interfaced and different DNS replies are issued.

The cache lives as long as the client instance.

### DIFF
--- a/FluentFTP/Streams/FtpSocketStream.cs
+++ b/FluentFTP/Streams/FtpSocketStream.cs
@@ -836,8 +836,7 @@ namespace FluentFTP {
 		/// <param name="host">The host to query</param>
 		private void SetCachedHostAddresses(string host, IPAddress ipad) {
 			if (Client.Status.CachedHostIpads.ContainsKey(host)) {
-				Client.Status.CachedHostIpads.Remove(host);
-				Client.Status.CachedHostIpads.Add(host, new IPAddress[1] { ipad });
+				Client.Status.CachedHostIpads[host] = new IPAddress[1] { ipad };
 			}
 			else {
 				Client.Status.CachedHostIpads.Add(host, new IPAddress[1] { ipad });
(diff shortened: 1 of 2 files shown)


FanDjango:
I suppose one would need to store the actual IP address itself in addition to the array of replied addressed, to ensure the subsequent queries produce the same destination IP addressed. I am thinking about doing this.

FluentFTP/Streams/FtpSocketStream.cs
```diff
@@ -805,7 +805,43 @@ public class FtpSocketStream : Stream, IDisposable {
...
+
+			return ipads;
+		}
+
+		/// <summary>
+		/// Set the connected IP Address associated with this host
+		/// </summary>
+		/// <param name="host">The host to query</param>
+		private void SetCachedHostAddresses(string host, IPAddress ipad) {
+			if (Client.Status.CachedHostIpads.ContainsKey(host)) {
```
robinrodricks, line 838: perfect pattern!

FluentFTP/Streams/FtpSocketStream.cs
```diff
@@ -805,7 +805,43 @@ public class FtpSocketStream : Stream, IDisposable {
...
+		}
+
+		/// <summary>
+		/// Get the IP Address(es) associated with this host
+		/// </summary>
+		/// <param name="host">The host to query</param>
+		private IPAddress[] GetCachedHostAddresses(string host) {
+			IPAddress[] ipads;
+
+			if (Client.Status.CachedHostIpads.ContainsKey(host)) {
```
robinrodricks, line 818: perfect!

FluentFTP/Streams/FtpSocketStream.cs
robinrodricks, line 838: If it exists I think you can just `dict[key] = value`, you don't need to remove/add imo.

FanDjango: Yes. Thanks

PR #1050, opened 2022-11-17
Better log message for stream dispose

FanDjango:
To help debugging, it is useful to know **which** stream is suddenly being disposed.

The needed flag for `IsControlConnection`: First I set that in `connect()`, but later on I realised it would be much cleaner to put that into the stream constructor.

A very nice side effect of adding this flag was to take back the parameter passing of the `isControlConnection` flag in the `CreateBufferStream` process.


robinrodricks:
Ahh brilliant work! Better logs are always good.

robinrodricks:
This will help a LOT i'm sure esp for debugging those pesky uploads/downloads/GetListings..

PR #1051, opened 2022-11-17
Make resume logic the same for upload/download sync/async

FanDjango:
ResumeDownload Sync and Async look good, but they are missing the try catch that can be seen in ResumeUpload Async,

ResumeUpload Sync is missing, although there is a ResumeUpload Async.

1. Make a nice try catch aggregate exception (using ResumeDownloadAsync as a model) in the other three cases
2. Code the missing ResumeUpload for Sync using the others as a model.
3. Add a log message so that the resume operation is actually noticed by users


robinrodricks:
Fantastic!

robinrodricks:
Now we need integration tests to validate our resume logic...

Issue #1054, opened 2022-11-19
SocketTaskExtensions Class is not supported on .Net Framework 4.6.2


sandra-dcs:
FTP Server OS: Windows 

FTP Server Type: Windows Server IIS 

Client Computer OS: Windows 

FluentFTP Version: 4.2 

Framework: .NET  4.6.2

.Net 4.6.2 does not support SocketTaskExtensions so it fails to compile.

The problem is in Streams\FtpSocketStream.cs

The following method fixes.

```
	/// <summary>
		/// Connect to the specified host
		/// Detects timeout and throws that explicitly
		/// </summary>
		/// <param name="ipad">The ip address to connect to</param>
		/// <param name="port">The port to connect to</param>
		private async Task<bool> ConnectAsyncHelper(IPAddress ipad, int port, CancellationToken token) {

			int ctmo = this.ConnectTimeout;

#if NETSTANDARD
			try {
				using (var timeoutSrc = CancellationTokenSource.CreateLinkedTokenSource(token)) {
					timeoutSrc.CancelAfter(ctmo);
					await EnableCancellation(m_socket.ConnectAsync(ipad, port), timeoutSrc.Token, () => DisposeSocket());
				}
			}
			catch (SocketException ex) {
				if (ex.SocketErrorCode == SocketError.OperationAborted ||
					ex.SocketErrorCode == SocketError.TimedOut) {
					throw new TimeoutException("Timed out trying to connect!");
				}
				throw;
			}
#else
			 try {
#if v472
				 
				    using (var timeoutSrc = CancellationTokenSource.CreateLinkedTokenSource(token)) {
					timeoutSrc.CancelAfter(ctmo);

                    await EnableCancellation(m_socket.ConnectAsync(ipad, port), timeoutSrc.Token, () => DisposeSocket());
					}
#else
				using (var timeoutSrc = CancellationTokenSource.CreateLinkedTokenSource(token)) {
					timeoutSrc.CancelAfter(ctmo);
					var connectResult = m_socket.BeginConnect(ipad, port, null, null);
					await EnableCancellation(Task.Factory.FromAsync(connectResult, m_socket.EndConnect), timeoutSrc.Token, () => DisposeSocket());
				}
#endif
			}
			catch (SocketException ex) {
				if (ex.SocketErrorCode == SocketError.OperationAborted ||
					ex.SocketErrorCode == SocketError.TimedOut) {
					throw new TimeoutException("Timed out trying to connect!");
				}
[...]

PR #1056, opened 2022-11-21
Cater to #1054

FanDjango:
In case a source compile targets NET Framework 4.6.2, as reported in #1054, there is a compile error.

It used to be in the old code, I had removed it when working on the connect timeouts.

Of course, just using the nuget package will work fine, but those of us who need to use the source might like this to be fixed.

Merging this is optional IMO, but perhaps for the sake of completeness one should include this in the code.

### DIFF
--- a/FluentFTP/Streams/FtpSocketStream.cs
+++ b/FluentFTP/Streams/FtpSocketStream.cs
@@ -1087,10 +1087,19 @@ namespace FluentFTP {
 			}
 #else
 			try {
+// fix #1054
+#if v472
+			    using (var timeoutSrc = CancellationTokenSource.CreateLinkedTokenSource(token)) {
+					timeoutSrc.CancelAfter(ctmo);
+			        await EnableCancellation(m_socket.ConnectAsync(ipad, port), timeoutSrc.Token, () => DisposeSocket());
+				}
+#else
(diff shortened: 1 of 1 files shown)


FluentFTP/Streams/FtpSocketStream.cs
```diff
@@ -1087,10 +1087,19 @@ public class FtpSocketStream : Stream, IDisposable {
 			}
 #else
 			try {
+// fix #1054
+#if v472
+			    using (var timeoutSrc = CancellationTokenSource.CreateLinkedTokenSource(token)) {
```
robinrodricks, line 1092: Since this code is the same, you can wrap only changed lines between an `#if/#endif` block. It does not have to be a semantically complete block, you can even wrap incomplete statements.

Like (pls double check)
```
	timeoutSrc.CancelAfter(ctmo);
#if v472
#else
	var connectResult = m_socket.BeginConnect(ipad, port, null, null);
	await EnableCancellation(Task.Factory.FromAsync(connectResult, m_socket.EndConnect), timeoutSrc.Token, () => DisposeSocket());
#endif
}
```

FanDjango: Ok, I'll modify. I hadn't "perfected" it yet, because I first wanted to know if we really want to do this, as making the number of IFs less is a bit of a goal, right?

PR #1078, opened 2022-11-28
Fix formatting

FanDjango:
Very strange, but VS2022 is falsely formatting the second commit on each pull request, I don't understand why. .editorconfig was my first suspicion, but it seems I need to stop and restart VS2022 after each single commit.


robinrodricks:
Released!

https://www.nuget.org/packages/FluentFTP/42.1.0

PR #1090, opened 2022-12-09
Boolean! Compare to false and not to null.

FanDjango:
`client.ValidateCertificateHandlerExists == null` is a typo. Boolean == null is always false.

Merge with outer if statement and correct the typo:

`profile.Encryption != FtpEncryptionMode.None && client.ValidateCertificateHandlerExists == false`

Much better.

### DIFF
--- a/FluentFTP/Client/Modules/ConnectModule.cs
+++ b/FluentFTP/Client/Modules/ConnectModule.cs
@@ -476,17 +476,15 @@ namespace FluentFTP.Client.Modules {
 		/// Create a default ValidateCertificate handler that accepts valid certificates.
 		/// </summary>
 		public static void SetDefaultCertificateValidation(BaseFtpClient client, FtpProfile profile) {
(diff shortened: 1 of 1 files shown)


robinrodricks:
Oh oh, good find. Surprising the compiler did not catch this.

FanDjango:
Hi Robin,

 

Compiler did notice. I was checking over all the millions of warnings and considering cleaning some of them up. So much XML stuff, it hides so many others. 

 

It was a warning though. I suppose I think it actually should be an error.

 

mit freundlichen Grüßen,

Michael Stiemke ***@***.***)

 

AnSyNova GmbH

Ringstr. 13

85405 Nandlstadt

+49 8756 969 49 79

FanDjango:
Compiler did notice. I was checking over all the millions of warnings and considering cleaning some of them up. So much XML stuff, it hides so many others.

It was a warning though. I suppose I think it actually should be an error.

PR #1096, opened 2022-12-10
Fix some XML warnings

FanDjango:
What's your opinion about handling some of the warnings, as time permits?

Added basic XML stuff - not fully fleshed - but anyone who visits these places and sees one could please take the time and flesh them out. Most of the ones with parameter names are more or less self-explaining, though, so it's not a huge thing.

Places like Profile.cs or Config.cs, or FtpClient.cs etc., which by design have no XML comments in the list of Fields or Methods I have bracketed with #PRAGMA disabling of the XML warnings.

In many places I have removed the warnings for unused "ex" variables in `catch` statements.

One or two methods I have moved within their files.

"Hide" of the base function in two places: **new** or **overide** to make the compiler happy and not issue a warning.

### DIFF
--- a/FluentFTP.Tests/Unit/ParserTests.cs
+++ b/FluentFTP.Tests/Unit/ParserTests.cs
@@ -19,7 +19,7 @@ namespace FluentFTP.Tests.Unit {
 				var test = testValues[i];
 
 				// parse it
-				FtpListItem item = null;
+				FtpListItem? item = null;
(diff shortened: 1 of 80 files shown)


robinrodricks:
Yes these are all good changes, you can merge it. In general, any constructive PR can be merged immediately, I have no issues.

Thanks for your work.

PR #1097, opened 2022-12-11
Clear up all warnings re. ParserTests.c

FanDjango:
Lots of warnings re. nullable vs. non-nullable.


robinrodricks:
Nice fixes, thanks!!

Issue #914, opened 2022-08-16
SetModifiedTime does not work on Azure FTPS connections


chack7:
**FluentFTP Version:** 39.3.0

**Framework:** .NET 4 

SetModifiedTime executes an MFMT command to set modified time.  This does not work on Azure FTPS connections which require MDTM command for both Getting and Setting modified time.

The correct command for Azure FTPS connections are

```
// Get Modified Time
MDTM /file/path

// Set Modified Time
MDTM YYYYMMDDHHMMSS /file/path
```

Many FTPS clients support this including BeyondCompare, FileZilla, solarwinds, etc...

The work around is to call the Execute command directly:

```
DateTime modTime = DateTime.UtcNow;
client.Execute($"MDTM {modTime:yyyyMMddHHmmss} {filepath}");
```

robinrodricks:
How do we detect azure? It does not show anything different. It looks like Windows OS.

robinrodricks:
So does the same thing work for Windows IIS FTP server?

chack7:
When I connect to FTPS on Azure using BeyondCompare, I get the following...

Recv> 220 Microsoft FTP Service
Recv> 211-Extended features supported:
         LANG EN*
         UTF8
         AUTH TLS;TLS-C;SSL;TLS-P;
         PBSZ
         PROT C;P;
         CCC
         HOST
         SIZE
         MDTM
         REST STREAM
        211 END

It does not mention MFMT.  I suppose if the 211-Extended features response doesn't include MFMT but does include MDTM might indicate support for MDTM YYYYMMDDHHMMSS /file/path support.

FanDjango:
@chack7 Could you please test the current master for this MDTM vs. MFMT issue on Azure? See PR #1103 which I have just merged.

PR #1103, opened 2022-12-13
If MFMT not supported but MDTM available...

FanDjango:
See issue #914

If MFMT is not supported (re. FEAT list), but MDTM **is supported**, attempt setting the file time with **MDTM**. This is not universally supported at all, but worth a try.


robinrodricks:
Wow, great feature!! Well done.

robinrodricks:
Simple and universal implementation, nice.

PR #1113, opened 2022-12-22
Check for async proxy

jnyrup:
Visual Studio warned that `AsyncFtpClient` could never be of type `FtpClientSocks4Proxy` or `FtpClientSocks4aProxy`.

I have not tested this is any way.


FanDjango:
Thanks for picking this up. I was wondering about this warning - it seemed to work, so was on back burner for me.

PR #1115, opened 2022-12-23
Add unit tests of `IsProxy()`

jnyrup:
Some parts of this library are more difficult to tests, but the changes in #1113 are easy-peasy to test, so here goes.


robinrodricks:
Thanks!

PR #1117, opened 2022-12-23
Logging exceptions

jnyrup:
Many logging frameworks separate the logging message from any other data, e.g. exceptions.
In visualization of logs logging message and exception then have separate columns.
This is similar to how parameterized SQL queries separates the query from the parameters.

Also for `ILogger`s it should not be necessary to explicitly prefix the logging message with the loglevel, as that is controlled in the logging format template.
E.g. for [serilog](https://github.com/serilog/serilog/wiki/Formatting-Output)

### DIFF
--- a/FluentFTP/Client/AsyncClient/Disconnect.cs
+++ b/FluentFTP/Client/AsyncClient/Disconnect.cs
@@ -17,7 +17,7 @@ namespace FluentFTP {
 					}
 				}
 				catch (Exception ex) {
-					LogWithPrefix(FtpTraceLevel.Warn, "AsyncFtpClient.Disconnect(): Exception caught and discarded while closing control connection: " + ex.ToString());
+					LogWithPrefix(FtpTraceLevel.Warn, "AsyncFtpClient.Disconnect(): Exception caught and discarded while closing control connection", ex);
 				}
 				finally {
 					m_stream.Close();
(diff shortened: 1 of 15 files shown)


robinrodricks:
Looks good, thanks.

FanDjango:
Would you not prefer to have the parameters not need to include a space character?

`var fullMessage = GetLogPrefix(eventType) + message + exception?.ToString();`

could prefer to be

`var fullMessage = GetLogPrefix(eventType) + " " + message + " " +  exception?.ToString();`

Thus the single parms would be clean?

FanDjango (approved):
I wonder what formatting for the message, the exception etc. would be most advantagous

Wiilf:
Multicolour ACII text preferably.

On Dec 23, 2022, at 3:05 PM, FanDjango ***@***.***> wrote:

﻿

@FanDjango approved this pull request.

I wonder what formatting for the message, the exception etc. would be most advantagous

robinrodricks:
> I wonder what formatting for the message, the exception etc. would be most advantagous

`GetLogPrefix` already adds spaces.

jnyrup:
> Would you not prefer to have the parameters not need to include a space character?

I agree and have fixed this.

FanDjango:
Hey @jnyrup last commit of 2022

FanDjango:
@jnyrup So sorry, there was a compile error. #1118 Just a leftover line needing to be removed. Is my fix ok, I hope?

PR #1119, opened 2022-12-30
Fix logging message

jnyrup:


FanDjango:
@jnyrup You got the last word for 2022, LOL

robinrodricks:
Do we need an urgent release or is the current release ok?

jnyrup:
> Do we need an urgent release or is the current release ok?

The current should be fine. This is only a fixup to the separation of the exception part.

PR #1123, opened 2023-01-06
Log exception

jnyrup:
Follow-up to #1122

### DIFF
--- a/FluentFTP/Streams/FtpSocketStream.cs
+++ b/FluentFTP/Streams/FtpSocketStream.cs
@@ -1283,8 +1283,7 @@ namespace FluentFTP {
 				// to be done and the exception can be re-thrown for
 				// handling down the chain. (Add logging?)
 				Close();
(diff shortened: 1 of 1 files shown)


FanDjango:
Hi @jnyrup Happy New Year!

I remember there was a reason for the `.trim()` and it looked nice that way on two lines. But regardless, of course this is nicer code.

robinrodricks:
Ok thanks.

Happy new year to you @FanDjango  and @jnyrup !!

jnyrup:
> I remember there was a reason for the `.trim()` and it looked nice (the output, that is) that way on two lines. But regardless, of course this is nicer code.

Makes sense.
You could change `LogWithPrefix` to include a newline between the custom message and the exception.
```cs
var exceptionMessage = exception is not null ? Environment.NewLine + exception.ToString() : null;
```

FanDjango:
Or even have a bool (with default false) to enable that behaviour

PR #1124, opened 2023-01-06
Simplify catch with 'when'

jnyrup:
Spotted some places where exception handling could be simplified with `when`.

### DIFF
--- a/FluentFTP/Streams/FtpSocketStream.cs
+++ b/FluentFTP/Streams/FtpSocketStream.cs
@@ -919,13 +919,8 @@ namespace FluentFTP {
 						((IInternalFtpClient)Client).LogStatus(FtpTraceLevel.Verbose, "...timeout connecting to IP #" + iPlusOne + "= " + logIp + ":" + port);
 					}
 				}
-				catch (Exception ex) {
-					if (lastIP) {
-						throw;
-					}
-					else {
-						((IInternalFtpClient)Client).LogStatus(FtpTraceLevel.Verbose, "...error connecting to IP #" + iPlusOne + "= " + logIp + ":" + port, ex);
-					}
+				catch (Exception ex) when (!lastIP) {
+					((IInternalFtpClient)Client).LogStatus(FtpTraceLevel.Verbose, "...error connecting to IP #" + iPlusOne + "= " + logIp + ":" + port, ex);
 				}
 			}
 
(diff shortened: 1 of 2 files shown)


FluentFTP/Streams/FtpSocketStream.cs
```diff
@@ -1097,18 +1087,14 @@ public class FtpSocketStream : Stream, IDisposable {
 					await EnableCancellation(m_socket.ConnectAsync(ipad, port), timeoutSrc.Token, () => DisposeSocket());
 				}
 			}
-			catch (SocketException ex) {
-				if (ex.SocketErrorCode == SocketError.OperationAborted ||
-					ex.SocketErrorCode == SocketError.TimedOut) {
-					throw new TimeoutException("Timed out trying to connect!");
-				}
-				throw;
+			catch (SocketException ex) when (ex.SocketErrorCode is SocketError.OperationAborted or SocketError.TimedOut) {
```
robinrodricks, line 1090: would this new statement be perfectly equivalent in behaviour to the old one? I'm asking for the case at line 1105.

FanDjango: @jnyrup My intention in the original code was to rethrow to a different kind of exception in those special cases. You have eagle eyes, @robinrodricks.

jnyrup: Before:
* catch all `SocketException`
  * if `OperationAborted` or `TimedOut` throw `TimeoutException`
  * else rethrow

After:
* catch `SocketException` if `OperationAborted` or `TimedOut` and throw `TimeoutException`
* else go to next catch handler;

I think these should be equivalent?

Just for inspiration, to exemplify what you can do with pattern matching, you can use the syntax to combine both catches to a single.

```cs
catch (Exception ex) when (ex is SocketException { SocketErrorCode: SocketError.OperationAborted or SocketError.TimedOut } or ObjectDisposedException) {
}
```

FanDjango: @jnyrup This is the one we are looking at, right?

```
			catch (SocketException ex) {
				if (ex.SocketErrorCode == SocketError.OperationAborted ||
					ex.SocketErrorCode == SocketError.TimedOut) {
				}
				throw;
			}
```
and after applying the change we have:
```
}
```
We are losing the `throw new TimeoutException("Timed out trying to connect!");` IMO. Or not?

jnyrup: Have a look at the new file, I think it's the diff that might be confusing.

The new code is
https://github.com/robinrodricks/FluentFTP/blob/e2a35c38dea517124a727bd28651c38dfc2f8179/FluentFTP/Streams/FtpSocketStream.cs#L1090-L1092
and
https://github.com/robinrodricks/FluentFTP/blob/e2a35c38dea517124a727bd28651c38dfc2f8179/FluentFTP/Streams/FtpSocketStream.cs#L1106-L1108

FanDjango: You are right, 

>  I think it's the diff that might be confusing.

It confuses me totally. I would need to see the result of the change in its entirety to understand. And there is a complication by (even nested) conditional compiles and a stupid common catch clause for both cond branches.

I trust you and will try to understand it when I see the finished code.

FanDjango: Which I have now looked at. I really don't understand the diff, but the new code looks totally good.

FanDjango: I made the changes on a branch of the current master: (sorry, screenshot):

image

So I still don't understand the first half of the PR diff. But the result is AOK.

FanDjango: Sorry for the trouble I might have caused.

PR #1126, opened 2023-01-07
Further optimise follow up for #1124

FanDjango:
@jnyrup So, look at this, if I rearrange the catch I can have a single common one at the end.

Too bad I cannot bring in the `ObjectDisposedException` as well, as it is encased in the outer #else-#endif


robinrodricks:
Looks good, thanks.

PR #1127, opened 2023-01-08
Write exception (optionally) on a new line

FanDjango:
This was suggested at the end of the discussion in #1123

### DIFF
--- a/FluentFTP/Client/BaseClient/Logger.cs
+++ b/FluentFTP/Client/BaseClient/Logger.cs
@@ -62,12 +62,12 @@ namespace FluentFTP.Client.BaseClient {
 		/// <param name="eventType">The type of tracing event</param>
 		/// <param name="message">The message to write</param>
 		/// <param name="exception">An optional exeption</param>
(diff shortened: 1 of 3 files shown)


robinrodricks:
Cool, thanks.

jnyrup:
I wouldn't recommend this.
See e.g. https://martinfowler.com/bliki/FlagArgument.html why having Boolean parameters might be problematic.

How about always printing exceptions on a newline?

FanDjango:
We actually have a far worse place where we do (I mean: I did) this sort of thing: See GetReply(...), or also ReadStaleData. You will shudder when you look at that, I suppose.

> How about always printing exceptions on a newline?

Well, that is probably the best solution then. I wasn't expecting @robinrodricks to merge this mere minutes after opening the PR, LOL and was expecting some discussion, I should have marked it draft.

robinrodricks:
We abuse arguments in this library. Its fine for now as its an internal method and not an API. I guess the bigger problem here is that our API methods have "too many arguments" and should probably use a config object.

PR #1129, opened 2023-01-08
Use `ConnectAsync` for net472

jnyrup:
`#if v472` is never true and caused `ConnectAsync` not to be used on net472.
See how VS highlights that with net472 selected, the else part is used.
image

The correct preprocessor directive would be `#if NET472`.

I also took the liberty to rearrange this a bit as most of the code is identical, except net462 should use `BeginConnect` and non-netstandard should also catch `ObjectDisposedException`.

The GitHub diff might not be the easiest way to look at this.

Fixup to #1056

### DIFF
--- a/FluentFTP/Streams/FtpSocketStream.cs
+++ b/FluentFTP/Streams/FtpSocketStream.cs
@@ -1077,29 +1077,18 @@ namespace FluentFTP {
 		/// <param name="port">The port to connect to</param>
 		/// <param name="token">The token that can be used to cancel the entire process</param>
 		private async Task<bool> ConnectAsyncHelper(IPAddress ipad, int port, CancellationToken token) {
-
-			int ctmo = this.ConnectTimeout;
-
-#if NETSTANDARD
 			try {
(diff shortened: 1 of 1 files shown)


FanDjango:
Well well, step by step one reaches maximum opt  :-)

I patched the diff and it certainly looks **a real lot** better that way. 

Thanks for finding my "v472" blunder, funny that no-one discovered that earlier.

Looks like this now:

```
		private async Task<bool> ConnectAsyncHelper(IPAddress ipad, int port, CancellationToken token) {
			try {
					timeoutSrc.CancelAfter(ConnectTimeout);
#if NET462
					var connectResult = m_socket.BeginConnect(ipad, port, null, null);
					await EnableCancellation(Task.Factory.FromAsync(connectResult, m_socket.EndConnect), timeoutSrc.Token, () => DisposeSocket());
#else
#endif
				}
			}
#if !NETSTANDARD
			catch (ObjectDisposedException) {
			}
#endif
			}

			return m_socket.Connected;
		}
```

PR #1131, opened 2023-01-10
Do not attempt Reconnect if we have not ever been connected before.

FanDjango:
It would "sort of" work, but fail in other places.  --> Revert back to an implicit Connect().

This happens when a user has not used Connect() (Or AutoConnect()) yet, but attempts to issue an API command, I did this by mistake today, and this would recognize that we are not connected and attempt a re-connect. This re-connect would for example fail to do a FEAT command and other stuff, causing unexpected results. Thus, in such a case, revert back to a normal full fledged connect.

I added a log message to show this happening.

Instead of adding a flag in Status, I just used the IP-Address internal cache having at least on entry. In case we ever invalidate that cache based on DNS time to live data, we would need to use a flag.

So @robinrodricks, should we use a flag?

Also, in case of **repeated** reconnects on a session, the current directory **can** be null on the second and subsequent reconnect. CWD does **not** set the status variable. It needs a specific PWD command to get it.

### DIFF
--- a/FluentFTP/Client/AsyncClient/Connect.cs
+++ b/FluentFTP/Client/AsyncClient/Connect.cs
@@ -43,6 +43,7 @@ namespace FluentFTP {
 			// If we have never been connected before...
 			if (this.Status.CachedHostIpads.Count == 0) {
 				reConnect = false;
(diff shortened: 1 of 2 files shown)


robinrodricks:
This is a good fix/idea.

If you want my opinion, I think using flags in `Status` is an elegant way to keep track of state.

Issue #1130, opened 2023-01-08
GetListing(recursive) and DownloadDirectory: failures/reconnects 


IrishChieftain:
**FTP Server OS:** Unix 

**FTP Server Type:** Vsftpd 

**Client Computer OS:** Windows 

**FluentFTP Version:** 43.0.1

**Framework:** .NET 6 / Worker Service

I am trying to download files from a remote Unix FTP server to a Windows server. It is a .NET 6 worker service which will run once every 24 hours. There are several top level folders (categories) on the remote server and under each of these are folders named after a date, such as 11_3_2022.  The service will look under each of the top level folders for sub-folders matching a particular date for that day. 

When testing a folder download with 4953 .wav files on FTP server (300 Mb), we usually fail to get between 21 and 27 files. It made no difference if the connection was static or not in our code. The actual failed count is maybe more because quite a few of the downloaded files have a .wav.filepart suffix.
**UPDATE:** Last two runs were both reporting the exact same 27 filenames that failed to download (26 .wav and 1 .wav.filepart files).

When downloading files at some point the client will connect to the PASV port of the server for a file like normal but after the 3 way handshake and the RETR command is sent to the server on the command channel, the command channel TCP connection is reset by the client (TCP RST).  This all happens within a fraction of a second.  The file never gets sent.  The client then reconnects and logs back in and continues downloading the next file but never got the previous one.

            using (var client = new FtpClient(uri, userName, password))
            {
                Connect(client);

                string processDate = "11_3_2022"; 
                string? directory;

                client.SetWorkingDirectory(ftpServerPath);
                string[] topDirectories = client.GetNameListing(ftpServerPath);
                FtpListItem[] items;
                List<FtpResult> results;

                for (int i = 0; i < topDirectories.Length; i++)
                {
                    directory = topDirectories.GetValue(i)?.ToString();
                    items = client.GetListing(directory + "/" + processDate, FtpListOption.Recursive);// Match processDate to folder

                    if (items.Length == 0) // If no sub-folder matching processDate, then skip
                        continue;
                    else
                    {
                        string lastTopFolderName = new DirectoryInfo(directory!).Name;
                        string? localTrailingFolders = lastTopFolderName + "\\" + processDate;

                        if (!client.DirectoryExists(callServerPath + localTrailingFolders))
                            Directory.CreateDirectory(callServerPath + localTrailingFolders);

                        int fails = 0;
                        try
                        {
[...]

Issue #1030, opened 2022-11-04
Sporadic reproducible GetListing, Download and other failures: IOException on control connection


FanDjango:
This has been reported quite a few times.

### What does the failure look like?

Typically, it manifests as an `IOException`. Depending on the API call being performed, this exception can cause 

- a `GetListing` result to be empty while the exeption is ignored
- other commands to fail with an `IOException`.

If you look closely at the `IOException`, you might see:
`an attempt was made to access a socket in a way forbidden by its access permissions`

### When does it happen?

It happens on an  **control connection** (only if encrypted) when you have issued many API calls (that execute commands and get the response). 

Reported by (ignore the incorrect title): 

@omanThYp https://github.com/robinrodricks/FluentFTP/issues/104#issuecomment-632437591
@oleksabor https://github.com/robinrodricks/FluentFTP/issues/104#issuecomment-953899294

#752 could such a case, but sadly there is no log

See also #723 : one of the "strange" ways this problem can show up.

FanDjango:
Some first results of testing, using @oleksabor initial approach as a basis. 

The goal was to find out a little bit more about the real problem and to remove conjecture.

1. I have modified the test program to **only** do the following (up to 500 times):

```
CWD /home/testuser
PWD
```

These commands **do not use** a data connection. No data connection (SSL) needs to be created and disposed. They only exercise the control connection.

These commands are iterated up to 500 times with no waits in between.

They always fail in the same way **on the 474th iteration**, on all of my available client systems.

2. The test suite will run all 500 iterations without failure if encryption is turned off.

3. The test suite encounters the _problem or multiple problems_ if encryption is turned on.

4. Disabling stream buffering does not help.

5. Disabling or enabling socket polling does not have any effect on the problem occurrence(s).

6. .NET 6.0 or .NET Framework 4.8 targets both exhibit the same problem(s).

7. Any problems occur after the same amount of iterations - deterministic and not random. Independent of host system.

[...]

PR #1135, opened 2023-01-12
Improve reconnect logic

FanDjango:
In addition to restoring the current working directory, one needs to restore the ASCII/BINARY data type setting on a reconnect.

Noticed this when I saw a reconnect happening on a failed GetFileSize command. The SIZE command (and many others) are sensitive to the DataType setting, which is reset to ASCII on many servers after a fresh connect.

### DIFF
--- a/FluentFTP/Client/AsyncClient/Connect.cs
+++ b/FluentFTP/Client/AsyncClient/Connect.cs
@@ -209,11 +209,28 @@ namespace FluentFTP {
 			}
 
 			if (reConnect) {
-				// go back to previous CWD
+				// Restore important settings from prior to reconnect
+
+				// restore Status.DataType
+				var oldDataType = Status.CurrentDataType;
+				Status.CurrentDataType = FtpDataType.Unknown;
+				await SetDataTypeNoLockAsync(oldDataType, token);
+
+				// restore Status.LastWorkingDir
 				if (Status.LastWorkingDir != null) {
 					await SetWorkingDirectory(Status.LastWorkingDir, token);
 				}
 			}
+			else {
+				// Set important settings
+
+				// set Status.DataType (also: FIX : #318)
+				Status.CurrentDataType = FtpDataType.Unknown;
+
+				// set Status.LastWorkingDir: no need
+			}
+
+			// Need to know where we are in case a reconnect needs to be done
 			_ = await GetWorkingDirectory(token);
 
 			// FIX #922: disable checking for stale data during connection
--- a/FluentFTP/Client/SyncClient/Connect.cs
+++ b/FluentFTP/Client/SyncClient/Connect.cs
@@ -199,18 +199,32 @@ namespace FluentFTP {
 				// Create the parser even if the auto-OS detection failed
 				CurrentListParser.Init(m_serverOS, Config.ListingParser);
 
-				// FIX #318 always set the type when we create a new connection
-				Status.CurrentDataType = FtpDataType.Unknown;
-
 				// Execute server-specific post-connection event
 				ServerHandler?.AfterConnected(this);
 
 				if (reConnect) {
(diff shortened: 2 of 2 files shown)


robinrodricks:
Thanks a lot for this!

Documented - https://github.com/robinrodricks/FluentFTP/wiki/Automatic-Reconnection

FanDjango:
Hi Robin,

 

yes, thanks to you to for documenting this so nicely. I am currently so busy down in the nitty gritty and then a new issue comes along (#1130) which shows up that #1030 will again and again come back to haunt us.

 

First, let me answer a different question of yours, slightly mixed up with stuff that has come to light with issue #1130:

 

"Waiting for response…" messages (albeit verbose).

 

It's been in there since… lemme think?... September? Long time, anyway - it was part of the GetReply revamp.

 

There have been cases in the past where it was not readily apparent at what stage of the internal processing the control connection had been dropped.

 

I will readily admit that this has mostly been after a RETR, STOR or APPE and not after a sundry other command.

 

Being of teutonic origin makes me a stickler in terms of "I need total information", it saves me from asking for test-runs with special branches often.

 

For me, it does not clutter up the logs, but instead gives the log a kind of bracketing structure that immediately causes a brain interrupt if the structure is incomplete.

 

COMMAND

<waiting for reply>

REPLY

 

Of course, one could make this smarter.

 

It is only useful if there is no response. That's one thing I could easily change.

 

Conversely, there was a report in issue #1130 of intermittent speed in his sequence of downloads and commands. I was actually thinking of adding response times to the <waiting for reply> messages, like making it a "Reply took 5 seconds" message. We do that for Ssl Handshake already…

 

So, as usual, one could be of many opinions on this. It's your product, so it's your decision.

 

Funny Note:

 

Issue #1130: The logging entry for disposing, by the way:

 

OpenPassiveDataStream(PASV, "RETR /home/ftpuser1/dire9/SDC/11_3_2022/9409107338 by ***@***.*** @ 11_28_02 AM.wav", 0)

Command:  PASV

Status:   Disposing FtpSocketStream(data connection)

Status:   Waiting for response to: PASV

Status:   Disposing FtpSocketStream(data connection)

Status:   Disposing FtpSocketStream(data connection)

Status:   Disposing FtpSocketStream(data connection)

Status:   Disposing FtpSocketStream(data connection)

Status:   Disposing FtpSocketStream(data connection)

Status:   Disposing FtpSocketStream(data connection)

Status:   Disposing FtpSocketStream(data connection)

Status:   Disposing FtpSocketStream(data connection)

Status:   Disposing FtpSocketStream(data connection)

Status:   Disposing FtpSocketStream(data connection)

Status:   Disposing FtpSocketStream(data connection)

Status:   Disposing FtpSocketStream(data connection)

Response: 227 Entering Passive Mode (128,136,216,217,50,202).

 

I recently added code to not show this message if the source is the finalizer. So these message in this example no longer appear.

 

But you can see that being verbose helps to understand a lot about what's going on:

 

1.       We have a lot of action between <Waiting for response to: PASV> and the <Response: 227…..>

2.       The recursive file listings (far above this excerpt) cause a lot of delayed garbage collectors calling the finalizer.

 

Wouldn't have seen that otherwise.

 

I repeat: With a heavy heart, I recently added code to not show this message if the source is the finalizer. So these message in this example no longer appear.

 

On the mainframe I am used to having detailed tracing and logging. I hate Windows where you click something and the user gets an error message like this:

 

"An error occurred"

 

Another Note:

 

I even tried some forced Garbage Collection, including waiting for the finalizers to finish after commands to SslStream, didn't help the SslStream corruption issue.

 

mit freundlichen Grüßen,

Michael Stiemke ***@***.***)

 

AnSyNova GmbH

Ringstr. 13

85405 Nandlstadt

+49 8756 969 49 79

PR #1152, opened 2023-01-14
Use expression instead of file for NuGet license

mus65:
You can't set both and expressions are better because it allows tools to automatically determine the license.

Also fix the copyright (copied from LICENSE.TXT)


robinrodricks:
Thanks

Issue #1195, opened 2023-02-13
IBM z/OS  Get JCL job output via FTP


osmansonmez:
**FTP Server OS:** Unix / Windows / Embedded

**FTP Server Type:** Vsftpd / glFTPd / BFTPd / ProFTPD / Pure-FTPd / IBM CS FTP (z/OS, OS/400) / Windows CE / Windows Server IIS / Vax / VMS / OpenVMS / Tandem / HP NonStop

**Client Computer OS:** Windows / Ubuntu / Debian / Linux Mint / Arch Linux

**FluentFTP Version:** 44.0.0.1

**Framework:** .NET 2 / 3.5 / 4 / 5 / 6 / 7 / 8 / UWP / Xamarin / Mono
.NET 5

We need to get JCL Job result from z/OS
I see a method OpenDataStream and if I send LIST {JobId} I can get Job Result.
I think that  if OpenDataStream can be public method we can use this easily.
 I write method like this. 

    /// <summary>
    /// Executes a command and get a stream
    /// </summary>
    /// <param name="command">The command to execute</param>
    /// <returns>The servers reply to the command</returns>
    public List<string> GetJCLResponse(string JobId)
		{
			List<string> listItems = new List<string>();
			using (var stream = OpenDataStream($"LIST {JobId} ", 0))
			{
				try
				{
					Log(FtpTraceLevel.Verbose, "+---------------------------------------+");
					{
						// GetListing will read file listings line-by-line (actually byte-by-byte)
						string buf;
						while ((buf = stream.ReadLine(Encoding)) != null)
						{
							if (buf.Length > 0)
							{
								listItems.Add(buf);
								Log(FtpTraceLevel.Verbose, "Listing:  " + buf);
							}
						}
					}

					Log(FtpTraceLevel.Verbose, "-----------------------------------------");
				}
				finally
				{
					stream.Close();
				}
			}

			return listItems;

		}

**Logs :**

```

<paste VERBOSE logs here>

```

FanDjango:
Perhaps read this:

https://github.com/robinrodricks/FluentFTP/wiki/IBM-zOS-and-OS-400-Support#accessing-the-jes2-internal-reader-to-submit-batch-jobs-and-how-to-list-and-retrieve-job-output

osmansonmez:
> 

FanGjango Thank you very much for quick response. İt is very helpful for us

osmansonmez:
Thank you. I have closed issue

[...]

PR #1196, opened 2023-02-14
GetCommandOutput

FanDjango:
See the issue #1195 

Add a way to issue arbitrary commands - and retrieve their output from the data connection, if they happen to do that.

Exotic systems sometimes have such commands (beyond the usual suspects like STOR, RETR, LIST, etc.) that we know nothing about.

This API call gives the user a way to access the output of such commands.

Note: This is for retrieving Line-by-line based text output. It is not for binary data in the current version and not meant as any kind of replacement for regular download operations.

Issue #1195 (look at the trailing end of the issue) gives us an example of such a functionality.

### DIFF
--- a/FluentFTP/Client/AsyncClient/GetCommandOutput.cs
+++ b/FluentFTP/Client/AsyncClient/GetCommandOutput.cs
@@ -15,8 +15,7 @@ namespace FluentFTP {
 	public partial class AsyncFtpClient {
 
 		/// <summary>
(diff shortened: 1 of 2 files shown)


robinrodricks:
Good work, thanks for this new API!

robinrodricks:
Released

https://www.nuget.org/packages/FluentFTP/45.0.0

PR #1201, opened 2023-02-18
Discover ssl session length

FanDjango:
As suggested by you, @robinrodricks, this API function will issue  up to `maxTrys=2000` commands `command="PWD"` with an inter-command delay of `delay=50` to help the user find out the recommended setting of `SslSessionLength` fot this server connection.

Usage:

```cs
	ftpClient.Connect();

	Console.WriteLine("Set SslSessionLength to less than: " + ftpClient.DiscoverSslSessionLength());
```


FanDjango:
Still unclean... strange

PR #1202, opened 2023-02-18
DiscoverSslSessionLength

FanDjango:
As suggested by you, @robinrodricks, this API function will issue  up to `maxTrys=2000` commands `command="PWD"` with an inter-command delay of `delay=50` to help the user find out the recommended setting of `SslSessionLength` fot this server connection.

Usage:

```cs
	ftpClient.Connect();

	Console.WriteLine("Set SslSessionLength to less than: " + ftpClient.DiscoverSslSessionLength());
```

### DIFF
--- a/FluentFTP/Client/SyncClient/DiscoverSslSessionLength.cs
+++ b/FluentFTP/Client/SyncClient/DiscoverSslSessionLength.cs
@@ -1,0 +1,63 @@
+﻿using System;
+using System.IO;
+using System.Text;
+using System.Collections.Generic;
+using FluentFTP.Exceptions;
+using FluentFTP.Helpers;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Runtime.CompilerServices;
+using FluentFTP.Client.Modules;
+using System.Security.Authentication;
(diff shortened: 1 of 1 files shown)


robinrodricks:
Looks good, if you have time please make an async version. If not I will try to do the moment I get time. Thanks.

FanDjango:
Sure, when I have time, but I actually did not think an Async version is needed at all? This API is more or less a one-shot thing a user would run.

robinrodricks:
True but now we have 2 clients. So some users might be using only the async one.

FanDjango:
I understand and agree. It's on my list, it's only a small thing..

 

mit freundlichen Grüßen,

Michael Stiemke ***@***.***)

 

AnSyNova GmbH

Ringstr. 13

85405 Nandlstadt

+49 8756 969 49 79

FanDjango:
It's on my list, I agree

robinrodricks:
Added async also.

https://www.nuget.org/packages/FluentFTP/45.1.0

FanDjango:
Wow. I removed it from my list...

Issue #648, opened 2020-11-16
Safe handle has been closed


HansR00:
**NOTE: I have a workaround by using version 32.3.1.0**

**FTP OS: Unix type L8

**FTP Server:** ProFTPD

**Computer OS: RPi Linux Raspbian 5.4.51-v7+ armv7l  and Mono 6.12.0.90

**FluentFTP Version: 33.0.2.0

Crash at connect (from my own logfile as no FTP logfile is produced) :

> 2020-11-16 08:40:26.783  InetSupport: Unknown Exception on FTPS connecting to ftp.meteo-wagenborgen.nl: Safe handle has been closed
> 2020-11-16 08:40:26.784  InetSupport: Failed FTPS connecting to ftp.meteo-wagenborgen.nl. Files will not be transferred
> 2020-11-16 08:40:26.784 Exception Unknown : Safe handle has been closed
> 2020-11-16 08:40:26.786 Data (cont): mscorlib
> 2020-11-16 08:40:26.798 Data:   at System.Runtime.InteropServices.SafeHandle.DangerousAddRef (System.Boolean& success) [0x0001c] in <9f0df102fe6e4cfea29d2e46f585d8a5>:0 
>   at System.Threading.WaitHandle.WaitOneNative (System.Runtime.InteropServices.SafeHandle waitableSafeHandle, System.UInt32 millisecondsTimeout, System.Boolean hasThreadAffinity, System.Boolean exitContext) [0x00008] in <9f0df102fe6e4cfea29d2e46f585d8a5>:0 
>   at System.Threading.WaitHandle.InternalWaitOne (System.Runtime.InteropServices.SafeHandle waitableSafeHandle, System.Int64 millisecondsTimeout, System.Boolean hasThreadAffinity, System.Boolean exitContext) [0x00014] in <9f0df102fe6e4cfea29d2e46f585d8a5>:0 
>   at System.Threading.WaitHandle.WaitOne (System.Int64 timeout, System.Boolean exitContext) [0x00000] in <9f0df102fe6e4cfea29d2e46f585d8a5>:0 
>   at System.Threading.WaitHandle.WaitOne (System.Int32 millisecondsTimeout, System.Boolean exitContext) [0x00019] in <9f0df102fe6e4cfea29d2e46f585d8a5>:0 
[...]

PR #1207, opened 2023-02-21
Important comment

FanDjango:
See issue #648 - needs some attention and see the comment itself


robinrodricks:
Hi Mike, don't worry about Mono. I don't want to target Mono anymore as we are doing .NET standard and .NET core which is enough for linux and everywhere.

robinrodricks:
We support Mono as a happenstance, not as a goal.

FanDjango:
I agree totally.

 

If possible, if it is true that only two statements are involved for fixing
most of it, then I would plead for "not supporting mono" but for documenting
the needed changes for users that want to try that.

 

If this is all it needs (and it sure looks like it):

 



 

mit freundlichen Grüßen,

Michael Stiemke ***@***.***)

 

AnSyNova GmbH

Ringstr. 13

85405 Nandlstadt

+49 8756 969 49 79

Issue #1205, opened 2023-02-20
 Can't connect to server via proxy when using EPSV mode for data connection


zubo100:
**FTP Server OS:** Unix

**FTP Server Type:** Vsftpd / glFTPd / BFTPd / ProFTPD / Pure-FTPd / IBM CS FTP (z/OS, OS/400) / Windows CE / Windows Server IIS / Vax / VMS / OpenVMS / Tandem / HP NonStop

**Client Computer OS:** Windows 

**FluentFTP Version:** 45.1.0

**Framework:** .NET 2 / 6

GetListing, Upload or Download from the server using EPSV over HTTP Proxy. 

**Logs :**

```
HTTP/1.0 200 Connection established

Status:   Waiting for a response
Response: 220 FTP Server 1.0
Command:  AUTH TLS
Status:   Waiting for response to: AUTH TLS
Response: 234 AUTH TLS successful
Warning:  SSL Buffering force disabled, is .NET 5.0 and later
Status:   FTPS authentication successful, protocol = Tls12
Status:   Time to activate encryption: 0h 0m 0s.  Total Seconds: 0,2593455.
Status:   SslStream: Tls12 (Aes256, TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384, 44550, 255)
Command:  USER ***
Status:   Waiting for response to: USER ***
Response: 331 Password required for ***
Command:  PASS ***
Status:   Waiting for response to: PASS ***
Response: 230 User *** logged in
Command:  PBSZ 0
Status:   Waiting for response to: PBSZ 0
Response: 200 PBSZ 0 successful
Command:  PROT P
Status:   Waiting for response to: PROT P
Response: 200 Protection set to Private
Command:  FEAT
Status:   Waiting for response to: FEAT
Response: 211-Features:
Response: AUTH TLS
Response: CCC
Response: CLNT
Response: EPRT
Response: EPSV
Response: HOST
Response: LANG en-US.UTF-8*;en-US
Response: MDTM
Response: MFF modify;UNIX.group;UNIX.mode;
Response: MFMT
Response: MLST modify*;perm*;size*;type*;unique*;UNIX.group*;UNIX.groupname*;UNIX.mode*;UNIX.owner*;UNIX.ownername*;
Response: PBSZ
Response: PROT
Response: RANG STREAM
Response: REST STREAM
Response: SIZE
Response: SSCN
Response: TVFS
Response: UTF8
Response: 211 End
[...]

PR #1217, opened 2023-02-24
Issue #1205 Repair logic to support proxy connections

FanDjango:
Ever since (some months ago) we have a DNS cache, where we store the SINGLE IPAD that we have successfully connected to, even if the server has multiple IP Addresses, we can rely on our own connect logic to pick up this **same** IPAD again on EPSV requests.

So therefore, the previous fix, by getting the IPAD from the socket endpoint, is no longer needed. Trouble was: On EPSV, this broke the proxy connect logic, where the proxy expects a DNS name and cannot make head not tail of the socket endpoint IP address.

Fixes issue #1205


robinrodricks:
https://www.nuget.org/packages/FluentFTP/45.2.0

PR #1234, opened 2023-03-14
#some times the `InfoMessages` maybe null

DUWENINK:
in my test case ,it works for me


FanDjango:
Ok, I suppose so. Do you have an example or description, just for my curiosity? And then we need to do the same for SYNC client as well?

DUWENINK:
yes ,I am test for muti-thread and the xunit report "Object null reference exception" and I find this

DUWENINK:
yes ,i am test for muti-thread and the xunit report "Object null reference exception" and I find this

robinrodricks:
Small fixes are good. Thanks!

FanDjango:
@robinrodricks and SYNC!?! And the exact circumstances certainly would be interesting? Who knows where else InfoMessages are queried and might also be null?

PR #1235, opened 2023-03-15
See #1234

FanDjango:


jnyrup:
Nice, I was just about to make a PR with the same changes + some more null checks of `Reply.InfoMessages`.
https://github.com/jnyrup/FluentFTP/commit/57b628fae15ee9002f3ecd9dbc75ec78d61f65ac

One way to improve remembering making changes for both sync and async code, could be to create a PR template.
For inspiration, see what we do in Fluent Assertions
https://github.com/fluentassertions/fluentassertions/blob/develop/.github/PULL_REQUEST_TEMPLATE.md

FanDjango:
Yeah, I have a few more ready too, I did a find all for InfoMessages for example. Let's not make a race out of it - I'll leave it for you, ok? Because I have another thing I am working on right now too...

FanDjango:
I had a look at yours, its congruent to mine but better in places, so why don't you PR them...
Still would really like to know the circumstances why InfoMessages may be null, in some cases one would perhaps throw something in GetReply?

PR #1236, opened 2023-03-15
More null checks of `Reply.InfoMessages`

jnyrup:
I can't tell the flow of what happens, but if we hit one of these `breaks` in `GetReplyInternal` we could skip writing to `reply.InfoMessages`.
image

Supersedes #1235.

### DIFF
--- a/FluentFTP/Client/AsyncClient/GetListing.cs
+++ b/FluentFTP/Client/AsyncClient/GetListing.cs
@@ -288,10 +288,12 @@ namespace FluentFTP {
 
 						Log(FtpTraceLevel.Verbose, "+---------------------------------------+");
 
-						foreach (var line in reply.InfoMessages.Split('\n')) {
-							if (!Strings.IsNullOrWhiteSpace(line)) {
-								rawlisting.Add(line);
-								Log(FtpTraceLevel.Verbose, "Listing:  " + line);
+						if (reply.InfoMessages != null) {
+							foreach (var line in reply.InfoMessages.Split('\n')) {
+								if (!Strings.IsNullOrWhiteSpace(line)) {
+									rawlisting.Add(line);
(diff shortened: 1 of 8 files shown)


FanDjango:
> I can't tell the flow of what happens, but if we hit one of these breaks in GetReplyInternal we could skip writing to

You are right.

Receive fails (from the beginning) but "connected" is true.

Not good, not good at all. There is a reason why "GetReply" is my favorite place to hack.

I did suspect that there was more to it than just "fixing" the symptoms ( @robinrodricks ! ). Thanks @jnyrup for your insight - I think in this case we need to throw - a system.timeoutexception.

I will work on that part. But fixing the "check for null" is still good practice, IMO. this PR is therefore still very important and should be merged.

FanDjango:
It's really also a question of what you expect "InfoMessages" to contain.

The `reply.message` will contain the message of the line that contains the return code and is a valid return code line.

The `reply.infomessages' will contain the message lines previous to the `reply.message` line. If the response contains only the message line, the infomessages should be null.

Or should the infomessage never be null - they should contain the message line as well?

FanDjango:
I would say that `InfoMessages` is empty, null, if there is a success reponse with associated success message, but no informational messages. 

In the users example, I suppose (`ObjectInfo`) which results in a `MLST´ is done and does not return any info.

I will test this tomorrow. It would have been so **much nicer, if we had a log of what the user did*, instead of jumping to conclusions...

FanDjango:
The reaction to a `MLST` might be server dependant - executing it for a non-existant file **should** not produce a _success_, but if the server does report _success_ and then simply issues no info messages we have got the case you analyzed.

But only if the user reports server type, provides a log and more information, will we ever know **exactly** what is happening.

robinrodricks:
Thanks so much!

Issue #1247, opened 2023-04-17
Dubious code in ConnectModule.IsPermanantConnectionFailure


fredericDelaporte:
**FTP Server OS:** No idea

**FTP Server Type:** `FileZilla Server 0.9.59 beta` as publicized by the server.

**Client Computer OS:** Windows 11

**FluentFTP Version:** 46.0.2

**Framework:** .NET 6

We had received once a failure from AutoDetect. The error was:
```
FluentFTP.Exceptions.FtpInvalidCertificateException: FTPS security could not be established on the server. The certificate was not accepted.
   at FluentFTP.Client.Modules.ConnectModule.AutoDetectAsync(AsyncFtpClient client, Boolean firstOnly, Boolean cloneConnection, CancellationToken token)
   at FluentFTP.AsyncFtpClient.AutoDetect(Boolean firstOnly, Boolean cloneConnection, CancellationToken token)
```

Looking at the code of `AutoDetectAsync`, it seems to throw directly only if `IsPermanantConnectionFailure` (sic, should be spelled `IsPermanentConnectionFailure`) has returned an exception.
https://github.com/robinrodricks/FluentFTP/blob/f2a171f83776e27f1d350a1ea3ae68b8e5a69f9e/FluentFTP/Client/Modules/ConnectModule.cs#L251-L260

Looking at `IsPermanantConnectionFailure`:
https://github.com/robinrodricks/FluentFTP/blob/f2a171f83776e27f1d350a1ea3ae68b8e5a69f9e/FluentFTP/Client/Modules/ConnectModule.cs#L372-L382
`FtpInvalidCertificateException` is supposed to only be thrown with an inner exception which could have provided more information. But unfortunately `FtpInvalidCertificateException` not only erases `Exception.InnerException` property but also do not call a base constructor allowing to set `Exception.InnerException`. As a result, standard exception logging cannot get the inner exception details.

See here:
https://github.com/robinrodricks/FluentFTP/blob/f2a171f83776e27f1d350a1ea3ae68b8e5a69f9e/FluentFTP/Exceptions/FtpInvalidCertificateException.cs#L15-L28

So, as the error only occurred once and do not reproduce, we cannot debug this, and the relevant information was lost due to the way the failing code path is coded.

By the way, reading a bit more the code of `IsPermanantConnectionFailure`:
[...]

PR #1250, opened 2023-04-18
Re-ordering of code re. #1247

FanDjango:
To address the noted "dead code" and general cleanup of `Is PermanentConnectionFailure.cs`:

Essentially only a reordering of the code if-segments.


robinrodricks:
Thank you!

PR #1260, opened 2023-05-10
Create a Config.EnableBlueCoatProxyHack boolean to enable Hack #409

### LINKED ISSUE #409: Proxy bluecoat issue when Proxy IP is like 172.x.x.x
After 2 days of research I just found the reason of my issue.
I have to go through a Bluecoat proxy (I'm using FtpClientUserAtHostProxyBlueCoat).
The connection was ok but then nothing worked anymore (list directory, upload, download, ...).
The problem comes from my proxy that I call by its IP: 172.x.x.x.
Indeed in the "OpenPassiveDataStream" method there are these lines

	//use host ip if server advertises a non-routeable IP
	m = Regex.Match(host, @"(^10\.)|(^172\.1[6-9]\.)|(^172\\.2[0-9]\.)|(^172\.3[0-1]\.)|(^192\.168\.)|(^127\.0\.0\.1)|(^0\.0\.0\.0)");

	if (m.Success)
	{
		host = m_host;
	}

and in my case it replaces the name of my proxy by the name of the FTP server and then, nothing works.
My suggestion is to test if you are not in a proxy connection :
```
	if (!this.IsProxy()) // Update
	{
		//use host ip if server advertises a non-routeable IP
		m = Regex.Match(host, @"(^10\.)|(^172\.1[6-9]\.)|(^172\\.2[0-9]\.)|(^172\.3[0-1]\.)|(^192\.168\.)|(^127\.0\.0\.1)|(^0\.0\.0\.0)");

		if (m.Success)
		{
			host = m_host;
		}
	}
```
There is the same thing in the "OpenPassiveDataStreamAsync" method.
Is this a valid solution, or am I an exception? 
What do you think of my update?

acorchia:
enable fix #409 only if the boolean is true

### DIFF
--- a/FluentFTP/Client/BaseClient/GetPassivePort.cs
+++ b/FluentFTP/Client/BaseClient/GetPassivePort.cs
@@ -73,7 +73,7 @@ namespace FluentFTP.Client.BaseClient {
 			port = (int.Parse(m.Groups["port1"].Value) << 8) + int.Parse(m.Groups["port2"].Value);
 
 			// Fix #409 for BlueCoat proxy connections. This code replaces the name of the proxy with the name of the FTP server and then nothing works.
-			if (!IsProxy()) {
+			if (Config.EnableBlueCoatProxyHack && !IsProxy()) {
 				//use host ip if server advertises a non-routable IP
 				m = Regex.Match(host, @"(^10\.)|(^172\.1[6-9]\.)|(^172\.2[0-9]\.)|(^172\.3[0-1]\.)|(^192\.168\.)|(^127\.0\.0\.1)|(^0\.0\.0\.0)");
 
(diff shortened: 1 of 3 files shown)


FanDjango:
I agree with the problem that you are addressing and can understand the motivation for this. I would prefer to bring the GetxxxPort logic back to the basics and reduce the complexity - instead of catering to special "proxy-type", "server-type", "network-type" hacks, I prefer a connection-specific type to address this problem. Please see my alternative suggestion #1262

acorchia:
This look good.

And I agree this is much more consistent with the logic of the tool.

 

De : FanDjango ***@***.***> 
Envoyé : mercredi 10 mai 2023 23:11
À : robinrodricks/FluentFTP ***@***.***>
Cc : acorchia ***@***.***>; Author ***@***.***>
Objet : Re: [robinrodricks/FluentFTP] Create a Config.EnableBlueCoatProxyHack boolean to enable Hack #409 (PR #1260)

 

I agree with the problem that you are addressing and can understand the motivation for this. I would prefer to bring the GetxxxPort logic back to the basics and reduce the complexity - instead of catering to special "proxy-type", "server-type", "network-type" hacks, I prefer a connection-specific type to address this problem. Please see my alternative suggestion #1262 <https://github.com/robinrodricks/FluentFTP/pull/1262>

FanDjango:
Ok, then I will close this PR, thanks for pointing this out.

Issue #1258, opened 2023-05-10
"Ugly hack" for BlueCoat proxy makes pasv connection fail on legitimate use in private network address ranges


acorchia:
**FTP Server OS:** Unix / Windows / Embedded
Linux but true for any
**FTP Server Type:** Vsftpd / glFTPd / BFTPd / ProFTPD / Pure-FTPd / IBM CS FTP (z/OS, OS/400) / Windows CE / Windows Server IIS / Vax / VMS / OpenVMS / Tandem / HP NonStop
Any
**Client Computer OS:** Windows / Ubuntu / Debian / Linux Mint / Arch Linux
Ubuntu but true for any
**FluentFTP Version:** ?

**Framework:** .NET 2 / 3.5 / 4 / 5 / 6 / 7 / 8 / UWP / Xamarin / Mono
6
<write details about your bug report / feature request here>
In GetPassivePort.cs line 75
			// Fix #409 for BlueCoat proxy connections. This code replaces the name of the proxy with the name of the FTP server and then nothing works.
			if (!IsProxy()) {
				//use host ip if server advertises a non-routable IP
				m = Regex.Match(host, @"(^10\.)|(^172\.1[6-9]\.)|(^172\.2[0-9]\.)|(^172\.3[0-1]\.)|(^192\.168\.)|(^127\.0\.0\.1)|(^0\.0\.0\.0)");

				if (m.Success) {
					host = m_host;
				}

This broke the connection of PASV as the host is not the one returned by the server !
At least enable this hack with a non-defaulted option in the config

```

FanDjango:
Thanks for the friendly report. 

I would greatly appreciate that you provide a nicely documented example of why/how

> This broke the connection of PASV as the host is not the one returned by the server !

so that we can understand the reasoning behind your polite request:

> At least enable this hack with a non-defaulted option in the config

Maybe in between you can also provide feedback on the we did for you in GnuTLS in your other issue, never heard from you again.

FanDjango:
So a legitimate (working correctly) proxy **inside** a (big) internal network.

Allow me to edit the title of the issue. It is an "ugly" hack.

FanDjango:
```cs
if (!IsProxy()) {
//use host ip if server advertises a non-routable IP
m = Regex.Match(host, @"(^10.)|(^172.1[6-9].)|(^172.2[0-9].)|(^172.3[0-1].)|(^192.168.)|(^127.0.0.1)|(^0.0.0.0)");

[...]

Issue #270, opened 2018-02-01
Unroutable address for passive mode


elmar69:
FTP Server: ProFTPD 1.3.5

I'm not shure about this really beeing an bug but filezilla does handle the Problem.
On Opening an passive connection the server does respond
"227 Entering Passive Mode (192,168,178,2,246,61)."
This raises the problem that trying to connect does timeout since the mentioned IP address is an unrouted private one.

Filezilla in this case does ignore the IP mentioned and does connect to the host IP after printing an warning.

It is an misconfiguration of the server but I would estimate it to happen often if filezilla does implement an explicit workaround to that.

martinvonwittich:
The problem in this situation is that ProFTPd doesn't know its own public IP address (because it is situated behind a NAT router), but the FTP spec forces the server to respond to `PASV` with both the IP address and the port. ProFTPd will respond with the IP address from the network interface where it originally received the FTP connection, which would be fine if this interface was directly connected to the Internet (then the interface would have the public IP), but in this case the interface has the private IP address 192.168.178.2.

In the past (with unencrypted connections) this problem was alleviated by the fact that NAT routers usually simply rewrite the PASV response and replace the private IP with the correct public IP, but this no longer works in a modern setup where encryption is enforced because due to the encryption, the router can neither see nor rewrite the PASV response.

A possible workaround on the server side would be to use the [MasqueradeAddress](http://www.proftpd.org/docs/directives/linked/config_ref_MasqueradeAddress.html) option to tell ProFTPd the correct public IP address. Unfortunately the typical use case for NAT routers are residential DSL connections with dynamic IPs that change every 24 hours, which makes this workaround complicated to set up because it would require a script that monitors the public IP, automatically updates the ProFTPd configuration and then restarts it.

A better solution is to use the more modern `EPSV` command specified by [RFC2428](https://tools.ietf.org/html/rfc2428) on the client side. The `EPSV` response will only contain the port and no IP address, and the client is then supposed to connect to the same IP address as for the control connection.

robinrodricks:
We had a fix for this in [22](https://github.com/robinrodricks/FluentFTP/blob/master/RELEASES.md#2200). Can you download the latest via nuget and see if it works?

robinrodricks:
I'll reopen this if you still get errors.

PR #1262, opened 2023-05-10
Better GetPassivePort (and GetEnhancedPassivePort)

FanDjango:
See issues #1258, #409, #270

Sequential logic pattern makes this easer to maintain.

Supersedes #1260, if everyone is agreeable to this.


FanDjango:
I will rebase and redo the PR

Issue #1251, opened 2023-05-04
Cannot connect to FTPS IIS server on windows 2019 from Azure functions V4


anandgmenon:
**FTP Server OS:** Windows

**FTP Server Type:** Windows Server IIS

**Client Computer OS:** N/A

**FluentFTP Version:** 46.0.2

**Framework:** .NETCore 3.1

There's an issue connecting to FTPS IIS server on Windows 2019 from Azure functions v4 running on .NETCore3.1. The SSL handshake seems to be timing out. 
The error from the library is : _Unable to read data from the transport connection: An existing connection was forcibly closed by the remote host. An existing connection was forcibly closed by the remote host_ .
On checking the network traces we noticed Client key exchange is not getting initiated somehow.
![Screenshot 2023-05-04 102453](https://user-images.githubusercontent.com/16523769/236115505-edcfc8ac-216b-44e9-8316-86f9cbf7403c.png)
The issue doesn't  seem to happen on functions V3 and also doesn't happen on .NET5 and above.

**Code:**

```cs
var client = new AsyncFtpClient(host: "www.anand-server.com", port: 21, user: "<user>", pass: "<pass>");
client.Config.EncryptionMode = FtpEncryptionMode.Explicit;
client.LegacyLogger = (level, msg) => log.LogInformation(msg);
await client.Connect();
var response = await client.GetListing("/");
```

We tried with a different FTP library and it can connect successfully.

**Logs :**

```

[2023-05-04T04:54:09.133Z] >         Connect(False)
[2023-05-04T04:54:09.134Z] Status:   FluentFTP 46.0.2.0
[2023-05-04T04:54:09.148Z] Status:   Connecting to IP #1= ***:21
[2023-05-04T04:54:09.389Z] Status:   Waiting for a response
[2023-05-04T04:54:09.620Z] Response: 220 Microsoft FTP Service [738643.204d]
[2023-05-04T04:54:09.632Z] Status:   Detected FTP server: WindowsServerIIS
[2023-05-04T04:54:09.641Z] Command:  AUTH TLS
[2023-05-04T04:54:09.649Z] Status:   Waiting for response to: AUTH TLS
[2023-05-04T04:54:09.916Z] Response: 234 AUTH command ok. Expecting TLS Negotiation. [269ms]

```

FanDjango:
> On checking the network traces we noticed Client key exchange is not getting initiated somehow.

Client Hello, Server Hello and then silence. I agree.

[...]

PR #1264, opened 2023-05-11
Tidy up the conditional compiles and test for each target in FtpSslStream.cs

FanDjango:
I decided to clean up the conditional compiles and check each one for validity.

It turns out that .NET Framework supports ShutDownAsync since .NET 4.7 ( !!! ) which we somehow didn't really see so clearly at the time.

Therefore the only one of our build targets left over would be .NET 4.6.2, which needs the ugly reflection hack to send a close notify.

**BUT** Beware: .NET Standard 2.0 **also does not have this capability** which means theoretically we should develop the same hack for this target as for .NET 4.6.2. So I added that.

If this ever becomes an issue for someone, this will be needed.

I stumbled across this whilst searching for solution to #1251 

Read this:
Here is a quote from: https://github.com/dotnet/standard/issues/598#issuecomment-352148072
"The SslStream.ShutdownAsync API was added to .NET Core 2.0. It was also added to .NET Framework 4.7. Logically, since .NET Core 2.0 and .NET Framework 4.7.1 are aligned with NETStandard2.0, it could
have been part of the NETStandard20 definition. But it wasn't due to when the NETStandard2.0 spec
was originally designed."

### DIFF
--- a/FluentFTP/Streams/FtpSslStream.cs
+++ b/FluentFTP/Streams/FtpSslStream.cs
@@ -19,11 +19,14 @@ namespace FluentFTP.Streams {
 	/// See: https://learn.microsoft.com/en-us/windows/win32/secauthn/using-sspi-with-a-windows-sockets-client?source=recommendations
 	///
 	/// Note:
+	/// It seems we need this reflection kludge ONLY for .NET Framework 4.6.2. All our other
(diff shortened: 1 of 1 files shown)


robinrodricks:
Thanks, too technical for me but the code looks good so accepting.

PR #1265, opened 2023-05-11
New connection type PASVUSE, better GetPassivePort.cs

FanDjango:
See issues #1258, #409, #270

Sequential logic pattern makes this easer to maintain.

### DIFF
--- a/FluentFTP/Client/AsyncClient/OpenDataStream.cs
+++ b/FluentFTP/Client/AsyncClient/OpenDataStream.cs
@@ -41,6 +41,7 @@ namespace FluentFTP {
 
 					case FtpDataConnectionType.PASV:
 					case FtpDataConnectionType.PASVEX:
+					case FtpDataConnectionType.PASVUSE:
 						type = FtpDataConnectionType.EPSV;
 						Log(FtpTraceLevel.Info, "Changed data connection type to EPSV because we are connected with IPv6.");
 						break;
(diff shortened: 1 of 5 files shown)


robinrodricks:
Thanks, very nice, but can we improve the enum name - suggestions:

- PASVPrivate
- PASVForce
- PASVServerHost

I hate the naming of this enum, the values were made by the original author.

I would have preferred things like "PassiveExtended" which is more in line with .NET standards. So at least for new values can we name it nicely.

FanDjango:
That's why I left it here until you have time. I knew... and I mentioned that in the respective issue.

How about introducing some "synonyms" for the hated ones?

FanDjango:
```
PASV
EPSV
PORT
EPRT
```
actually correspond to the command issued to the server "behind the scenes". Note the archaic 4-character command syntax.

All others are "inventions" with modified meaning:
```
AutoPassive
AutoActive
```
Meaning: "Try the extended action (EPSV, EPRT) and if not supported, fall back to the normal action (PASV, PORT)".

```
PASVEX
```
This one is really the first "odd one out".

I suggest:

Keep: "AutoPassive", "AutoActive"

Keep: "PORT", "EPRT", "PASV", "EPSV" and tolerate **"PASVEX"** (for API continuity)

Add synonyms for the "non-4-char" one:

"PASVEX": "ExtendedPassive". It provides EPSV functionality for servers that **only** support PASV.

Add: "PassiveAllowUnroutable" for this PR's new one

FanDjango:
I have added the mentioned synonyms etc. to the PR.

robinrodricks:
Very nice, thank you!

Issue #1286, opened 2023-07-12
HASH response from files.com not being parsed correctly


smboyer:
**FTP Server OS:** Unix
**FTP Server Type:** BrickFTP (files.com / Exavault.com)
**Client Computer OS:** Windows
**FluentFTP Version:** 46.02
**Framework:** .NET 6

The response to command HASH from files.com (who recently acquired exavault) is not being parsed correctly.  Complete logs pasted below.  Logging via Serilog. As you can see in the following entries the Hash value is being returned as "0-55".  
```
[INF] Command:  HASH /ftp-test/ANewFile.txt
[INF] Response: 213 MD5 0-55 fcb217f0a094ad05f572e3678615b9d1 [46ms]
[INF] Compare Results = 
Size:Equal; 
Checksum:NotEqual; 
ModDate:NotEqual; 
Hash.Value: 0-55
```

**Logs :**

```

[INF] >         Connect(False)
[INF] Connecting to IP #1= ***:21
[INF] Response: 220 Service ready for new user. [738712.896d]
[INF] Command:  AUTH TLS
[INF] Response: 234 Command AUTH okay; starting TLS connection. [47ms]
[WRN] SSL Buffering force disabled, is .NET 5.0 and later
[INF] FTPS authentication successful, lib = .NET SslStream, cipher suite = Tls12 (Aes128, TLS_RSA_WITH_AES_128_CBC_SHA, RsaKeyX, 4096) [262ms]
[INF] Command:  USER ***
[INF] Response: 331 User name okay, need password for ***. [49ms]
[INF] Command:  PASS ***
[INF] Response: 230 User logged in, proceed. [156ms]
[INF] Command:  PBSZ 0
[INF] Response: 200 Command PBSZ okay. [43ms]
[INF] Command:  PROT P
[INF] Response: 200 Command PROT okay. [42ms]
[INF] Command:  FEAT
[INF] Response: 211 End [43ms]
[INF] Text encoding: System.Text.UTF8Encoding+UTF8EncodingSealed
[INF] Command:  OPTS UTF8 ON
[INF] Response: 200 Command OPTS okay. [41ms]
[INF] Command:  SYST
[INF] Response: 215 UNIX Type: BrickFTP Server [41ms]
[INF] Command:  PWD
[INF] Response: 257 "/" is current directory. [43ms]
[INF] Uploading: ANewFile.txt
[...]

Issue #722, opened 2021-05-14
Verify Not working - Hash command not being parsed correctly


dbman86:
**FTP OS:** not sure - using files.com

**Computer OS:** Windows 10

**FluentFTP Version:** 34.0.0

Server is returning this value from Hash command:

[R] HASH /pathtomyfile.
[R] 213 MD5 0-170500096 3197bf4ec5fa2d441c0f50264ca52f11

however, when the GetChecksum() method parses this to create FtpHash object, it is parsing the value as 0-170500096 instead of  3197bf4ec5fa2d441c0f50264ca52f11.

Not sure if this is the authoritative standard but this indicates the value FluentFTP is taking for the hash value is really the bit rate of the file... ?
https://www.jscape.com/blog/bid/48215/New-FTP-Command-HASH-for-Requesting-Hash-of-a-File

robinrodricks:
Please try this build.

[FluentFTP.35.0.0_BETA6.zip](https://github.com/robinrodricks/FluentFTP/files/7071196/FluentFTP.35.0.0_BETA6.zip)

robinrodricks:
Released.

https://www.nuget.org/packages/FluentFTP/35.0.0

robinrodricks:
Hi @dbman86, can you try the latest build from nuget and tell me if it resolves your issue? I need feedback on this please.

FanDjango:
Here is the **real** fix for this problem:

And see also #1287.

QUOTE from #1286
BrickFTP (files.com / Exavault.com) responds in a non - standard fashion (no filename).

Changed the regex to handle that situation.

Incidentally: The fix for https://github.com/robinrodricks/FluentFTP/issues/722 (which was never confirmed) which attempts to remove "213" from the server response as a fix is obsolete (the server response code is removed prior to the call of this function anyway) and not the real reason why https://github.com/robinrodricks/FluentFTP/issues/722 was reported. https://github.com/robinrodricks/FluentFTP/issues/722 was another, first occurrence of "missing" filename. And lo and behold, it was the same server type as the one reported here.
ENDQUOTE

PR #1287, opened 2023-07-13
Fix for #1286

FanDjango:
BrickFTP (files.com / Exavault.com) responds in a non - standard fashion (no filename).

Changed the regex to handle that situation.

Incidentally: The fix for #722 (which was never confirmed) which attempts to remove "213" from the server response as a fix is obsolete (the server response code is removed prior to the call of this function anyway) and not the real reason why #722 was reported. #722 was **another, first** occurrence of **"missing" filename**. And lo and behold, it was the same server type as the one reported here.


robinrodricks:
Release.

https://www.nuget.org/packages/FluentFTP/47.0.0

PR #1293, opened 2023-07-20
Add requireEncryption and includeImplicit

FanDjango:
Add two more parms (defaulting so that no code changes are needed in the user base) to `AutoDetect`.

**`bool requireEncryption`** ( = false )
**`bool includeImplicit`** ( = true )

These defaults reflects the previous codes **intentions** ( even if it didn't work so well... ).

I would actually **recommend** 

**`bool requireEncryption`** ( = **true** )
**`bool includeImplicit`** ( = **false** )

Can we do that?


robinrodricks:
Can we please take an object `FtpAutoDetectConfig` and add all the 4 options there?

Issue #1289, opened 2023-07-19
AutoDetect throws Security.Authentication.AuthenticationException if server supports TLSv1 only


dmansurov83:
**FTP Server OS:** Unix

**FTP Server Type:** ProFTPD

**Client Computer OS:** Windows 

**FluentFTP Version:** 47.0.0

**Framework:** Framework 4.6.2

I'm trying to call AutoDetect, but it throws an exception:
System.Security.Authentication.AuthenticationException: A call to SSPI failed, see inner exception. ---> System.ComponentModel.Win32Exception: The client and server cannot communicate, because they do not possess a common algorithm

**Logs :**

```

# AutoDetect(False, False)
2023-07-19 06:51:40.2089| INFO|Connecting to ftp://ftp.***.**

# Connect(False)
Status:   FluentFTP 47.0.0.0(.NET Framework 4.6.2)
Status:   Connecting to IP #1= ***:21
Status:   Waiting for a response
Response: 220 ProFTPD 1.3.5b Server (SNPI) [::ffff:**.**.**.**] [738719.286d]
Status:   Detected FTP server: ProFTPD
Command:  AUTH TLS
Status:   Waiting for response to: AUTH TLS
Response: 234 AUTH TLS successful [77ms]
Status:   Closing/Disposing FtpSocketStream(control connection)
Error:    FTPS Authentication failed, lib = .NET SslStream
Error:    A call to SSPI failed, see inner exception
```

Looks like this is because of the protocols priority:
``` 
    private static List<SslProtocols> DefaultProtocolPriority = new List<SslProtocols>()
    {
      SslProtocols.Tls11 | SslProtocols.Tls12,
      SslProtocols.Default
    };
```
Am I right? Is any way to make profile autodetection work?

FanDjango:
Can you provide the tls.conf file of the server? I am trying to set up such a szenario. Mine fails too, but with a different exception.

Can you "Connect" manually, if so - with which code/parms?

dmansurov83:
> Can you provide the tls.conf file of the server?

No, server isn't manages by us and I don't know the exact server config.

> Can you "Connect" manually, if so - with which code/parms?
I'm trying to use the manual profile like:
``` 
var activeProfile = new FtpProfile()
					{
						Host = _ftpClient.Host,
						Credentials = _ftpClient.Credentials,
						Encoding = Encoding.UTF8,
						Encryption = FtpEncryptionMode.Explicit,
[...]

PR #1305, opened 2023-07-24
Auto-print config file properties

FanDjango:
This is my take on how to do this.

So there is no need to put the 
```cs
		public object[] ToArray() {
			return new object[] {
				"CloneConnection="+CloneConnection,
				"FirstOnly="+FirstOnly,
				"IncludeImplicit="+IncludeImplicit,
				"RequireEncryption="+RequireEncryption,
				"ProtocolPriority="+ProtocolPriority,
			};
```
array as duplicate typing into the config file.

The output is generated automagically (here is an example):
```txt
>         AutoDetect(CloneConnection=true, FirstOnly=false, IncludeImplicit=true, RequireEncryption=false, ProtocolPriority=Tls11 Tls12, ProtocolPriority=Default)
>         AutoConnect()
>         AutoDetect(CloneConnection=false, FirstOnly=true, IncludeImplicit=true, RequireEncryption=false, ProtocolPriority=Tls11 Tls12, ProtocolPriority=Default)
Status:   Auto-Detect trying encryption mode "Auto" with "Tls11, Tls12"
>         Connect(False)
```
**Note:** To show how it works for arrays or lists, I have added a second protocol into the list, which normally contains only one. **it is just an example, one should not use "Default"**

Hopefully, config file objects do not become incredibly complex, as then the serializer must be improved to handle 2nd, 3rd or however deep subchilding.

The simple idea was to use XML serializing outwards on the config object. Then to read it back in as an XML document and to examine the nodes, which lo-and-behold have now been formatted for us.

Since this would be overkill for primitive types (coming from the legacy API call formatting) this is only executed if a true "object" is passed.

### DIFF
--- a/FluentFTP/Client/AsyncClient/AutoDetect.cs
+++ b/FluentFTP/Client/AsyncClient/AutoDetect.cs
@@ -18,10 +18,10 @@ namespace FluentFTP {
 		/// <param name="config">The coresponding config object for this API</param>
 		/// <param name="token">The token that can be used to cancel the entire process</param>
 		/// <returns></returns>
-		public async Task<List<FtpProfile>> AutoDetect(FtpAutoDetectConfig config, CancellationToken token = default(CancellationToken)) {
+		public async Task<List<FtpProfile>> AutoDetect(FtpAutoDetectConfig config = null, CancellationToken token = default(CancellationToken)) {
 			var results = new List<FtpProfile>();
 
-			LogFunction(nameof(AutoDetect), new object[] { config });
+			LogFunction(nameof(AutoDetect), config);
 			ValidateAutoDetect();
 
 			return await ConnectModule.AutoDetectAsync(this, config, token);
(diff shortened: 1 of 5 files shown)


robinrodricks:
Your approach is ok but its very slow to use XML Serializer for each function call. Thats why I had the most "direct" method which was converting an object into an array. If you want to make a faster serializer then just use reflection. Its easy like this:

```cs
using System;
using System.Reflection;

    public static string PrintObjectProperties(object obj)
    {
        Type type = obj.GetType();
        PropertyInfo[] properties = type.GetProperties(BindingFlags.Public | BindingFlags.Instance);

        StringBuilder result = new StringBuilder();
        foreach (var property in properties)
        {
            string propertyName = property.Name;
            object propertyValue = property.GetValue(obj);

            StringBuilder.Append($"{propertyName} = {propertyValue}\n");
        }

        return result.ToString();
    }
```

FanDjango:
Roger Wilco

robinrodricks:
Thank you

FanDjango:
I give up. I have more important things to do than muck around with this.

All I wanted was to fix Auto-Detect, add 2 or 3 useful optional parameters and close #1289 ASAP, which could have been released ages ago.

robinrodricks:
@FanDjango can you merge your latest code whatever half-works. I'll debug it and fix it this week.

robinrodricks:
Released.

https://www.nuget.org/packages/FluentFTP/47.1.0

robinrodricks:
@FanDjango sorry about the issues you faced. I fixed it and released it.

PR #1307, opened 2023-07-31
Avoid creating strings when logging is disabled

jnyrup:
When logging is disabled we still do a lot of work only not to log it.
This PR introduces `AnyLoggingIsEnabled()` to avoid this work if no kind of logging is enabled.


robinrodricks:
Good one, thanks!

FanDjango:
I agree. I wonder to perhaps somehow to pass this on to GnuTLS?

Issue #1320, opened 2023-09-01
Follow .NET Standard, not obvious how to use the library.


jsgoupil:
Currently, I'm on version 47.1.0.

I would not expect to have to read so much documentation to make this work, nor doing 10 trials and errors to download 1 file.

Since this library is followed by a lot of people and probably the de-facto library for FTP communication, maybe it would be great if it followed .NET standard? But this is probably a bigger discussion if the authors want to do so because it would be quite the breaking change.

 * Don't have an Async / Sync class. Use the same class.
 * awaitable methods should be Async suffixed.
 * Avoid ambiguity on your methods. Doing `asyncFtpClient.DownloadBytes(url);` will pick up 2 ambiguous methods.
 * Have the `CancellationToken` on the last argument, not in the middle.
 * `out var` are usually at the end, not at the beginning.
 * Provide a parser to download quickly a full URL `ftp://username:password@host/file.png` without having to do the whole dance of parsing this `Uri` myself. This would align with `HttpClient` where you can do `httpClient.GetByteArrayAsync(requestUri)`
 * `Please call Connect() before trying to read the Capabilities!` I think you could auto connect for me.
 * `501 Message: Invalid number of parameters` I have to call the correct constructor based on my username and password situation. I shouldn't have to do that.

Something that should be 1 line ends up to be this:
```
var path = new Uri(url);
var userInfo = path.UserInfo.Split(':');
var username = userInfo.Length > 0 ? userInfo[0] : string.Empty;
var password = userInfo.Length > 1 ? userInfo[1] : string.Empty;
var ftpClient = new AsyncFtpClient(path.Host, path.Port);

// Need a different constructor call otherwise we get 501 from the server.
if (!string.IsNullOrEmpty(username) || !string.IsNullOrEmpty(password))
{
    ftpClient = new AsyncFtpClient(path.Host, username, password, path.Port);
}

await ftpClient.AutoConnect();
return await ftpClient.DownloadBytes(path.PathAndQuery, CancellationToken.None);
```

FanDjango:
> `Please call Connect() before trying to read the Capabilities!`  ??? => I think you could auto connect for me.

I would like to have more details on how this exception came to be triggered. I am curious as to the how you are causing this to happen.

> `501 Message: Invalid number of parameters`. I have to call the correct constructor based on my username and password situation. I shouldn't have to do that.

Here also, more details on how this happens would be helpful. This 501 was a response to what command?

[...]

PR #1321, opened 2023-09-02
Add new API function "DownloadUriBytes"

FanDjango:
After a rather strange discourse on issue #1320, from which one small idea stands out:

"It would be nice to have a (more or less) one-liner replacement for ...GetBytes(ftp_uri) as seen in .NET with HttpClient".

Apart from the fact that Microsoft is neglecting, deprecating or dropping most of its interfaces to ftp, this approach is not as flexible as having full control over the way an ftp server is accessed.

Nonetheless, here is a quick way to get something from an ftp url and let's add it to the API (the code to do this is simple, I wonder what the TO was ranting about):

**`DownloadUriBytes(...)`**

Sample code shows up the functionality and ease of use:

```cs
using (var client = new FtpClient()) {

	// Add config statements as needed, for example:
	client.Config.ValidateAnyCertificate = true;

	byte[] outBytes;

	// Just do the download, no need to connect, disconnect, it will be done for you.
	client.DownloadUriBytes(out outBytes, "ftps://myuser:mypass@myftpserver.com:21/var/log/test1.log");

}
```

### DIFF
--- a/FluentFTP/Client/AsyncClient/DownloadUriBytes.cs
+++ b/FluentFTP/Client/AsyncClient/DownloadUriBytes.cs
@@ -1,0 +1,51 @@
+﻿using System;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+using FluentFTP.Helpers;
+
+namespace FluentFTP {
+	public partial class AsyncFtpClient {
+
+		/// <summary>
+		/// Downloads the specified uri and return the raw byte array.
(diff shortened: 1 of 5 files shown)


robinrodricks:
Very interesting. I will review this and think on it. Thanks!

robinrodricks:
Documented

https://github.com/robinrodricks/FluentFTP/wiki/File-Transfer

PR #1322, opened 2023-09-02
Add optional CWD wrapping for GetListing commands

FanDjango:
and make the wrapping optional as a global config option.


robinrodricks:
Oh wow, this has been wanted for some time. But never implemented. Looks great from my end. Approved!

PR #1331, opened 2023-09-11
Some fixups to review

FanDjango:


robinrodricks:
In Auto mode, can we disable the user from calling `SetWorkingDirectory`? We can throw an exception.

Like inside the `SetWorkingDirectory`, we do:

```
if (Config.Navigate == Auto){
   throw new InvalidOperationException("You cannot manually set the working directory in Automatic navigation mode. Instead just use absolute filepaths in your FTP API commands and the system will do it internally.");
}
```

FanDjango:
IMO there is no need to disable CWD because it also works fine if you give a relative path (in both auto modes) mit freundlichen Grüßen,Michael Stiemke ***@***.***)AnSyNova GmbHRingstr. 1385405 Nandlstadt‪+49 8756 969 49 79‬<http://www.ansynova.com>BTC: 1AnSyNovappA6rEgGKpFcmGe3pTyXzgccS--------------------------------------------------------Vertretungsberechtigter Geschäftsführer:Dipl. Inf. (TU) Michael StiemkeRegistergericht: Amtsgericht MünchenRegisternummer: HRB 103017Umsatzsteuer-Identifikationsnummer gemäß § 27aUmsatzsteuergesetz: DE ‪160163922‬ -------- Ursprüngliche Nachricht --------Von: Robin Rodricks ***@***.***>Datum: Di., 12. Sept. 2023, 06:33An: robinrodricks/FluentFTP ***@***.***>Cc: FanDjango ***@***.***>, Author ***@***.***>Betreff: Re: [robinrodricks/FluentFTP] Some fixups to review (PR #1331)
In Auto mode, can we disable the user from calling SetWorkingDirectory? We can throw an exception.
Like inside the SetWorkingDirectory, we do:
if (Config.Navigate == Auto){
   throw new InvalidOperationException("You cannot manually set the working directory in Automatic navigation mode. Instead just use absolute filepaths in your FTP API commands and the system will do it internally.");
}

FanDjango:
Also, IMO there is no need to disable CWD because it does not hurt if the user uses it (he might need it for other purposes). I don't understand where this fixation on disabling CWD comes from, do I misunderstand something? mit freundlichen Grüßen,Michael Stiemke ***@***.***)AnSyNova GmbHRingstr. 1385405 Nandlstadt‪+49 8756 969 49 79‬<http://www.ansynova.com>BTC: 1AnSyNovappA6rEgGKpFcmGe3pTyXzgccS--------------------------------------------------------Vertretungsberechtigter Geschäftsführer:Dipl. Inf. (TU) Michael StiemkeRegistergericht: Amtsgericht MünchenRegisternummer: HRB 103017Umsatzsteuer-Identifikationsnummer gemäß § 27aUmsatzsteuergesetz: DE ‪160163922‬ -------- Ursprüngliche Nachricht --------Von: Robin Rodricks ***@***.***>Datum: Di., 12. Sept. 2023, 06:33An: robinrodricks/FluentFTP ***@***.***>Cc: FanDjango ***@***.***>, Author ***@***.***>Betreff: Re: [robinrodricks/FluentFTP] Some fixups to review (PR #1331)
In Auto mode, can we disable the user from calling SetWorkingDirectory? We can throw an exception.
Like inside the SetWorkingDirectory, we do:
if (Config.Navigate == Auto){
   throw new InvalidOperationException("You cannot manually set the working directory in Automatic navigation mode. Instead just use absolute filepaths in your FTP API commands and the system will do it internally.");
}

robinrodricks:
>  I don't understand where this fixation on disabling CWD comes from, do I misunderstand something?

I have long hated this way of working:
```
CWD /my/folder
DownloadFile text.txt
CWD /sub/folder
DownloadFile video.mp4
```

Instead I have preferred this from day 1:
```
DownloadFile /my/foldertext.txt
DownloadFile /sub/foldervideo.mp4
```

As the system is aware of exactly what path you want, and can correctly navigate the directories automatically. Completely abstracting that detail away from users. So in Auto mode I would prefer to disallow manual CWD, as it encourages users to use absolute paths and trust FluentFTP to handle the dir navigation.

My fixation is as I want to move users away from manual path navigation which I consider very "command-line" style.

FanDjango:
Though I discover some holes in your logic, I'm certain I'll understand better after cosidering your standpoint further.

FanDjango:
Now now. 

The original intent was to make recalcitrant servers that do not support blanks in pathnames actually work. 

The method that does this is, to CWD to the needed directory, thereby avoiding having to enter the blanks.

So, such servers **need** the, let's call it, **CWD_then_CMD**. They don't work with any other method.

Servers that don't need this method, will work if you use this method. They will also work with the, let's call this one the, **FULL_PATH_in_CMD** method. In effect, they will work with **both** methods.

Per default, FluentFTP does not impose usage of any "method" on the user and lets him decide. This will be the majority of the users.

The (very) few users that **need** the special method(s) don't really need any restrictions on what they code, they just want it to work using a CWD-trick that they could easily handle themselves. These users are ones who would prefer to use the absolute path and would like the shortcut of having FLuentFTP do the CWD for them.

> I have long hated this way of working

Well, that's fine for you. Others might just love the opposite. But please understand: Those that need "AUTO" mode are **already** those users that use absolute paths (with possible blanks). No need to artificially restrict something that won't be used anyway. Those users that use the **hated (by you)** style won't need to use auto anyway.

> My fixation exists because I want to move users away from manual path navigation which I consider very "command-line" style.

In that case, you should restrict the use of manual path navigation for the **normal** user, and not just for the case of a problem fixup (because that's what it is) that will only be used by very few users.

What you originally got: A fix for specific problem from me ("WRAP_CMDS_WITH_CWD") 
What you then wanted: Some brobdignagian marketing type of super-feature ("AUTO_NAVIGATION")

Sounds sexy, but **it does not hit the mark**.

robinrodricks:
Let it be for now. I have no intention of such a lengthy argument.

As long as you can implement the Auto and SemiAuto and Manual modes as we discussed previously, I will for now withdraw my views on imposing restrictions on manual CWD.

PR #1342, opened 2023-09-19
GetReply, if used after OpenRead, should turn off ignoreStaleData

FanDjango:
As a final final touch, if a user executes `GetReply(...)` himself after using `OpenRead(...)`, we can turn off the boolen `Status.IgnoreStaleData`, as a reply will not be forthcoming.


robinrodricks:
Released.

https://www.nuget.org/packages/FluentFTP/48.0.0

PR #1369, opened 2023-11-10
Timeouts during auto detection

FabBadDog:
I added an option to treat timeouts during the autodetection as a failed detection attempt, instead of aborting with an exception.

# Adjustments

- I added a boolean property `AbortOnTimeout` to the `FtpAutoDetectConfig` which is `true` by default.
- I adjusted the `ConnectModule` to honor the flag and treat TimeoutExceptions accordingly.

# Background

I use FluentFTP in a .NET 4.6.2 project and I had an issue connecting to a FileZilla server with FTPS enabled.
While my code works in .NET 5 and higher, I get a Timeout error with 4.6.2.
Investigating this, I determined that the Socket internally is getting the error 997 (IO PENDING), which should not necessarily be an aborting error, but is not handled my the .NET Framework. 

FluentFTP is throwing a `TimeoutException` whenever the Socket is not connected after calling .Connect(), no matter what the real cause was.

# Alternative Ideas
I do not understand why `TimeoutExceptions` should ever be treated as aborting errors during AutoDetect. In my understanding a timeout means that the configuration which has been tried did not work, but not that no configuration can work.

So I think a better solution would be to just remove the TimeoutException from the permanent errors, but I did not want to change the current behavior so drastically without knowing enough.

### DIFF
--- a/FluentFTP/Client/Modules/ConnectModule.cs
+++ b/FluentFTP/Client/Modules/ConnectModule.cs
@@ -115,7 +115,7 @@ namespace FluentFTP.Client.Modules {
 						}
 
 						// check if permanent failures and hard abort
-						var permaEx = IsPermanentConnectionFailure(ex);
+						var permaEx = IsPermanentConnectionFailure(ex, config.AbortOnTimeout);
 						if (permaEx != null) {
 							if (config.CloneConnection) {
 								conn.Dispose();
@@ -209,7 +209,7 @@ namespace FluentFTP.Client.Modules {
 
 				// try each SSL protocol
 				foreach (var protocol in config.ProtocolPriority) {
-					// Only check the first combination for FtpEncryptionMode.None 
+					// Only check the first combination for FtpEncryptionMode.None
 					if (encryption == FtpEncryptionMode.None && config.ProtocolPriority.IndexOf(protocol) > 0) {
 						continue;
 					}
@@ -250,7 +250,7 @@ namespace FluentFTP.Client.Modules {
 						}
 
 						// check if permanent failures and hard abort
-						var permaEx = IsPermanentConnectionFailure(ex);
+						var permaEx = IsPermanentConnectionFailure(ex, config.AbortOnTimeout);
 						if (permaEx != null) {
 							if (config.CloneConnection) {
 								conn.Dispose();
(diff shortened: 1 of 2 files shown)


robinrodricks:
My understanding when I first wrote this, was that a timeout only occurs when the host/port is unreachable. Meaning the connection details itself are incorrect. And so we don't bother to waste any more time trying different protocols since the host/port refuses to be reached, ie. cannot be reached, and so it makes sense to hard abort.

On the contrary, can you think of any situation where a timeout occurs, but the host/port are correct?

robinrodricks:
Also is this specific to a certain .NET platform? Then we can automatically enable your fix for that platform and do away with the extra config.

FanDjango:
If this is happening from async code, WSA_IO_PENDING should not "be turned into a FluentFTP" timeout exception" immediately. It should rather cause a wait for timeout seconds - as one would expect the async operation to actually complete (or not) and if this wait times out, only then would a timeout exception be issued. With such additional code, I surmise that your auto-connect would actually time-out as with .NET 5.

The other half of the discussion is as @robinrodricks stated. I dimly remember some issues in the past, where this logic was introduced because users were complaining about long wait times on auto-connect due to repeatedly timing out on the same host:port endpoint.

robinrodricks:
> I dimly remember some issues in the past, where this logic was introduced because users were complaining about long wait times on auto-connect due to repeatedly timing out on the same host:port endpoint.

Yes something like this came up. 

I found the commit where this "hard abort" was first made. https://github.com/robinrodricks/FluentFTP/commit/fdf9e55aaa1fd28bd0c256888b774c2b30d0d667

If anyone has time/energy you can look through it and see why it was done.

But I still agree with my original reasoning: _**"a timeout only occurs when the host/port is unreachable. Meaning the connection details itself are incorrect"**_

robinrodricks:
If this fix is wonderful for you and occurs on newer platforms AS WELL, then I don't mind merging it asap and releasing. But if its only for older platforms, we can consider making it the default behavior. Or add this flag only for older platforms. I'm not sure how to proceed here.

FabBadDog:
I think we might find a better solution which does not need a new setting. I think that would be preferable.I need to do some tests and get back to you on Sunday.On Nov 10, 2023 20:40, Robin Rodricks ***@***.***> wrote:
If this fix is wonderful for you and occurs on newer platforms AS WELL, then I don't mind merging it asap and releasing. But if its only for older platforms, we can consider making it the default behavior. Or add this flag only for older platforms. I'm not sure how to proceed here.

FanDjango:
**Connection Exceptions** (i.e. SYN ACK SYN/ACK failures)

> a timeout only occurs when the host/port is unreachable. Meaning the connection details itself are incorrect...

Unclear statement: It is reachable but there is no answer.

and for the sake of completeness:

...or the server is overloaded and does not answer or the network connection to the server is down.

DNS and routing failures produce different exceptions.

robinrodricks:
> But if you connect to host:21 and get timeout, why abort when maybe host:990 and SSL will work?

In this case a setting is probably better as users can configure it anyhow they want.

robinrodricks:
A tiny setting aint gonna hurt anyone. Lets accept it now.

Issue #1371, opened 2023-11-14
UploadFile returns Success but file is only partially uploaded


KLuuKer:
**FTP Server OS:** Synology

**FTP Server Type:** Synology 7.1.x

**Client Computer OS:** Windows

**FluentFTP Version:** 34.0.2 ~ 48.0.3

**Framework:** .NET core 2.2 ~ 8

Maybe #1358 has a similar issue like mine but just doesn't know it.

So i basically have always uploaded to a very remote server with bad connection circumstances.
A couple of weeks ago i'm using wifi (yes i know) to upload files to the ftp server and am suffering even more upload issues.

And today i finally noticed the `UploadFile` method returning success while the upload wasn't fully uploaded yet.
But if i run `GetFileSize` it's able to correctly return the correct size from the remote file.
So as a work around i have added a extra check after i'm done uploading a file to check the size,
and to resume the upload until it's completed.

**Logs :**

```

<paste VERBOSE logs here>

```

**Code:**
```
// here i create a AsyncFtpClient, no weird configs\overrides fyi

for (var i = 1; i < 99; i++)
{
    try
    {
        if (!ftpclient.IsConnected || !ftpclient.IsAuthenticated)
        {
            await ftpclient.Connect(uploadProfile);
        }

        var result = await ftpclient.UploadFile(
            archivePath,
            $"/{DateTime.Now:yyyy-MM-dd}/" + archiveFileName,
            FtpRemoteExists.Resume,
            true,
            FtpVerify.Retry,
            new Progress<FtpProgress>(p => progress.Tick((int)p.Progress, p.ETA, $"{p.TransferSpeedToString()} uploading to ftp... attempt: {i}")));

        //// i have some code here now that double checks the filesize and prevents it from breaking the loop.

        if (result == FtpStatus.Success)
        {
            break;
        }
    }
    catch (Exception ex)
    {
       //// some logging code etc, also just restarts the loop
    }

    if (ftpclient.IsConnected)
    {
        await ftpclient.Disconnect();
    }
}
```

FanDjango:
> And today i finally noticed the UploadFile method returning success while the upload wasn't fully uploaded yet.
But if i run GetFileSize it's able to correctly return the correct size from the remote file.
So as a work around i have added a extra check after i'm done uploading a file to check the size,

Success, but not fully uploaded? Need to see a verbose log of that happening. In the log, I can see the file size, the number of bytes uploaded and the completion message.

[...]

Issue #1367, opened 2023-10-31
When there are many files in the server folder,upfiles very slow. 


luy710:
FileExists->GetNameListing-> OpenDataStream("NLST " + path, 0L); When there are many files in the server folder, it significantly affects the upload speed. Is there any consideration for optimization?

FanDjango:
Make sure that MSLD is used. Check if FEAT shows MLSD is supported. Show logs of why NLST is being used for FileExists instead of NLST.

Logs!

FanDjango:
We merged a fix for this into master. Can you give some feedback?

Issue #1358, opened 2023-10-26
UploadFile seems brocken with version 48.0.1


giswa:
**FTP Server OS:** Unix 
**FTP Server Type:** ProFTPD 
**Client Computer OS:** Windows
**FluentFTP Version:** 48.0.1
**Framework:** .NET 4.8

Using Fluent FTP 48.0.1, I get the error "450 Path does not exists" from server when uploading a single file  FtpClient.UploadFile().
The same command is working on version 33.0.3

I'll try to get some more detailed log asap.

Cheers,
Giswa

FanDjango:
> The same command is working on version 33.0.3

I suppose the **exact** command is secret, which works in 33.0.3 (ancient...) but **not** in 48.0.1.

> I'll try to get some more detailed log asap.

Very good. Until then, other users please remain calm when reading this.

giswa:
the exact command is : 

`client.UploadFile(fullFileName, "/" + remoteDirectory + "/" + Path.GetFileName(fullFileName), FtpRemoteExists.Overwrite, True) ; `
which returns : Code: 450 Message: /TestFolder: No such file or directory
full log : 
```
Warning:  Reconnect needed due to disconnected control connection
Status:   Command stashed: PWD

# Connect(True)
Warning:  Cannot reconnect, reverting to full connect
Status:   FluentFTP 48.0.1.0(.NET Framework 4.7.2)
Status:   Connecting to IP #1= ***:990
Status:   Waiting for a response
Response: 220 ***.**.***.*** FTP server ready [738818.691d]
Command:  AUTH TLS
Status:   Waiting for response to: AUTH TLS
Response: 234 AUTH TLS successful [5ms]
Status:   FTPS authentication successful, lib = .NET SslStream, cipher suite = Tls12 (Aes256, 44550, 256) [49ms]
Command:  USER ***
Status:   Waiting for response to: USER ***
Response: 331 Anonymous login ok, send your complete email address as your password [4ms]
Command:  PASS ***
Status:   Waiting for response to: PASS ***
Response: 230 #### Welcome to *** FTPS *** #### [11ms]
Command:  PBSZ 0
Status:   Waiting for response to: PBSZ 0
Response: 200 PBSZ 0 successful [5ms]
Command:  PROT P
Status:   Waiting for response to: PROT P
Response: 200 Protection set to Private [4ms]
Command:  FEAT
Status:   Waiting for response to: FEAT
Response: 211-Features:
Response: PBSZ
[...]

PR #1379, opened 2023-11-22
Fix for 1371, 1358

FanDjango:
Finally managed to debug why in case there **is no server handler** assigned or detected, `FileExists(...)` will ignore the SIZE FEAT and use NLST instead.


FanDjango:
Fix for #1371, #1358, and possibly for #1367

PR #1382, opened 2023-11-24
Correctly set supported method in socks5 negotiation

rmja:
This now matches the sync negotiation


FanDjango:
Thanks for catching this. Another case of Sync-Async-non-sync (joke).

PR #1387, opened 2023-11-30
Remove unnecessary awaits

bannikovilea:
There were several places in the asyncclient where the extra await could be omitted.

If they are removed, we will wait where necessary.

If there was a plan to do this, tell me what the profit was?


robinrodricks:
Can you see why the entire file is showing up as changed and fix it? Only the changed lines should show changed.

robinrodricks:
Amazing, so you have made methods designed to be async, into non-async methods. What good is this for anyone?

And everyone must update their app code to call these now non-async methods, for what purpose exactly?

This seems to be an exercise in futility.

bannikovilea:
> Amazing, so you have made methods designed to be async, into non-async methods. What good is this for anyone?

Why did they become non-asynchronous? In this case, Task is already returning. But in my case, no additional state will be created for the state machine, since the task will return without a wrapper

robinrodricks:
I'm sorry I cannot accept this, because as per our design principles this is the AsyncFtpClient class, where all major FTP API methods are async. If you want sync operation, use the sync class.

bannikovilea:
Sorry, but I don't understand. In most cases, when an async method calls another and there’s no chain (e.g. public method calling a private method directly) return Task is fine, and avoids the overhead of the await state machine for that method. 

The method still remains asynchronous, since the end user will have to  await for the result in any case. Accordingly, why make extra calls? 

A thousand users will not need to change the code. Since they will await it in any way. In my approach, we will not create unnecessary states for the state machine and the operation of the algorithm will be more efficient and optimal.

Can you send me an article to read about why asynchrony didn’t work here?

bannikovilea:
> Exception propagation?

This should work. If the result of a method is awaited, then since the exception will be generated in the underlying methods, it will be successfully propagated to the top.

Tell me if I'm wrong, why it won't work?)

bannikovilea:
Problems with exception propagation can occur when using async void. It should work in our scenarios

FanDjango:
Dispose processing?

FanDjango:
I don't know.

FanDjango:
As far as I know, overloads (that do nothing other than wrapping the actual (async) method) can just return tasks to be awaited and need not await the base method. If the base method itself is part of an API, you can't change its signature. Solve that by moving the base out to a "common" method and make the original base API call just another overload.

So, yes, I think you could rewrite these overloads in such a way as to keep the API valid and unchanged and eliminate the superfluous await.

robinrodricks:
> A thousand users will not need to change the code. Since they will await it in any way. In my approach, we will not create unnecessary states for the state machine and the operation of the algorithm will be more efficient and optimal.

Hmmm so you are saying that the library users will have ZERO IMPACT?

FanDjango:
Trouble is (or was):



> Can you see why the entire file is showing up as changed and fix it? Only the changed lines should show changed.
> 
> image

You never responded to that. Robin might have seen what you are up to.

Look at it this way (and I think, starting with only one example might be good):

image

And experiment with that, first, in order to clear this discussion up a little bit.

FanDjango:
I don't have time to check this, as I am busy with something else right now. And since it is not of utmost priority I think this is a design discussion and not a problem issue.

FanDjango:
> Good question, next question?

I'm sorry, that answer was flippant. I apologize.

> What about the fact that the methods are not marked async, will the IDE behave in the same way?

Our unit tests (for the example shown, AutoDetect in the ASYNC client) compile unchanged.

FanDjango:
If this is to become a design discussion (which I advocate), I would like to add the possible redesign of using ValueTask (in most cases) instead Task in those places where we internally know that this pattern would be sufficient.

bannikovilea:
The async keyword is only used when await is necessary. Guys, have you ever wondered why interfaces don't use the async keyword? It allows you to use the await operator inside a method by assembling a state machine.
image

So in the below code we will also get a warning in the IDE about a pending wait in the method.
```
public class GodKnow
    {
        public async Task TheLupa()
        {
            // someSyncCode
            await Task.Delay(1);
        }

        public Task ThePupa()
        {
            return Task.Delay(1);
        } 
    }

    public class Test
    {
        public async Task MethodWithLupa()
        {
            var godKnow = new GodKnow();

            godKnow.TheLupa();
        }

        public async Task MethodWithPupa()
        {
            var godKnow = new GodKnow();
            
            godKnow.ThePupa();
        }
    }
```
image


From the above, I would conclude that the method signature for millions of users will remain the same.

I think the ValueTask approach will also be useful. Maybe I'll try this approach, but I think you'll do better ;-) And maybe I'll come back later and remove a couple of unnecessary awaits xD.

bannikovilea:
Well, if you think that this method will execute asynchronously because you added the async keyword, you are wrong.
image

A modern IDE will tell you what the error is)
image

bannikovilea:
> Yes I am always worried to make major changes to critical functions that could impact everyone.

What about autotests with the most common use cases?

robinrodricks:
> Well, if you think that this method will execute asynchronously because you added the async keyword, you are wrong.

This I understand.

> What about autotests with the most common use cases?

https://github.com/robinrodricks/FluentFTP/wiki/Automated-Testing

> From the above, I would conclude that the method signature for millions of users will remain the same.

ChatGPT mirrors my concerns:

image

bannikovilea:
I also asked a question to the chat gpt.
image

image

robinrodricks:
**Point 1 of your ChatGPT response contradicts DownloadFile's code.**

As you can see, this NEEDS an await operator. You have removed it and are now saying it "does not have await operators". How does that logic work?? **DownloadFileToFileAsync** is a full async method. It is not a sync method. You NEED await operators for it.

-------

image

bannikovilea:
>As you can see, this NEEDS an await operator.

Why is an operator needed? There are no blocking calls here. There is processing of input data and throwing out an exception. The rest of the blocking operations are in the internal method and should be executed normally. All asynchronous calls will be in the internal method; they are not in the method you provided.

bannikovilea:
I was wrong. The call stack will indeed change. Found a great [video that explains ](https://www.youtube.com/watch?v=Q2zDatDVnO0)

Useful documentation can be found [here](https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#prefer-asyncawait-over-directly-returning-task) 

In the current implementation, the request can be closed.

But returning a task instead of waiting will still be a more efficient solution in terms of performance and resources.

FanDjango:
Aw, schucks. I was planning to actually test this end of the week. But now...

PR #1389, opened 2023-12-11
Update FtpSocketStream.cs to fix apparent typo "#if NET50_OR_LATER" to "#if NET5_0_OR_GREATER"

sean-bloch:
Appears to be a typo. Updating the directive to "NET5_0_OR_GREATER", matching other examples in the same file also "unhides" the code as expected in visual studio.


robinrodricks:
Thanks!

PR #1405, opened 2023-12-13
Docker Integration Tests: Build Dockers: Fix run.sh missing sources.list in py-slim debian

FanDjango:


FanDjango:
Dockerfile Build was failing because the supplier of py-slim broke his image.

PR #1407, opened 2023-12-14
Documentation PR for FluentFTP.csproj

FanDjango:
Time to document the current state of affairs in .csproj re TFMs.


robinrodricks:
Moved to here - https://github.com/robinrodricks/FluentFTP/wiki/Code-Standards

FanDjango:
I kind of thought too that we would move that to the Wiki. But this way they are also in the commit history. I'd like to remember without the Wiki the gist of what I/We changed at this time in case of problems ensuing out of this.

PR #1411, opened 2023-12-15
Update some selected NuGet packages

FanDjango:
We were a little "behind" in some cases. These work.

The packages used in the test suite cannot be updated without some small changes to the source - keeping those on the bavk burner for a rainy day.


FluentFTP.Logging/FluentFTP.Logging.csproj
```diff
@@ -39,7 +39,7 @@
         <!-- Reference the first version of FluentFTP that exposes IFluentLogger. Do not change this version. -->
         <PackageReference Include="FluentFTP" Version="43.0.1" />
         <!-- Reference the oldest version of MELA possible. Do not change this version. -->
-        <PackageReference Include="Microsoft.Extensions.Logging" Version="2.1.0" />
```
robinrodricks, line 42: Your changes are normally excellent, but in this case,  we do not want to change this version. See the comment above. There is no reason to update this imo.

FanDjango:
That one slipped by me, sorry. Good that you saw this.

PR #1420, opened 2023-12-18
"Demonize" NOOP handling

FanDjango:
### Initial changes

1. Added a `daemon.cs` which is started as a task from `client.connect() sync&async` when successful.
```cs
			if (!Status.DaemonRunning) {
				m_task = Task.Run(() => { Daemon(); });
			}
```
2. Added a termination for the daemon in the dispose of `FtpSocketStream` (control connection)
```cs
			if (Client.Status.DaemonRunning) {

				if (this.IsControlConnection) {
					// if we are terminating the control connection, stop the NOOP daemon
					DateTime tn = DateTime.Now;
					do {
						((IInternalFtpClient)Client).LogStatus(FtpTraceLevel.Verbose, "Waiting for daemon termination");
						Thread.Sleep(150);
						if (!Client.Status.DaemonRunning) {
							break;
						}
						Thread.Sleep(850);
					} while (DateTime.Compare(DateTime.Now, tn.AddMilliseconds(5000)) <= 0);
				}
				else {
					// if terminating a data connection, tell the NOOP daemon to revert to GetReply usage
					Client.Status.DaemonGetReply = true;
				}
			}
```
3. Daemon knows an active connection is transferring data via the stream setting a status and thus behaves differently than when a normal idle is taking place. The `connect()` in `FtpSocketStream` is the place to look:
```cs
			// the NOOP daemon needs to know this
			if (!IsControlConnection) {
				Client.Status.DaemonGetReply = false;
			}
```

4. Added a capability to automatically prefex all commands with an execution of a NOOP command to make detection of connectivity issues easier.

This (more or less) would have been all, apart from the trivial code of the daemon itself.

### Locking changes

But it needs to be prevented that the daemon will `stream.WriteLine(...)` in the middle of some other such `stream.WriteLine(...)` taking place, so the entire FluentFTP locking scheme needed to be looked at.

Can of worms...

There was locking in place via a Lock Object m_lock. This was placed in all sorts of code points and not really well maintained any more. **And it was not present at all in ASYNC (actually because lock() doesn't work in async)**. Actually it was **unneeded** in the first place. Multithreading needs a client object **and** a separate connection for each instance, so what for the locking?

So I removed it completely. This touches many .cs files in SYNC client, sorry, but the changes are minimal. But because removing a `lock( ) { }` block changes indentation over many lines, the diffs look really awful.

**Instead** I introduced a SemaphoreSlim to be used around all `stream.WriteLine(...)` and `Getreply(...)` occurences to lock these against the daemon interfering . **This works in SYNC and in ASYNC**.

It runs all the integration tests and my local ones.

He who wants to pre-Test this: It is in https://github.com/FanDjango/FluentFTP/tree/DAEMON


MisterSausage:
I had yesterday's code running for 12 hours overnight, and it worked very well. Log attached. The test looped through performing a couple of operations on the server, sleeping for a random time of up to two hours, and then trying to recover if the connection was lost. Everything went according to plan.

With the difficult Pure-FTPd server I have access to, the background no-op commands prevented the firewall or whatever it is from severing the connection successfully. However, even though the server advertises that it'll time out connections after being idle for 30 minutes, I found that it forcibly times out and disconnects clients that have done no actual transfers after about 60 minutes - an example of this can be seen in the log at 20:46:45.

There's probably no reasonable way around this short of actually downloading a file in the background, which would be silly - so in other words there's no way for the no-op daemon to keep a connection alive forever. I think this is more than good enough, because it'd be antisocial to push it much further, and certainly with the `IsStillConnected()` method there's now a reliable way to detect connection loss and recover reliably, while reusing the original `FtpClient` object.
[FTP Idle Test Log.txt](https://github.com/robinrodricks/FluentFTP/files/13708672/FTP.Idle.Test.Log.txt)

FanDjango:
> it'd be antisocial to push it much further

Full Ack.

It's more about servers that would shut down a control connection (and thus abort) during a very long running file transfer .

The rest is incidental.

So actually, the previous **non-demonized** NOOP mechanism would have sufficed, together with the `IsStillConnected(...)` API.

Therefore it is up to @robinrodricks wether he wants the daemon or not. I kinda like it - and I found a number of issues with locking etc. that we should not revert. But it is more complex than the previous solution...

My opinion: I would go with the daemon. It would allow us as a next step to count NOOPs sent, and improve the GetReply processing that devours NOOP responses with a timeout by devouring via count them down.

Big thanks for your testing efforts - very very valuable.

FanDjango:
Oof, that took a while:
````
2023-12-18T20:47:00	VERBOSE	Status: Closing/Disposing FtpSocketStream(control connection)
2023-12-18T20:47:00	VERBOSE	Status: Waiting for daemon termination
2023-12-18T20:47:01	VERBOSE	Status: Waiting for daemon termination
2023-12-18T20:47:02	VERBOSE	Status: Waiting for daemon termination
2023-12-18T20:47:03	VERBOSE	Status: Waiting for daemon termination
2023-12-18T20:47:04	VERBOSE	Status: Waiting for daemon termination
2023-12-18T20:47:05	VERBOSE	Status: Daemon terminated
````
Should be near instantaneous. I will inverstigate.

MisterSausage:
I hope @robinrodricks does want the daemonised no-op mechanism, because it's very powerful in combination with `IsStillConnected()`. In my case, as you said, doing without it would have sufficed, but it's better _with_ it, because I was seeing those weird broken connections any time from about two minutes of idle, and the newly-generated background traffic put a stop to that entirely.

Thank you again for all your hard work!

FanDjango:
Very much still work in progress... please be patient.

FanDjango:
There is going to be on more close - and re-PR - to clean this up. I hate rebasing beyond merges and force-pushes. It is just easier to redo.

Currently cleaning up code and testing.

This should be finished on Friday or so.

MisterSausage:
Let me know if you want a second pair of eyes to do any further testing.

FanDjango:
I've just now debugged the Semaphore.Slim handling and during the course of this of course had to modify some internal methods.

Tomorrow I will be ready, and I will make the new PR, which will then be cleaner, and we can discuss each and every change. Looking forward to your input. Thanks.

PR #1426, opened 2023-12-21
Daemon, new GetReply, remove m_lock

FanDjango:
### Initial changes

1. Added a `daemon.cs` which is started as a task from `client.connect() sync&async` when successful.
```cs
			if (!Status.DaemonRunning) {
				m_task = Task.Run(() => { Daemon(); });
			}
```
2. Added a termination for the daemon in the dispose of `FtpSocketStream` (control connection)
```cs
			if (Client.Status.DaemonRunning) {

				if (this.IsControlConnection) {
					// if we are terminating the control connection, stop the NOOP daemon
					DateTime tn = DateTime.Now;
					do {
						((IInternalFtpClient)Client).LogStatus(FtpTraceLevel.Verbose, "Waiting for daemon termination");
						Thread.Sleep(150);
						if (!Client.Status.DaemonRunning) {
							break;
						}
						Thread.Sleep(850);
					} while (DateTime.Compare(DateTime.Now, tn.AddMilliseconds(5000)) <= 0);
				}
				else {
					// if terminating a data connection, tell the NOOP daemon to revert to GetReply usage
					Client.Status.DaemonGetReply = true;
				}
			}
```
3. Daemon knows an active connection is transferring data via the stream setting a status and thus behaves differently than when a normal idle is taking place. The `connect()` in `FtpSocketStream` is the place to look:
```cs
			// the NOOP daemon needs to know this
			if (!IsControlConnection) {
				Client.Status.DaemonGetReply = false;
			}
```

4. Added a capability to automatically prefex all commands with an execution of a NOOP command to make detection of connectivity issues easier.

This (more or less) would have been all, apart from the trivial code of the daemon itself.

### New Configs

```cs
		/// <summary>
		/// Install the Noop Daemon. It is needed for all of the Noop functionality
		/// </summary>
		public bool NoopInstallDaemon { get; set; } = false;

		/// <summary>
		/// Gets or sets the length of time in milliseconds after last command
		/// (NOOP or other) that a NOOP command is sent./>.
		/// This is called during downloading/uploading and idle times. Setting this
		/// interval to 0 stops NOOPs from being issued.
		/// The default value is 0 - no NOOPs are issued.
		/// </summary>
		public int NoopInterval { get; set; } = 0;

		private List<string> _noopIdleCmds = new List<string> { "NOOP", "PWD", "TYPE I", "TYPE A" };

		/// <summary>
		/// These commands are to be used when the dataconnection is not active, i.e. no transfer
		/// is taking place. Currently allowed: NOOP, PWD, TYPE I, TYPE A
		/// </summary>
		public List<string> NoopIdleCmds {
			get => _noopIdleCmds;
			set => _noopIdleCmds = value;
		}

		private List<string> _noopXferCmds = new List<string> { "NOOP" };

		/// <summary>
		/// These commands are to be used when the dataconnection is active, i.e. a transfer
		/// is taking place. Currently allowed: NOOP
		/// </summary>
		public List<string> NoopXferCmds {
			get => _noopXferCmds;
			set => _noopXferCmds = value;
		}

		/// <summary>
		/// Issue a NOOP command tp precede any command issued on the control connection
		/// to test connectivity in a reliable fashion. Note: This can incur some control
		/// connection overhead and does not alleviate inactivity timeouts, it just helps
		/// to identify connectivity issues early on.
		/// </summary>
		public bool NoopAddNoopCmd { get; set; } = false;
```

### Locking changes

But it needs to be prevented that the daemon will `stream.WriteLine(...)` in the middle of some other such `stream.WriteLine(...)` taking place, so the entire FluentFTP locking scheme needed to be looked at.

Can of worms...

There was locking in place via a Lock Object m_lock. This was placed in all sorts of code points and not really well maintained any more. **And it was not present at all in ASYNC (actually because lock() doesn't work in async)**. Actually it was **unneeded** in the first place. Multithreading needs a client object **and** a separate connection for each instance, so what for the locking?

So I removed it completely. This touches many .cs files in SYNC client, sorry, but the changes are minimal. But because removing a `lock( ) { }` block changes indentation over many lines, the diffs look really awful.

**Instead** I introduced a SemaphoreSlim to be used around all `stream.WriteLine(...)` and `Getreply(...)` occurences to lock these against the daemon interfering . **This works in SYNC and in ASYNC**.

It runs all the integration tests and my local ones.

He who wants to pre-Test this: It is in https://github.com/FanDjango/FluentFTP/tree/DEMON


FanDjango:
I am having trouble synchronizing work / rebase and merges from multiple workstations, I keep messing up the PR. Freezing other work for now until the PR gets merged.

PR #1431, opened 2023-12-28
Give NOOP daemon more power

FanDjango:
The NOOP daemon was too hesitant in some cases where dropped connections produce horrendous exceptions. Simply ignoring these was not the right tactic. When the daemon discovers the problem, it is not enough to stop the timer loop - no, it is necessary to shut down the control connection instead of leaving it in a strange state.

### DIFF
--- a/FluentFTP/Client/BaseClient/Daemon.cs
+++ b/FluentFTP/Client/BaseClient/Daemon.cs
@@ -17,6 +17,8 @@ namespace FluentFTP.Client.BaseClient {
 			Status.DaemonEnable = true;
 			Status.DaemonAnyNoops = false;
 
+			bool gotEx = false;
+
 			do { // while(true)
 
 				if (m_stream == null || !m_stream.IsConnected) {
@@ -33,14 +35,26 @@ namespace FluentFTP.Client.BaseClient {
 						Config.NoopActiveCommands[rnd.Next(Config.NoopActiveCommands.Count)];
 
 					m_sema.Wait();
+
 					try {
+
 						// only log this if we have an active data connection
-						if (!Status.DaemonCmdMode) {
+						if (Status.DaemonCmdMode) {
+							Log(FtpTraceLevel.Verbose, "Command:  " + rndCmd + " (daemon)");
+						}
+						else {
 							((IInternalFtpClient)this).LogStatus(FtpTraceLevel.Verbose, "Sending " + rndCmd + " (daemon)");
 						}
 
 						// send the random NOOP command
-						m_stream.WriteLine(m_textEncoding, rndCmd);
+						try {
+							m_stream.WriteLine(m_textEncoding, rndCmd);
+						}
+						catch (Exception ex) {
+							((IInternalFtpClient)this).LogStatus(FtpTraceLevel.Verbose, "Got exception (#1): " + ex.Message + " (daemon)");
+							gotEx = true;
+							break;
+						}
 
 						LastCommandTimestamp = DateTime.UtcNow;
 
@@ -54,7 +68,9 @@ namespace FluentFTP.Client.BaseClient {
 								success = ((IInternalFtpClient)this).GetReplyInternal(rndCmd + " (daemon)", false, 10000, false).Success;
 							}
 							catch (Exception ex) {
-								((IInternalFtpClient)this).LogStatus(FtpTraceLevel.Verbose, "Got exception (#1): " + ex.Message + " (daemon)");
+								((IInternalFtpClient)this).LogStatus(FtpTraceLevel.Verbose, "Got exception (#2): " + ex.Message + " (daemon)");
+								gotEx = true;
+								break;
 							}
 
 							// in case one of these commands was successfully issued, make sure we store that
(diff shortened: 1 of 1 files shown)


MisterSausage:
Currently testing this (as of commit 97dae98), because I'm seeing what I believe is a bug in the public release 49.0.1.

Log:
```
*** Calling IsStillConnected() ***
2023-12-29T02:56:57	VERBOSE	Status: Testing connectivity using Socket.Poll()...
2023-12-29T02:56:57	VERBOSE	Command: NOOP
2023-12-29T02:56:57	VERBOSE	Status: Waiting for response to: NOOP
2023-12-29T02:56:57	INFO	Response: 200 Zzz... [150ms]
2023-12-29T02:56:57	VERBOSE	Status: Closing/Disposing FtpSocketStream(control connection)
*** Closed control connection even though NOOP response was OK ***
*** IsConnected = True, IsAuthenticated = True, even though control connection apparently closed ***
2023-12-29T02:56:57	VERBOSE	> GetModifiedTime("<file>")
2023-12-29T02:56:57	INFO	Command: MDTM <file>
2023-12-29T02:56:57	VERBOSE	Status: Waiting for response to: MDTM <file>
2023-12-29T02:56:57	VERBOSE	Status: Waiting for daemon termination
2023-12-29T02:56:58	VERBOSE	Status: Waiting for daemon termination
2023-12-29T02:57:00	VERBOSE	Status: Waiting for daemon termination
2023-12-29T02:57:01	VERBOSE	Status: Waiting for daemon termination
2023-12-29T02:57:02	VERBOSE	Status: Waiting for daemon termination
2023-12-29T02:57:03	VERBOSE	Status: Daemon terminated
*** Seems to struggle to terminate daemon ***
```

This is an intermittent issue - it only happens occasionally, and unpredictably.

1. When calling `IsStillConnected()`, the control connection is closed even though the server responds to the NOOP properly. Presumably there's an exception being thrown that's not being logged or handled correctly.
2. When the control connection is closed, `IsConnected` and `IsAuthenticated` remain `true`.
3. The follow-up command (`GetModifiedTime()`) then fails ("Unable to read data from the transport connection: An established connection was aborted by the software in your host machine..").
4. It goes on to struggle to terminate the daemon.

I regret that I still haven't been able to trawl through the code, but my gut instinct is telling me there's a synchronisation issue at play here that's rearing its head while the daemon is running.

If I may suggest:
1. Closing the control connection should _always_ set `IsConnected` and `IsAuthenticated` to false, without fail.
2. Exceptions should always be logged, other than in rare situations where they're expected to be thrown.

I'll use the latest master, try to create a contrived test where the issue is likely to be triggered more frequently, leave it running for 24 hours or so, and follow up.

MisterSausage:
Sorry to drop this in here, but I don't want to have to create a pull request.

Just taken a look at Daemon.cs, and `m_sema.Release()` should be in a `finally` block. If any exceptions occur, the main loop is broken out of, but the semaphore is never released.

MisterSausage:
Didn't need to test for long. Confirmed that the problem occurs only when the no-op daemon is enabled, up to and including commit 97dae98, and including when the minor weirdness regarding the release of `m_sema` is resolved. Got to be a synchronisation problem!

FanDjango:
First of all, many thanks for testing this.

My setup here is a ProFTPd server, where I have configured a `TimeoutNoTransfer = 60` (LOL), which is even nicer to test with than a `TimeoutIdle`, which of course also works. It just shows up very nicely that NOOPs etc. won't keep a modern server connected. Plus: I don't have to wait long to see what happens.

ProFTPd is **very** well behaved. Other servers don't even send a 5xx message before terminating. Other scenarios (I try pulling the cable, even) are also interesting.

I am struggling to reconcile a few opposing problems now:

I see **significant** changes in the behaviour of how C# .NET reacts to server side connection drops between: .NET Framework and .NET 5+, **also** between Debug and Release (probaly due to optimization). Therefore testing and tweaking is turning into a bit of a bother.

> Just taken a look at Daemon.cs, and m_sema.Release() should be in a finally block. If any exceptions occur, the main loop is broken out of, but the semaphore is never released.

> Edit: It's released, but in an unnecessary call at the end of the method.

Yes, looks like a boo-boo. But it worked fine here in all stress tests of server side termination I could muster.

> Got to be a synchronisation problem!

Yes. Take this example: Connection drops. NOOP from the Daemon hangs there for 10 seconds before it fails. While it is hanging, your code decides to do some command and the subsequent actions overlap.

Concerning your log excerpt:

> *** Seems to struggle to terminate daemon ***

I noticed that too and have by now found the reason

Let me do some more work on this - hope you can then test it also.

FanDjango:
Ok, have a look now, please. I merged some code that should address your reported symptoms.

MisterSausage:
Fantastic! Thank you. I'll test it again - bear with me for 24 hours.

MisterSausage:
Test completed, and everything seems to be working well. Nice job!

FanDjango:
Thanks. Still adding stuff to handle edge cases, though.

FanDjango:
@MisterSausage 
@robinrodricks 

Ok, The last set of commits conclude the rest of what I feel is possible to do for this.

I picked up on a couple of other things that needed tidying up, but the main effort was provoking exceptions and synchronisation problems **over and over** again, with the help of a server that drops the connection after 10 seconds of inactivity (yes!) and a NoopInterval set to 1 second (again, yes!).

The only issue that **remains** is that sometime, when the connections are being disconnected, disposed, shut down or fail, at the end of a users program, visible exceptions might occur but are caught ok.

The problem is that right down at the core (FtpSocketStream) will blithely issue Write's to the socket level without catching exception at that low level, rather passing this further up the chain. **But:** since quite some time (in terms of versions) we have numerous code points entering that socket Write directly, async.

With what the design **today** would require, one would do a complete redesign of the low level read/write mechanism, perhaps with a priority queue or some such serialisation method. But that is beyond the scope of a seemingly simple mod to support a basic sort of NOOP support.

Regardless, turning off NOOP will disable all of this, turning it on with sensible timings (NoopInterval >> 15 seconds) is quite stable. Together with NoopTestConnectivity = True, you get real comfortable. Add judicious use of IsStillConnected(), you are covered.

Issue #1444, opened 2024-01-09
DownloadStream method ignores stopPosition


alexgubanow:
**FTP Server OS:**  - Windows

**FTP Server Type:** MOVEit transfer

**Client Computer OS:** Ubuntu / MacOS

**FluentFTP Version:** 49.0.1

**Framework:** .NET 7.0

DownloadStream method ignores stopPosition and keeps downloading. In same time GetFileSize() gives correct size. Same behaviour observed with progress callback and without one.
My goal is download first 5Mb of file, i can ignore rest.

TimLCondor:
I am trying this since yesterday in every possible way but as you said, it keeps downloading the whole file for me as well:
```
var bytes = await client.DownloadBytes(path, stopPosition: sizeof(long));
Trace.WriteLine("Peaked " + bytes.Length + " bytes"); // Prints "Peaked 7689435 bytes" after a while
```
And with OpenRead, to do this manually, I get a timeout exception.

The problem here is that I only want to read the file header.

alexgubanow:
spawning the PR right now, basically the stopPosition should match exactly the downloaded length, if we over downloaded - keeps till end. Im not sure my solution is expected fix for it, but it works.

PR #1445, opened 2024-01-10
#1444 break download loop if stopPosition reached

alexgubanow:
fixes the exit on stopPosition scenario in DownloadFileInternalAsync method
not sure if this is the right fix, as i had to set earlySuccess = true; otherwise i would get 550 error

### DIFF
--- a/FluentFTP/Client/AsyncClient/DownloadFileInternal.cs
+++ b/FluentFTP/Client/AsyncClient/DownloadFileInternal.cs
@@ -107,7 +107,7 @@ namespace FluentFTP {
 						long bytesProcessed = 0;
 
 						sw.Start();
-						while ((readBytes = await downStream.ReadAsync(buffer, 0, buffer.Length, token)) > 0) {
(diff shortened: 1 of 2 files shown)


alexgubanow:
oh right, on it

FanDjango (commented):
` && (offset < fileLen || readToEnd)`

Is that really needed? When you test, what happens if that part is left out? I thought that .Read can't be sent more than the fileLen by the server.

alexgubanow:
yes i have tested this :)
we have here two while loops, if you managed to download entire file without interruptions - you will exit inner loop in the end of file.
Hmm, if you are saying downStream should prevent us from reading more than fileLen - that would be the actual problem. 
Any ideas for a better fix?

FanDjango:
> yes i have tested this :)

Of course.

I see it now. You are right to leave the inner loop before it reads a potentially HUGE file to the end.

The outer loop needs to be terminated and earlySuccess is a good idea, since no success response from the server is to be expected.

FanDjango:
Someone might complain that "progress" will now report a wrong percentage, but duh. Better to get this to work at all.

FanDjango:
I'll merge this and we can take it onwards from that.

FanDjango:
Thanks for this.

Issue #1452, opened 2024-01-11
Multi-File operations should provide a overall progress


J0nathan550:
Would be great to have seperated function that will show the "Full-Progress" of `DownloadDirectory` and other functions that is related to downloading. 

This library has specific implementation of Progress. But this Progress only show optimized progress when it downloads something.

The other function should be the same with a little exception, it should always calculate what does it do, file checking and everything related to it.

Would be great to see it. 

Original Issue was:

Hello! I ran into issue where the progress of the `DownloadDirectory` doesn't show full progress, it's only showing the progress when you downloading file, but not when it is checking if file is good or not. Here is the footage. 

https://youtu.be/1JpgldcjjEA

FanDjango:
The `FtpProgress` class contains many of the necessary properties etc. to handle the "Overall Progress" for mutil file operations.

Taking DownloadDirectory as an example, what needs to be done to provide progress?

FanDjango:
Let's say the directory has 5 files, all of them nice and big.

You now implement a progress callback.

If you show just the "progress" it wil go from 0->n five times.
But the "FileIndex" property will increase 1, 2, 3, 4, 5.
So you if **you would know** the individual file sizes, you could calculate that "overall progress" in percent.

J0nathan550:
> The `FtpProgress` class contains many of the necessary properties etc. to handle the "Overall Progress" for mutil file operations.
> 
> Taking DownloadDirectory as an example, what needs to be done to provide progress?

When we provide the progress to function with `FullProgress` it should in loop fire the IProgress every single time when something happended to a file, for example we checked the file, and we like it, we skip it, and after skipping we add += 1, to file count. When we do so, update the progress imidiately. Which should give us effect of file checking.

FanDjango:
Whatever is done, it should all be stored in FtpProgress object. So the user can access the information and do with it what he likes.

J0nathan550:
> Whatever is done, it should all be stored in FtpProgress object. So the user can access the information and do with it what he likes.

Yes, would be just perfect either provide specific class that will collect all events no matter what, or implement functions that will fire all of the time when something happened, not only when file was downloaded.

FanDjango:
So not just bytes, files processed, but also "Downloading", "Checking", etc. etc. Need an enum of different actions.

J0nathan550:
> So not just bytes, files processed, but also "Downloading", "Checking", etc. etc. Need an enum of different actions.

[...]

PR #1453, opened 2024-01-12
Updated the Progress of downloading files.

J0nathan550:
Fixed issue in: #1452

In this PR I updated the progress updating of downloading files both in Sync Client and Async Client. 
I did not update the Upload Files, since I think that updating progress in there will be the same, and when you download files, there is a case where you have some files and you simply want to skip them. 
Problem of current progress is that, when you are downloading, it did not showed progress on "skipped" file, it just ignore it, and didn't fire any progress update.  

Now with this PR, it increases the count of files checked / need to download. 

Reason behind of this modification was, that when you want to show download progress to user, user can think that program doesn't do anything, when on background it does things. It is just a nice little thing to keep user notified in progress that something is happening with file. 

Current progress with PR: https://youtu.be/V-f3sCHDpjU
Video of previous bug without PR: https://youtu.be/1JpgldcjjEA

### DIFF
--- a/FluentFTP/Client/AsyncClient/DownloadFile.cs
+++ b/FluentFTP/Client/AsyncClient/DownloadFile.cs
@@ -66,7 +66,7 @@ namespace FluentFTP {
 
 			LogFunction(nameof(DownloadFile), new object[] { localPath, remotePath, existsMode, verifyOptions });
 
-
+			DateTime functionStarted = DateTime.Now;	
 			bool isAppend = false;
 
 			// skip downloading if the local file exists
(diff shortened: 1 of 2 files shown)


FanDjango:
I will merge this right away. I can see that calling `ReportProgress` will update the filecount information - this was being skipped in those "skip" cases :-))

This is a first step to a more comprehensive "progress reporting".

In the end, one will need to add more information to the progress object beyond what it contains today.

FanDjango:
Thanks for bringing this topic up and for this quick fix.

PR #1494, opened 2024-02-16
Call Dispose/DisposeAsync on BufferedStream instead of Flush/FlushAsync

kmgallahan:
Dispose and DisposeAsync call flush internally and execute additional cleanup logic.


FanDjango:
Thanks for this. Let's be testing the current master for a while at this stage of the async changes.

Issue #1508, opened 2024-03-19
MLSD Last Modified DateTime.Kind


cedmundsNull:
When getting a listing with MLSD, the last modified DateTime.Kind is Unspecified.

Should it be updated to UTC?

As far as I understand, MLSD should [always return UTC times](https://datatracker.ietf.org/doc/html/rfc3659#section-2.3).

Currently, I have implemented the following workaround:

```
var listing = await client.GetListing(path);

  if (client.Config.ListingParser == FluentFTP.FtpParser.Machine)
      foreach (var item in listing)
          item.Modified = DateTime.SpecifyKind(item.Modified, DateTimeKind.Utc);
```

FanDjango:
Good catch, I agree.

FanDjango:
There is a nice PR now addressing this issue, so as suggested by @cedmundsNull in #1509: closing

PR #1509, opened 2024-03-20
machine listing returns datetimekind.utc

cedmundsNull:


cedmundsNull:
close #1508

FanDjango:
Thanks for this!

I am wondering if this should be modified to be an optional thing. 

Perhaps many users are already doing conversions depending on their servers adhering to the RFCs or not, and the "unspecified" setting that FluentFTP is currently using.

Perhaps one should have a `client.Config.xxxx` that configures to use UTC or "unspecified" on MLST so as not to break any current users logic?

**on the other hand**

"Breaking" changes in order to handle protocol issues more correctly are not unknown in FluentFTP. Current users who have up to now been doing their own conversion would need to remove that (if any).

Good of you to handle the test scenarios as well.

I will merge this, and in the course of the next weeks we will see if there are any other effects to be considered.

Note: @robinrodricks Next release would be semantically a bump on API change?

PR #1513, opened 2024-03-21
Extend the built-in verification of all major file transfer methods with size comparison

DasRaschloch:
-- Added file size comparison logic to all high-level transfer methods
-- Changed the corresponding descriptions
-- Renamed OnlyChecksum to OnlyVerify due to misleading name


FanDjango:
Keeping OnlyChecksum will make current users happy. Thanks for this. I suppose someday someone will also argue for a datetime comparision as well, and all the associated options (like for rsync), but this here should suffice for now.

- it allows a basic verification for servers that do not support hash (formerly always "true")
- it speeds up verification in some cases

**But: maybe one should add an option to verify _only_ with size and not _size & checksum_?**

DasRaschloch:
> But: maybe one should add an option to verify _only_ with size and not _size & checksum_?

Yes, this should definitely be added to save time if a simple size comparison is sufficient and the transferred files are very large.

> I will play around with that idea later today. 

I think the easiest way to do this, would be to just extend FtpVerify. But as you already said, this functionality could be extended in the future, and then the question is how granular the user can set the verification. As there can be different combinations of verification methods. So a new enum might make more sense in terms of future-proofing.

robinrodricks:
I think this is fine, however I feel the logic should be like this:

1. First see if the server supports checksum, in that case do a checksum and nothing else
2. If the server does NOT support checksum, then ONLY check the size

This will not add any additional slow downs.

In the current implementation, it will slow down for point 1 above (if the server supports checksum) because it will do a file size check AND a checksum check.

FanDjango:
Change 1: `FtpVerifyMethod.SizeOrChecksum` should be `FtpVerifyMethod.ChecksumOrSize` to reflect the ordering of operation

But I propose a different approach, as the `FtpVerifyMethod.ChecksumOrSize` etc. etc. verbiage does not cater for any other methods imaginable.

Enum to be "Flags" type:

0x0000 - No verification
0x0001 - Size
0x0002 - Date/Time (currently no need to code it, but for later implementation)
0x0004 - Hash

So one could set Size | Hash, just Size or just Hash (or someday, Size | DateTime | Hash, like rsync does under linux).

It is understood though, that if more than one option is set, the lower value will per performed first and the sequence will be aborted with a verify fail if that option fails to verifiy.

Of course you could then allow config options like "FtpVerifyMethod.ChecksumOrSize" to set the right flag bits.

DasRaschloch:
> @DasRaschloch Do you want, or shall I continue?

i am almost done implementing the proposed changes into the AsyncClient

not including your new proposal.

FanDjango:
That's fine... go for it.

FanDjango:
What could be the best way to "take" all of your code, make **just** a few changes and show you how I would do it? Just so we can compare? I'll try it in a branch of my own, ok? and then let's discuss.

DasRaschloch:
> What could be the best way to "take" all of your code, make **just** a few changes and show you how I would do it? Just so we can compare? I'll try it in a branch of my own, ok? and then let's discuss.

sure

FanDjango:
I opened a PR to show what could be done as an alternative. It is not complete as to Robins requests totally, needs some scrutiny and possibly some special enums, but the building blocks are **yours** and the Enums are different, is all.

DasRaschloch:
Hmm, with one Enum it is certainly easier to configure

DasRaschloch:
Yeah, i think i like yours better, simply because you can still use one variable and don't need to add more.

From a separation of concerns point of view, two Enums would be "better", but who cares.

FanDjango:
> From a separation of concerns point of view, two Enums would be "better", but who cares. Ease of use is more important.

You know, when we would take a moment, we would see the merits of @robinrodricks idea:

The params of the API functions hold the Verify**Options**. Things like _Retry_ (which is already in config) and _how to do the verify_ should be in config - keeps the API call  statement small.

So maybe his **exact** list of things _todo_ together with the latest version of your verify code and a new config variable with its own Enum (how to do the verify) (which you have created in your version) is actually the best synthesis. Then the parameters need not be changed, and the verify code can just retrieve the config variable to decide what it does.

FanDjango:
So I am removing my sample PR, I really liked your new Enum, and the verify code. I didn't like all the signature changes and params. Let's put them in config like Robin said, right under the retry setting, using your Method enum.

DasRaschloch:
> Let's put them in config like Robin said, right under the retry setting, using your Method enum.

Probably the best idea.

DasRaschloch:
Can someone please verify that this is now correct and should result in the desired behaviour?
-- The FtpVerifyMethod enum has been added to FtpConfig.
-- The user can now use checksum, file size and modified date checks to verify that a transfer was successful.
-- These checks can be combined according to the wishes of the user.
-- If not specified, checksum is used as the default verification.
-- If only checksum is selected and the server does not support hashing, verification will fall back to a simple size comparison.

FanDjango:
I'm using/testing it right now...

FanDjango:
I made some changes, added some things from Robins list. I put it up as PR #1515 - just pick it up into your PR and I will merge later this afternoon.

FanDjango:
Looks really good now, no changes to API / Signatures, nice clean logic.

DasRaschloch:
The only thing I am not sure about is the addition of `SizeOnly`, `SizeOrChecksum` and `SizeThenChecksum`. The first two are just aliases for existing values and the last one can be set by the user. 
Instead, I would suggest that we add a line to the summary explaining how the user can combine different verification methods.

FanDjango:
> The first two are just aliases for existing values and the last one can be set by the user.

Of course you are right.

It is just that Robin likes "non-technical, explanatory" terms that in his opinion makes things clearer to the more or less non-ftp-technical user.

This way we support both?

Add as many explanatory lines as you feel are needed...

DasRaschloch:
> It is just that Robin likes "non-technical, explanatory" terms that in his opinion makes things clearer to the more or less non-ftp-technical user.

yes, and that is in general a good thing, but could be misleading here, as we now have multiple values doing the same thing but with different descriptions.

FanDjango:
I prefer yours. So remove the Robin ones. In that case we will wait for him to merger.

Note that the only majort change I made to your PR, was getting the config value inside the verification methods instead of passing them as paramters.

DasRaschloch:
> Note that the only majort change I made to your PR, was getting the config value inside the verification methods instead of passing them as paramters.

yeah, i didnt think about that.... This way its cleaner.

FanDjango:
I will merge this now. Any further changes can be done in an ongoing fashion.

DasRaschloch:
The Human Mnemonics are not yet removed

PR #1519, opened 2024-03-27
fileLen may be negative

andreasstolpmann:
If `GetFileSize` fails to determine the size `fileLen` will be -1 here. In that case we would end up with a negative amount of bytes to write.


FanDjango:
You are right. Thanks a lot - good catch!

FanDjango:
And I will add the coresponding change for SYNC

PR #1534, opened 2024-04-05
Restore call to FtpSslStream graceful TLS termination

FanDjango:
The recent changes to ASYNC and also cleanup of SYNC dispose had inadvertently taken the FtpSslStream "graceful" TLS termination code out of the sequence, effectively disabling it. These two "Close" statements bring this (hard earned) functionality back.

Note that the ASYNC flavor of this close is not totally block free.

FtpSslStream needs a slight recode to also honor the Async disposing paradigms introduced recently. TBD


robinrodricks:
Released!

https://www.nuget.org/packages/FluentFTP/50.0.0

image

PR #1554, opened 2024-04-16
Correct implementation of Buffering

FanDjango:
### DO NOT MERGE, EXPERIMENTAL

Up to now it was:

**Non-encrypted:**
Raw Socket/NetStream

**Encrypted:**
Raw Socket/NetStream
Wrapped by BufferedStream
Wrapped by SslStream

Which is just plain wrong. One should present buffered data to ssl for encryption to reduce overhead.

I changed the above stream configuration to be..

**Both encrypted and non-encrypted**
Raw Socket/NetStream
Wrapped by SslStream
Wrapped by BufferedStream <-- moved this

I changed the way the BufferedStream is created: Only in Open***DataStream (Async, Sync, Active, Passive) and thus the buffering is only used on data connections and not in control connections.

It works seamlessly over all .NET versions and with encryption (both SslStream and GnuTLS) or without encryption etc. but:

**I notice to my surprise on my tests that activation of buffering makes no significant difference on .NET Framework, .NET 5+, .NET Standard. It seems all of these .NET versions employ enough inner buffering already themselves. No need to double up.**

**We might as well drop this feature**, as it **was** before this PR effectively forcibly turned off in most use cases **anyway** (proxy, .NET 5+, .NET Standard, NOOP, control connection). It was really only active on .NET Framework, so I wonder who was actually using it... In the past some issues came up where users (falsely) attributed hangs after SSL authentication to be caused by Buffering, so the code was simply changed to disable it outright without real in depth testing or analysis.

So this PR is an invitation for discussion:

1. **We remove buffering**. Means we remove the config variable as well maybe, this would be (after V50) once more an API change for users (->V51).
2. **We use this new buffering code**. We can leave the config variable (Config.SslBuffering) as is, but the name is no longer really indicative of the functionality, or we change the name to Config.Buffering.
3. _And we perhaps change the default of the config variable to_ "No buffering".
4. **We keep the current logic** - and remove the fake/unneeded forcible disable logic so it is at least at the users discretion to enable it. 
5. **We don't touch the logic at all**.

### I would actually advocate removing it completely. After some testing, I find it makes no difference between active and inactive, either in the "old" variant or the "new" one.

### DIFF
--- a/FluentFTP/Client/AsyncClient/OpenActiveDataStream.cs
+++ b/FluentFTP/Client/AsyncClient/OpenActiveDataStream.cs
@@ -138,6 +138,8 @@ namespace FluentFTP {
 			stream.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.KeepAlive, Config.SocketKeepAlive);
 			stream.ReadTimeout = Config.DataConnectionReadTimeout;
 
+			stream.CreateBufferStream();
+
 			return stream;
 		}
 
--- a/FluentFTP/Client/AsyncClient/OpenPassiveDataStream.cs
+++ b/FluentFTP/Client/AsyncClient/OpenPassiveDataStream.cs
@@ -148,6 +148,8 @@ namespace FluentFTP {
 					token: token);
 			}
 
+			stream.CreateBufferStream();
+
 			return stream;
 		}
 	}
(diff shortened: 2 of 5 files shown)


robinrodricks:
I am in favour of **1. We remove buffering**.

No need to keep redundant junk features esp since they are only needed for older .NET framework targets.

FanDjango:
So be it...

Since a BufferedStream takes complete control of the underlying streams, disposal logic gets a lot easier when we don't use it.

Also, if the underlying stream is not seekable (i.e. no way to position) it is difficult to dispose a BufferedStream if there is still data pending.

And it is really mostly for unidirectional buffering.

I will close this "Discussion PR" and work in progress removing BufferedStream as per #1 above.

PR #1561, opened 2024-04-19
Remove BufferedStream feature, further refactoring of Dispose logic

FanDjango:
See #1554

All integration tests and my internal ones pass.

I will play with it some more in the next couple of days and then merge into master. We could leave it there for a short while and then cut a V51.


robinrodricks:
PR looks good, I am only concerned with this `Error disposing(sync)`. I would rather suppress all errors relating to disposing, otherwise our issues list will be flooded with dumb errors about disposing failures.

FanDjango:
>  I would rather suppress all errors relating to disposing...

Yes, that's a very valid point.

My thoughts were: As the code gets more and more mature, places where "it shouldn't" really happen are of prime interest and I would appreciate the error reports. I was actually so confident as to believe that we wouldn't get so many such reports, as the code is getting to be very clean (but that may be a very subjective view).

I'll remove the log entry later on...

robinrodricks:
Thanks, merged.

PR #1563, opened 2024-04-24
Utilize caching Regex calls

crobibero:
`new Regex()` doesn't cache the build, but `Regex.Match()` does

[`\d` matches much more than `[0-9]`](https://learn.microsoft.com/en-us/dotnet/standard/base-types/character-classes-in-regular-expressions#decimal-digit-character-d)

### DIFF
--- a/FluentFTP.CSharpExamples/CustomParser.cs
+++ b/FluentFTP.CSharpExamples/CustomParser.cs
@@ -30,11 +30,11 @@ namespace Examples {
 			Match m = null;
 			var regex =
 				@"(?<permissions>[\w-]{10})\s+" +
-				@"(?<objectcount>\d+)\s+" +
-				@"(?<user>[\w\d]+)\s+" +
-				@"(?<group>[\w\d]+)\s+" +
-				@"(?<size>\d+)\s+" +
-				@"(?<modify>\w+\s+\d+\s+\d+:\d+|\w+\s+\d+\s+\d+)\s+" +
+				@"(?<objectcount>[0-9]+)\s+" +
+				@"(?<user>[\w0-9]+)\s+" +
+				@"(?<group>[\w0-9]+)\s+" +
+				@"(?<size>[0-9]+)\s+" +
+				@"(?<modify>\w+\s+[0-9]+\s+[0-9]+:[0-9]+|\w+\s+[0-9]+\s+[0-9]+)\s+" +
 				@"(?<name>.*)$";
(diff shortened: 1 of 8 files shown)


robinrodricks:
I don't mind this PR, but so many critical sections are changed (like machine listing) and I feel like "if it aint broke don't fix it".

Even your link for the `\d` page is broken.

crobibero:
The link not working is Microsoft's issue, just search for "Decimal digit character: \d" on that page to get to where I attempted to link.

> \d matches any decimal digit. It is equivalent to the \p{Nd} regular expression pattern, which includes the standard decimal digits 0-9 as well as the decimal digits of a number of other character sets.


Test results:
The two failing tests fail on my machine on master as well.
image

robinrodricks:
@FanDjango  your views?

FanDjango:
Ok, I did some research, and...

**Concerning the \d->[0-9] change**

\d is true for all representations of numbers. Depending on code page such can be:
image

It is recommended to use [0-9].

In the case of FluentFTP, it makes no big difference, but it doesn't hurt either. I had actually coded [0-9] in my recent PR, but changed that to \d when I saw that that was ubiquitous elsewhere in FluentFTP.

**Concerning chaching enabled regex patterns**

None (Ok, maybe one) of the places are in hot code paths where the caching of regex search patterns would be sensitive to performance issues. That's what I would look at. But the changes, as long as they are semantically correct, are worth doing.

The problem remains: Are they correct? If something fails due to this, we'll just need to fix it. Shucks, I made so many changes in the past months - I am standing by anyway to catch any bloopers I might have caused.

So, let's all look them over again with eagle eyes and go for it.

**Concerning failed tests**
The failed parser test is a unix file list that contains a date reference that the test code thinks is 2023 and not 2024. It needs a fix or removal of the offending directory entry.

The other failure is in the custom tests. If they are not configured as per the users site, (it)(they) fail.

FanDjango:
I fixed the tests and am commiting that to master right now.

1. .NET 5+ AutoConnect was not reacting detecting a timeout properly
2. Unix directory listing parse tester was expecting wrong date

Tests (using current master) should now look like this (if custom tests are not configured):
image

But I don't think you need to re-run your test re. this PR.

robinrodricks:
Merged because regex change is minor and hopefully does not cause an issues.

robinrodricks:
Thanks for your contribution.

PR #1566, opened 2024-04-27
Implement DownloadPipe

bjornharrtell:
I'm investigating ways to pipe ftp downloads through async logic to avoid waiting for full download or having to store it temporarily.

It looks like simply using System.IO.Pipelines and the stream interface of Pipe would allow this.

I've not put this into practise yet but working on it, but what do you think about this addition to transfer methods?

### DIFF
--- a/FluentFTP/Client/AsyncClient/DownloadPipe.cs
+++ b/FluentFTP/Client/AsyncClient/DownloadPipe.cs
@@ -1,0 +1,48 @@
+﻿using System;
+using System.IO;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using FluentFTP.Streams;
+using FluentFTP.Helpers;
+using FluentFTP.Exceptions;
+using FluentFTP.Client.Modules;
+using System.Threading;
+using System.Threading.Tasks;
+using System.IO.Pipelines;
+
+namespace FluentFTP {
+	public partial class AsyncFtpClient {
+
+		/// <summary>
+		/// Downloads the specified file into the specified pipe asynchronously.
+		/// High-level API that takes care of various edge cases internally.
+		/// Supports very large files since it downloads data in chunks.
+		/// </summary>
+		/// <param name="pipe">The pipe that the file will be written to.</param>
+		/// <param name="remotePath">The full or relative path to the file on the server</param>
+		/// <param name="restartPosition">The size of the existing file in bytes, or 0 if unknown. The download restarts from this byte index.</param>
+		/// <param name="token">The token that can be used to cancel the entire process</param>
+		/// <param name="progress">Provide an implementation of IProgress to track download progress.</param>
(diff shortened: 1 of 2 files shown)


FanDjango:
IMHO, a nice idea, if you were to add the integration tests for this to prove the concept and some code samples for the example collection?

bjornharrtell:
@FanDjango actually, I just realized there is no need for this interface as I can just get the stream interface outside and use DownloadStream. Not sure why I didn't see that from the start...

FanDjango:
Well, don't get me wrong, but if it is such an "easy" additional API, that saves the user of the library some lines of code, why not do it? Even if it can be "wrapped" with some code on the users part easily?

bjornharrtell:
@FanDjango initially I wasn't sure stream was an abstraction that would work.. but it does. And that frees you from using a specific pipe implementation (even though System.IO.Pipelines is official/widely used). In any case, I think it's good to avoid dependency on System.IO.Pipelines or other pipeline implementation in this library.

I suppose an example in the "File Transfer" section would be nice. I could possibly try provide a simple one.

dbeattie71:
@bjornharrtell You should be able to use https://github.com/dotnet/Nerdbank.Streams to accomplish this.

robinrodricks:
> Well, don't get me wrong, but if it is such an "easy" additional API, that saves the user of the library some lines of code, why not do it? Even if it can be "wrapped" with some code on the users part easily? 

I am always in agreement with this basic principle. `DownloadPipe` seems to be a nice new API rather than forcing users to use `OpenRead`/`OpenWrite` and those old stream methods that don't have many of the high level functionality.

@bjornharrtell thoughts?

robinrodricks:
Seems interesting

bjornharrtell:
Still no longer see the benefit. The code involved to use a a pipe is simply this:

```csharp
var pipe = new Pipe();
using var writeStream = pipe.Writer.AsStream();
var downloadTask == client.DownloadStream(writeStream, fileName, 0, null, token);
// use pipe to fx. forward to another stream
```

Can't really make it more simple since the resource handling of the writeStream should be the callers responsibility.

dbeattie71:
> Still no longer see the benefit. The code involved to use a a pipe is simply this:
> 
> ```cs
> var pipe = new Pipe();
> using var writeStream = pipe.Writer.AsStream();
> var downloadTask == client.DownloadStream(writeStream, fileName, 0, null, token);
> // use pipe to fx. forward to another stream
> ```
> 
> Can't really make it more simple since the resource handling of the writeStream should be the callers responsibility.

I always used Nerdbank when I wanted to do this, I didn't realize AsStream was added in .NET 8, nice!

robinrodricks:
@bjornharrtell thanks for the code. Can you give me a code sample for UploadStream as well so I can update the wiki with an FAQ on pipes?

bjornharrtell:
Sorry haven't used the upload feature so that would require more time from me than I currently have at the moment.

PR #1572, opened 2024-05-08
Override Read(Span), ReadAsync(Memory), Write(ROS), WriteAsync(ROM) in streams

### LINKED ISSUE #1570: Override Read(Span), ReadAsync(Memory), Write(ROS), WriteAsync(ROM) in streams
Available in NETSTANDARD2_1 and NET5_0_OR_GREATER, the base class functions in Stream for backwards compatibility rent an array and copy the span to it and then call the old array based functions.

As an optimization `FtpDataStream` and `FtpSocketStream` can override these functions and pass the call directly onto the wrapped stream.

```
public override void Write(ReadOnlySpan<byte> buffer) {
    if (BaseStream == null) {
        return;
     }

    BaseStream.Write(buffer);
    m_lastActivity = DateTime.UtcNow;
}
```

Like how the old FtpClient has been changed https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Requests/src/System/Net/FtpDataStream.cs#L231

bbowyersmyth:
Available in NETSTANDARD2_1 and NET5_0_OR_GREATER, the base class functions in Stream for backwards compatibility rent an array and copy the span to it and then call the old array based functions.

As an optimization FtpDataStream and FtpSocketStream can override these functions and pass the call directly onto the wrapped stream.

Closes #1570

### DIFF
--- a/FluentFTP/Streams/FtpSocketStream.cs
+++ b/FluentFTP/Streams/FtpSocketStream.cs
@@ -552,6 +552,22 @@ namespace FluentFTP {
 #endif
 		}
 
+#if NETSTANDARD2_1 || NET5_0_OR_GREATER
+		/// <summary>
+		/// Reads data from the stream
+		/// </summary>
+		/// <param name="buffer">Buffer to read into</param>
+		/// <returns>The amount of bytes read from the stream</returns>
+		public override int Read(Span<byte> buffer) {
+			if (BaseStream == null) {
+				return 0;
+			}
+
+			m_lastActivity = DateTime.UtcNow;
+
+			return BaseStream.Read(buffer);
+		}
+#endif
 
 		/// <summary>
 		/// Reads data from the stream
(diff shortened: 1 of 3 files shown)


robinrodricks:
The changes look good. But are these new methods automatically used by the .NET Framework or the Stream system? As we are not invoking it directly.

FluentFTP/Streams/FtpSocketStream.cs
```diff
@@ -591,6 +607,44 @@ public class FtpSocketStream : Stream, IDisposable {
...
+		/// <param name="token">The <see cref="CancellationToken"/> for this task</param>
+		/// <returns>The amount of bytes read from the stream</returns>
+		public override async ValueTask<int> ReadAsync(Memory<byte> buffer, CancellationToken token) {
+			if (BaseStream == null) {
+				return 0;
+			}
+
+			m_lastActivity = DateTime.UtcNow;
+			using (var cts = CancellationTokenSource.CreateLinkedTokenSource(token)) {
+				cts.CancelAfter(ReadTimeout);
```
robinrodricks, line 624: what's all this about?

bbowyersmyth:
@robinrodricks Yes stream and other places in system will now prefer to use the new overloads if they don't already have an array. For example `FileStream.CopyToAsync` uses the memory overload https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/IO/Strategies/FileStreamHelpers.Windows.cs#L299C39-L299C49

So they may already be getting called. Will know if we can put a break point in there when the tests are run. Otherwise we will need to add more.

FluentFTP/Streams/FtpSocketStream.cs
bbowyersmyth: Same code as the existing ReadAsync above. It will be setting a timer to timeout the async call.

FanDjango:
I have also been investigating bypassing the BUF copying inherent in the FtpDataStream and FtpSocketStream on the way to (or from) network stream.

Newer .NET versions give one this option:

**Prefer the memory-based overloads of ReadAsync/WriteAsync methods in stream-based classes**

For example: `await s.ReadAsync(buffer.AsMemory(0, buffer.Length), token);`

Would of course be a code change and another set of #IF/#ENDIF brackets.

bbowyersmyth:
Added 2 more UploadDownload tests as the existing ones weren't using the streams directly.
Fixed a couple of compiler warnings in the file also.

robinrodricks:
This looks interesting. I'm merging it for now. If we have serious issues arising from this change, we can revert it.

Thanks for your contribution!

PR #1604, opened 2024-07-19
Final name for the daemon blocking Semaphore

FanDjango:
Well, @robinrodricks, you will still get your chance to improve the names :-) As you can see, I am floundering...


robinrodricks:
No problem with private var names. I usually fuss over the Config (FtpConfig) prop names coz thats what a lot of users will have access to.

PR #1605, opened 2024-07-19
Poll

FanDjango:
Adding a "PollDaemon" to the already existing "NoopDaemon".

1. Both can be used singly or simultaneously or **not at all**.
2. The Noop Daemon only sends **commands**, on the control connection, if a control connections is connected.
3. The Poll Daemon only "invokes" socket.Poll operations on the control conections **and** the data connections, if such a connection is connected.
4. The intervals can be configured independently.
5. The existing "Poll" mechanism is superseded, still works but can be removed soon, (if not right away?).

While developing and testing this, some needed improvements to the existing Noop daemon appeared:

1. Cancellation of the Deamon via Token
2. Subtle timing changes or oder of execution changes to better handle edge cases - this invariably involves locking/semaphore operation to handle the cases where the daemon might interfere with normal command handling.

Some "todo"s for seldom to be seen edge cases remain. I will address them in the near future.

One **big** caveat: Those who want to use VS debug mode with breakpoints are well advised to disable the daemons, because they derive their action triggers from the timing of "last actions" on the socket connections - if you sedately step through your FTP app, these will seem to misbehave timing-wise.

### DIFF
--- a/FluentFTP/Client/AsyncClient/Connect.cs
+++ b/FluentFTP/Client/AsyncClient/Connect.cs
@@ -236,11 +236,21 @@ namespace FluentFTP {
 
 			// FIX #922: disable checking for stale data during connection
 			Status.AllowCheckStaleData = true;
+						Status.InCriticalSequence = false;
 
-			Status.InCriticalSequence = false;
+			if (Config.Noop) {
+				if (Status.NoopDaemonTask == null) {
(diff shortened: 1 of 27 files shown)


FanDjango:
Closing this - I am going to add some stuff before pulling this entirely

1. Timeout and continue, for Daemon Termination
2. Timeout and continue, for Daemon Initialization (because this can fail on certain .NET versions.

This will be back asap as another PR

robinrodricks:
What does Polling do? Can u help me understand it so I can document it?

PR #1613, opened 2024-07-26
Fixed uncatchable Exception after cancellation of read in FTPSocketStream

Scartzz:
If you wanted to cancel a read operation, a non-catchable exception always flew here. This could not be handled due to the fact that it was registered on the cancellation token. 
The ReadTimeout still works, as the Read method throws an exception as soon as the cts expires.
In the new setup, the connection is still closed and an exception is now thrown correctly.


Scartzz:
Could be that this PR fixes the issue #1590

FanDjango:
Good catch...

FanDjango:
I mean, can you demonstrate with an instance of execution and a log, what you experience and also how this relates (perhaps) to #1590? You cancel in which circumstances and you get what?

FanDjango:
And yes, currently working on "hardening" FtpSocketStream.cs, so I would be keen on this being integrated into the next release.

robinrodricks:
Thanks, will be useful.

PR #1630, opened 2024-08-02
Remove socketPoll, improve Daemon, fix/maintain code

FanDjango:
### Daemon:
**Use** Task.Factory.StartNew
**Use** CancellationToken
**Improve** code and use intermediate connection state to indicate lost connection

### FtpSocketStream
**Improve** IsConnected - handle sync **and async** close situations, remove extranseous socket.Poll calls, use intermediate connections state to decide return valie
**Improve** Close/Dispose logic further - add detailed logging of Close/Dispose failures for future debugging in possible user logs when issues are reported

### New Enum FtpRealConnectionStates
**Allow connection failure status tracking** with the help of NoopDaemon before IsConnected decides true/false based on socket state only.

### DIFF
--- a/FluentFTP/Client/AsyncClient/Connect.cs
+++ b/FluentFTP/Client/AsyncClient/Connect.cs
@@ -86,7 +86,6 @@ namespace FluentFTP {
 
 			m_hashAlgorithms = FtpHashAlgorithm.NONE;
 			m_stream.ConnectTimeout = Config.ConnectTimeout;
-			m_stream.SocketPollInterval = Config.SocketPollInterval;
 			await ConnectAsync(m_stream, token);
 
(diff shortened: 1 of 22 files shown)


FanDjango:
### Re-opened PR

I have re-opened the pull request. I wanted it to be rebased and squashed so that all changes are easier to see.

### Previous questions, decisions made

`config.SocketPollInterval` - if its not used, then it can be deleted
**It is now deleted**

Noop Daemon be active by default? - does this have any risk of malfunctioning or causing some side effect? if it has NO risk and ONLY a benefit, then it can be enabled by default
**It is now disabled by default**

robinrodricks:
Thanks for your enormous efforts put into this release! Muchas gracias!

https://www.nuget.org/packages/FluentFTP/51.0.0

Issue #1635, opened 2024-08-21
Response: 500 MLST command not understood [Axway Server]


anandgmenon:
**FTP Server OS:** Unix

**FTP Server Type:** Axway

**Client Computer OS:** Windows

**FluentFTP Version:** 49.0.2

**Framework:** .NET 6

Hey I'm back with another issue/question. So one of our users faced an issue with the `GetObjectInfo()` method. It seems to be returning
_Response: 500 'MLST /CER/INBOX/TEST_CLM.096FB412.20240820161946524': command not understood._ See detailed logs below. My understanding from the code is FluentFTP checks if MLST is supported on the server or else fallback to get folder listiting to get the file metadata. Do we have a bug in this logic ?

**Logs :**

```
>         Connect(False)
Status:   FluentFTP 49.0.2.0(.NET 6.0)
Status:   Connecting to IP #1= ***:21
Status:   Waiting for a response
Response: 220 10.2.242.199 FTP server 5.5-20220825 ready.     [739117.852d]
Command:  USER ***
Status:   Waiting for response to: USER ***
Response: 331 Password required for ***. [51ms]
Command:  PASS ***
Status:   Waiting for response to: PASS ***
Response: 230 virtual user *** logged in from /52.179.140.75:60198. [69ms]
Command:  FEAT
Status:   Waiting for response to: FEAT
Response: 211-Extensions supportedResponse: AUTHResponse: USERResponse: PASSResponse: QUITResponse: PORTResponse: PASVResponse: EPSVResponse: EPRTResponse: TYPEResponse: STRUResponse: MODEResponse: RETRResponse: STORResponse: ABORResponse: DELEResponse: CWDResponse: XCWDResponse: LISTResponse: NLSTResponse: SITEResponse: SYSTResponse: STATResponse: HELPResponse: NOOPResponse: MKDResponse: XMKDResponse: RMDResponse: XRMDResponse: PWDResponse: XPWDResponse: CDUPResponse: XCUPResponse: SIZEResponse: MDTMResponse: RNFRResponse: RNTOResponse: RESTResponse: FEATResponse: ADATResponse: PROTResponse: PBSZResponse: APPEResponse: XCRCResponse: CCCResponse: COMBResponse: MLSDResponse: UTF8
Response: 211 END [23ms]
Status:   Text encoding: System.Text.UTF8Encoding+UTF8EncodingSealed
Command:  OPTS UTF8 ON
[...]

PR #1636, opened 2024-08-21
Rename FEAT capability MLSD -> MLST

FanDjango:
See #1635 

The capability should be called MLST in the code and not MLSD. Global rename needed.

### DIFF
--- a/FluentFTP.Tests/Unit/ParserTests.cs
+++ b/FluentFTP.Tests/Unit/ParserTests.cs
@@ -62,7 +62,7 @@ namespace FluentFTP.Tests.Unit {
 		public void Machine() {
 
 			var client = new FtpClient();
-			client.SetFeatures(new List<FtpCapability> { FtpCapability.MLSD });
+			client.SetFeatures(new List<FtpCapability> { FtpCapability.MLST });
 			var parser = new FtpListParser(client);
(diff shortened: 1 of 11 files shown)


robinrodricks:
It looks ok, but it was my understanding that FTP servers can use MLSD and/or MLST?

FanDjango:
In the RFC, it is written: Either both, or none of them. The availability of both is advertised with MLST. An advertisement MLSD does not exist.

But as one can see, some FTP servers ignore the RFC and advertise only MLSD. Some of them then support only MLSD. Some of them support both commands. 

This PR changes only the name MLSD to MLST. It does not change the current logic of allowing both MLST and MLSD advertisements to set the enum "MLST" to true.

It still leaves the decision open, to

1. Adhere to the RFC.

This will hamper **all** none RFC compliant servers in different ways, depending on their deviation.

2. Consider MLST and MLSD advertisements to be synonymous. This is the current implementation, even with this PR applied.

3. Fix #1635 by considering a MLSD advertisement to mean: MLSD command (but not MLST) is supported. This will break an unkown number of servers that need "MLST and MLSD advertisements to be synonymous" (2 above).

Advice?

robinrodricks:
I think this should cover it: - https://github.com/robinrodricks/FluentFTP/issues/1635#issuecomment-2303868824

FanDjango:
Yes. It does.

PR #1650, opened 2024-09-22
fix LogMaskModule throwing exception when UserName is empty (fixes #1649)

### LINKED ISSUE #1649: FtpClient throws exception from log masking when Credentials.UserName is empty
**FluentFTP Version:** 51.1.0
**Framework:** Happens on both .NET Framework 4.8 and .NET 8.0

# Summary

`LogMaskModule` tries to replace `UserName` instances with a fixed mask string, but if `UserName` is empty, then `String.Replace` throws an exception saying "**System.ArgumentException: 'The value cannot be an empty string. (Parameter 'oldValue')'**".

# Applicability

FTP spec (RFC 959) doesn't allow an empty username, but there's nothing that prevents an FTP client from sending an empty `USER` command, and the server from accepting it. 

Ideally, `LogMaskModule` should check for empty UserName and skip processing. `FtpClient` should reject an empty UserName directly in the constructor preferably with a manual override for exotic use cases. 

# Workarounds

- Enable `LogUserName = true` in configuration (which isn't ideal)

## Repro:

Here's a simple test program. Removing NetworkCredential parameter fixes this issue:

```csharp
using System.Net;
using FluentFTP;

namespace FtpTest
{
    internal class Program
    {
        static async Task Main(string[] args)
        {
            string path = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
            Directory.CreateDirectory(path);
            Console.WriteLine($"Server path: {path}");
            var server = new Zhaobang.FtpServer.FtpServer(new IPEndPoint(IPAddress.Loopback, 4444), path);
            var cancelTokenSource = new CancellationTokenSource();
            var cancelToken = cancelTokenSource.Token;
            var task = Task.Run(() => server.RunAsync(cancelToken));

            var client = new AsyncFtpClient("127.0.0.1", new NetworkCredential()
            {
                UserName = "",
                Password = "",
            },4444, new FtpConfig()
            {
                LogToConsole = true,                
            });
            if (!await client.DirectoryExists("test/mest"))
            {
                await client.CreateDirectory("test/mest", force: true);
            }
            await client.UploadFile("D:\\test.txt", "test/test.txt");
            Console.WriteLine("DONE! ");
            cancelTokenSource.Cancel();
            Console.WriteLine("CANCELLED TOO! ");
            await task;
        }
    }
}
```

ssg:
This is my recommended fix. However, I wasn't able to build this due to many `IDE0032` errors, and I didn't want to touch anywhere in the codebase. Should build fine though.

### DIFF
--- a/FluentFTP.Tests/Unit/ConstructorTests.cs
+++ b/FluentFTP.Tests/Unit/ConstructorTests.cs
@@ -1,0 +1,62 @@
+﻿using System;
+using System.Net;
+using Xunit;
+
+namespace FluentFTP.Tests.Unit {
+	public class ConstructorTests {
+
+		const string defaultHost = "127.0.0.1";
+
+		[Fact]
+		public void FtpClient_NullCredentials_Throws() {
+			Assert.Throws<ArgumentNullException>(() => new FtpClient(defaultHost, null));
+		}
+
+		[Fact]
+		public void AsyncFtpClient_NullCredentials_Throws() {
+			Assert.Throws<ArgumentNullException>(() => new AsyncFtpClient(defaultHost, null));
+		}
+
+		[Fact]
+		public void FtpClient_EmptyUserNameCredential_Throws() {
+			Assert.Throws<ArgumentException>(() => new FtpClient(defaultHost, new NetworkCredential("", "")));
+		}
+
+		[Fact]
+		public void FtpClient_NullUserName_Throws() {
+			Assert.Throws<ArgumentNullException>(() => new FtpClient(defaultHost, null, ""));
+		}
+
+		[Fact]
+		public void FtpClient_NullUserPassword_Throws() {
+			Assert.Throws<ArgumentNullException>(() => new FtpClient(defaultHost, "test", null));
+		}
+
+		[Fact]
+		public void FtpClient_EmptyUserName_Throws() {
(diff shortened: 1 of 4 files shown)


ssg:
Ouch. I didn't intend the last two commits to make it into this PR. Let me know if you want them removed, I can refresh this.

FanDjango:
> However, I wasn't able to build this due to many IDE0032 errors

Huh? Not able to build?

ssg:
Yes, IDE0032 errors show up as compiler errors, not warnings.

FanDjango:
Very interesting. Not so on my system. Could we as a first step try to find out how this can be?

ssg:
Mine is also 17.11.4. I’ll look into it.

ssg:
Funny, I `git checkout`'ed that commit, and now, I'm not getting the error. Anyway, I force pushed that specific commit, and discarded all other changes.

ssg:
It came back :) But let's disregard it if it works in the build. 

image

FanDjango:
Probably IDE0032 is configured to be an error in your VS config.

FanDjango:
Looks good

ssg:
> Probably IDE0032 is configured to be an error in your VS config.

I don’t even know where such a config option would be. And I searched.

FanDjango:
VS must keep this **somewhere**:
image

PR #1667, opened 2024-10-19
Improve dictionary lookups

ssg:
I was investigating a race condition in my app. I haven't found the root cause yet, but this part took my attention in FluentFTP's code when browsing through the stack:

- The `ContainsKey` + retrieve pairs are redundant when a single `TryGetValue` would work. 
- A dictionary `.Add` would throw `ArgumentException` for duplicate keys, but `cache[host] = value` wouldn't.

I know that there is no reason for those pieces of code to fail for a single-threaded type, but still it's marginally faster, and more fault tolerant :)


FanDjango:
Thanks for these improvements. Try... versions are definitely better as they internally block the possibilty of a multi-thread intrusion. One of those code places was my "fault" originally,

PR #1683, opened 2024-11-29
Step 2 (final) to fix FileZilla Server dockerfile

FanDjango:
This was a bit of work. Instead of building all the required libraries and FileZilla Server itself, now use the provided .deb debian package install into a vanilla debian system.

A bit of finagling was needed to get the download link from the filezilla sites download php which adds a volatile key to the link.


robinrodricks:
Incredible work!

robinrodricks:
Released.

https://www.nuget.org/packages/FluentFTP/51.2.0

robinrodricks:
Released.

https://www.nuget.org/packages/FluentFTP/52.0.0

Issue #1698, opened 2025-01-01
Confusing auto-connect logic


kenkendk:
**FTP Server OS:** Any

**FTP Server Type:** Any

**Client Computer OS:** Any

**FluentFTP Version:** 52.0.0

**Framework:** .NET 8

This is not really an error with FluentFTP, but more a confusing auto-feature, that was hard to isolate because it "mostly works". Only in some cases where a retry (outside of FluentFTP) is attempted, this would get triggered. This got mixed up in #1690 because that is one way to trigger it.

I distilled it down to the following code that works:
```csharp
var client = new AsyncFtpClient() { ... };
var dir = await client.GetWorkingDirectory();
await client.UploadStream(...);
```

But if you do not read the working directory, you get an exception with the message "Please call Connect() before trying to read the Capabilities!":
```csharp
var client = new AsyncFtpClient() { ... };
await client.UploadStream(...);
```

It is of course an error not to call `await client.Connect()`, so one could argue that this is as intended.
But the fact that some methods will automatically (re-)connect and others not makes it confusing. In my case it was hard to pin down, because the `List` method appears to connect automatically.

I would suggest either making all methods auto-connect or all methods not auto connecting.
Alternatively, you could consider checking the state of the client on the various calls to give an earlier exception/stack-trace that makes it clear it is the caller making the mistake, and not an internal check that fails.

FanDjango:
> But the fact that some methods will automatically (re-)connect and others not makes it confusing

I agree.

> I would suggest either making all methods auto-connect or all methods not auto connecting

I have spent very many hours myself tracking down the sort of problems you mention.

**not auto connecting** would be my choice.

The **only** kind of "hidden" or "secret" re-connect then left over, would be the one that happens in the case of stale data being left over on the socket prior to the next command that needs to be sent to the server, although even that reconnect is a sort of kludge to clear up strange connection states. Getting that to work was up to now quite difficult, think about re-creating the state of the control session, the current directory, the current transfer type (binary/ascii) etc. Any custom servers (like z/OS, for example) might introduce more things that would need to be restored.

It was not my decision to do these things, auto-connect, auto-re-connect. They have a number of advantages, but meh.

The intention was to make things easier for the user, so he does not need to do the kind of checking he isn't really familiar with, FTP being such a strange protocol which is also not implemented consistently across all servers regarding the RFCs.

But the implementation is not totally clean, and thus you have admirably described the current state of affairs. Thanks for that.

[...]

Issue #683, opened 2021-02-23
Capabilities & HashAlgorithms can cause sync-over-async


jnyrup:
**FTP OS:**
Unix (based on vsFTPd being for unix-like systems)

**FTP Server:**
vsFTPd 2.2.2 (based on STAT)
Url: ftp://tgftp.nws.noaa.gov

**Computer OS:**
Windows Server 2019

**FluentFTP Version:**
33.0.3

While investigating the two stack trace posted below I started wondering why `Capabilities` and `HashAlgorithms` contained
```cs
if (m_stream == null || !m_stream.IsConnected) {
    Connect();
}
```

My best guess is to initialize them, if someone calls them before calling `Connect` or `ConnectAsync` which initializes them.
I'm not sure why it would be necessary to reconnect to get those values if `m_stream` was nulled or disconnected, but it can cause a sync-over-async path, `VerifyTransferAsync -> ... -> Connect()` as seen in the two stack traces.

Without being familiar with the codebase or all the edge cases of ftp connections, I think this might improve the situation.
```cs
private List<FtpCapability> m_capabilities;

/// <summary>
/// Gets the server capabilities represented by an array of capability flags
/// </summary>
public List<FtpCapability> Capabilities {
    get {
        if (m_capabilities == null && (m_stream == null || !m_stream.IsConnected)) {
            Connect();
            if (m_capabilities == null) {
                m_capabilities = new List<FtpCapability>();
            }
        }

        return m_capabilities;
    }
    protected set => m_capabilities = value;
}

private FtpHashAlgorithm? m_hashAlgorithms;

/// <summary>
/// Get the hash types supported by the server, if any. This
/// is a recent extension to the protocol that is not fully
/// standardized and is not guaranteed to work. See here for
/// more details:
/// http://tools.ietf.org/html/draft-bryan-ftpext-hash-02
/// </summary>
public FtpHashAlgorithm HashAlgorithms {
    get {
        if (!m_hashAlgorithms.HasValue && (m_stream == null || !m_stream.IsConnected)) {
            Connect();
            if (!m_hashAlgorithms.HasValue) {
                m_hashAlgorithms = FtpHashAlgorithm.NONE;
            }
        }

        return m_hashAlgorithms.Value;
    }
    private set => m_hashAlgorithms = value;
}
```

**Logs :**

Not sure if trace logs are relevant here.

```
System.NullReferenceException: Object reference not set to an instance of an object.
[...]

PR #1704, opened 2025-01-08
Ftp self connect mode

FanDjango:
Referring to #683 (long read but important)

#683 reported that accessing certain properties (specifically Capabilities and Hash Algorithms) **when the (async)client is not connected** would cause a cross-Async->Sync call to (sync)Connect(). That analysis was correct (**at that time**).

This was a result of extensive code changes and additions in the async client.

The (rather quick) fix was to remove any call to connect() in the property getters.

Later on in the history of development, reconnects due to stale data and just generally allowing API calls without a preceeding connect led to a certain confusion about which APIs are needing a connect in the first place (or not).

#1698 Reported the errors that can occur when accessing properties without a prior connect.

Addressing these was a bit involved.

This PR **brings back** the premature removal of connect() calls in certain property getters **and** introduces a config setting to steer the availability of "**automatic**" (re-)connects. It overcomes the danger of a cross-Async->Sync call to (sync)Connect(), which was the **original** problem identified by #683, not by a quick fix of just removing the connect calls but rather coding them in such a way that correct functionality is guaranteed.

The new config setting:
```cs
		/// <summary>
		/// Configure the behaviour of the Self Connect feature for the control connection
		/// Always (default): If the control connection is needed to process an API process (either for entering commands to the server
		/// or to retrieve information about the server (FEAT capabilities, HASH algorithms, Current Working directory) it will be (re-)
		/// connected if not available.
		/// OnConnectionLost: As with "Always", BUT only if there had been a connection in the first place before it got lost.
		/// Never: Connections will not be made unless you explicitly call Connect(...) yourself. If a connection is needed and not
		/// established, your API processes will fail instead of attempting to (re-)connect.
		/// </summary>
		public FtpSelfConnectMode SelfConnectMode { get; set; } = FtpSelfConnectMode.Always;
```

The new config enum for `client.Config.SelfConnectMode` is the following and it keeps things ASIS as the default is "Always":

```cs
	public enum FtpSelfConnectMode {

		/// <summary>
		/// Always (default): If the control connection is needed to process an API process (either for entering commands to the server
		/// or to retrieve information about the server (FEAT capabilities, HASH algorithms, Current Working directory) it will be (re-)
		/// connected if not available.
		/// </summary>
		Always,

		/// <summary>
		/// OnConnectionLost: As with "Always", BUT only if there had been a connection in the first place before it got lost.
		/// </summary>
		OnConnectionLost,

		/// <summary>
		/// Never: Connections will not be made unless you explicitly call Connect(...) yourself. If a connection is needed and not
		/// established, your API processes will fail instead of attempting to (re-)connect.
		/// </summary>
		Never,
	}
```

**As a second stage in the future, it will be necessary to migrate the two property getters, Capabilites and Hash Algorithms to be 
methods instead of properties. Such a migration is also pending (as a further example) for the `IsConnected` property of the `FTPSocketStream`.**

Such getters are not really supposed to invoke complex  operations, although there is no formal rule against this. In all three cases (`FtpSocketStream.IsConnected`, `client.Capabilites`, `client.HashAlgorithms`) converting these to methods is **quite involved**).


robinrodricks:
Yes absolutely. In principle you are right. I hated that "auto connect" functionality from day 1, but it was inherited from the original code base (read: 10 years old).

I could never understand - if I want to perform an FTP operation, why should the client "auto connect" on its own? ("self connect" in your terminology.)

I am ok removing this self connect completely. If not connected we can throw an error saying ("You are not connected to the FTP Server! Please call client.Connect() before using this API!").

It will be a major version increment. But I would rather users connect intentionally and not a side effect of reading props or using any other API.

"Self Connect" sounds strange but is ok since "Auto Connect" means something else in our library. I like your enum. It fits with our design principles. If you needed approval, you have it.

FanDjango:
Thanks @robinrodricks.

That leaves us to make a decision:

1. Remove all implicit `Connect(...)`s alltogether, but keeping the one that **re-**connects on stale data being encountered (that one is inside `Execute(...)`.
2. Use the new PR's code but decide on which default behaviour would be the best.

Let's gently nudge people to need a connection before APIs are allowed (where applicable) and make the default be `OnConnectionLost` **instead** of `Always`.

The `Never` behaviour would literally turn off any re-connects on reaching the SSL transaction limit or spurious stale date on the connection and quite a few users would suddenly have processes that unexpectedly fail.

Those who would like complete absolute control of the protocol states would set "Never".

robinrodricks:
> Make the default `OnConnectionLost` instead of Always.

Yes, this sounds good.

PR #1717, opened 2025-02-09
Remove language specific checks - PART 1

FanDjango:
Having language specific strings as a check for some condition is not good.

Example:

```
			"size not allowed in ascii",
			"n'est pas autorisé en mode ascii",
			"не разрешено в режиме ascii"
```

There is a better way to do this.

Note that for a better understanding of the what and the why for
`550 SIZE command not allowed in ASCII mode`
read this:

Note: Some servers do not support calculating the file size for a file that
will be converted to ASCII on download - Pre-calculating that costs time and
storage for the server, which can hurt performance server-side.
The tradeoff is to issue a TYPE I first if needed, potentially causing overhead
that detriments the transfer of multiple files.

Since this "550 SIZE command not allowed in ASCII mode" error message (when doing a SIZE command), is the **o n l y** error message tagged as 550, **all other 550 messages refer to  the files not being findable, regardless of sub-reason: path not found, spelling mistake, whatever.

So there is no need to query the ServerStringModule for possibly wrong error message texts.


robinrodricks:
A daring change, but well put and I like that improves the robustness of the library.

PR #1718, opened 2025-02-09
Remove language specific checks - PART 2

FanDjango:
This german text is not being used at all. No reference to it.


robinrodricks:
Ok great!

Issue #116, opened 2017-06-01
Do you support FTP commands CCC, PBSZ, PROT?


northwestfrog:
The following is only a question.

Let me start by the fact you provide a nice API.
I want to use this API in a context of FTP over explicit TLS.
In the list of supported commands (may be implicit) I did not see the three following commands CCC, PBSZ, PROT. Are these commands supported?

CCC is so the command socket gets unencrypted. Issued after a successful logon. Having it unencrypted allows the connection to an FTP server behind a firewall and having the PORT and PASV commands doctored by the firewall

Also, please provide some details as to the support of certificates that are not fully compliant

Thanks

hgupta9:
Not to my knowledge, but you can always try `ftp.Execute("CCC")` to send any custom command such as [CCC](https://www.smartftp.com/support/kb/clear-command-channel-ccc-f2551.html). To revert back to the insecure way, we'll have to upgrade the FtpClient class to support that.

PBSZ and PROT are automatically sent when you call Connect() and when EncryptionMode and DataConnectionEncryption are set. You can set DataConnectionEncryption to false even when in FTPS, perhaps this is what you are looking for? I'm not the original author of the library so I cannot confirm exactly how it works.  See [this](https://github.com/hgupta9/FluentFTP/blob/master/FluentFTP/FtpClient.cs#L1101).

northwestfrog:
You may want to implement the CCC command (see RFC4217). This removes the encryption from the basic communication channel/socket (as opposed to the DataConection) This would become handy so PORT and PASV commands are visible to a properly configured firewall.
Performing a ftp.Execute("CCC") would irremediably put an end to the communication with the FTP server. The client side would still consider an encrypted communication channel and the server side would consider this to be unencrypted from that point.

hgupta9:
I agree its a good idea, but I don't know how to unencrypt the channel just yet. I'll have to study the source code further.

hgupta9:
I have looked at the code and I think it can be implemented... Are you interested in helping debug it? The API will be client.DisableEncryption().

northwestfrog:
I will be a pleasure to debug this for you

hgupta9:
I tried to support it by sending "CCC" then closing the SSL stream, but after the 200 OK response, the server responds with garbage data. All further commands timeout, whether I send it to the SSL stream or the plaintext stream.

[...]

PR #1721, opened 2025-02-18
Added SwitchToClearText.

tsvetko:
Added switch to clear text method in async client.
The methods sends a CCC command and switches FTPScoketStream to unsecure mode.

### DIFF
--- a/FluentFTP/Client/AsyncClient/SwitchToClearText.cs
+++ b/FluentFTP/Client/AsyncClient/SwitchToClearText.cs
@@ -1,0 +1,19 @@
+﻿using System.Threading.Tasks;
+
+namespace FluentFTP {
+	public partial class AsyncFtpClient {
+
+		/// <summary>
+		/// Send a 'CCC' comamnd to the FTP server and if successful closes the undelying ssl stream.
+		/// </summary>
+		public async Task SwitchToClearText() {
+
+			var reply = await Execute("CCC");
+			if (reply.Success) {
+				m_stream.SwitchToUnsecuredMode();
+				Config.EncryptionMode = FtpEncryptionMode.None;
+			}
+		}
+
+	}
+}
--- a/FluentFTP/Streams/FtpSocketStream.cs
+++ b/FluentFTP/Streams/FtpSocketStream.cs
@@ -1866,6 +1866,15 @@ namespace FluentFTP {
 			};
 		}
 
+		/// <summary>
+		/// Closes the undelying ssl stream.
+		/// </summary>
+		internal void SwitchToUnsecuredMode() {
+			if (m_sslStream != null) {
+				DisposeSslStream();
+				m_sslStream = null;
+			}
+		}
 	}
 
 #pragma warning restore CS1998 // Async method lacks 'await' operators and will run synchronously


FanDjango:
Hi, this is nice! I guess you are using this and it works (of course).

Let's say it gets merged: But we also need to have this in the Sync client (shouldn't be difficult, right?).

**And** we would need a "NotImplemented" exception in case of a custom stream, as CCC is currently not implemented there.

tsvetko:
Hi, 
the proposed change adds support for CLEAR COMMAND CHANNEL mode. 
In this mode the user credentials are transmitted via secure channel and later the rest of the commands are transmitted via unprotected one. 
The use case is really limited since it requires the server to have white listed its client IPs to have a basic security but some of our clients are using it. 
Please let me know if you are willing such a feature to be introduced to FluentFTP and I will implement your comments. If you believe it is not appropriate I can close the PR.

FanDjango:
No, it's fine. It's time to add this, finally, after I cleaned up the dispose logic for the streams...

Can you post a log of a CCC command and the aftermath? I don't have any server that allows it currently. I need to see how clean the dispose and the subsequent command traffic in clear text looks. Just a small fragment, if possible.

FanDjango:
Also, here are some more points that need to be considered.

Note: These do not preclude merging this, they are just things that perhaps would need to be done additionally as a result:

1. Provide the new API call as sync as well.
2. Throw the apropriate exception if server FEAT response does not show CCC as being supported. Perhaps provide a config variable to override this logic to enable CCC even if it is not mentioned in FEAT.
3. Throw "not implemented" if using a custom stream (like GnuTLS). Dunno if this will **ever** be supported.
4. Exception throw: Restricting the use to connections that don't use TLS1.3 or higher. I think CCC would disrupt TLS1.3+ session resume and session ticket scheme, when data connections are then opened for "fresh encryption" from an unencrypted connection.
5. (Automatic?) restricting the use of the FluentFTP auto-restart mechanisms (on disconnect, on stale data) **after** a CCC is in effect.

So to speak: A notice to any user of CCC API that there are numerous things to watch out for when using this API.

> Please let me know if you are willing such a feature to be introduced to FluentFTP and I will implement your comments. If you believe it is not appropriate I can close the PR.

Yes, let's do it.

I will also complete some of the above things once you say you are ready/finished.

There is also the question of the **name** for this API. Normally I would have expected @robinrodricks to have chimed in by now concerning "SwitchToClearText". I also remember there was an issue from 2017? where he was already confronted with a request for CCC, which was actually implemented but in the end it had numerous failings. Some (tiny) code remnants testify to that.

robinrodricks:
There is already a `PlainTextEncryption` API. Can you investigate it.

FanDjango:
> There is supposed to be a PlainTextEncryption API which sends "CCC" command

I'll search for when it disappeared. I cannot remember ever having seen it.

robinrodricks:
CCC was implemented in this thread. 

https://github.com/robinrodricks/FluentFTP/issues/116

I have no issues merging this PR, provided the contributor fixes the issues you highlighted.

FanDjango:
> CCC was implemented in this thread.

Yes, that's the thread I discovered and mentioned in here, previously. But that API call got lost, I will research and tell you exactly when it disappeared.

FanDjango:
I found it. Took me a while, but here is what happened to it.

> There is supposed to be a PlainTextEncryption API which sends "CCC" command.

Firstly, it **was not an API command**, it was a `config` parameter to deactivate encryption **right after** user/password auth. Of course, the connect code used a CCC command and some code in the stream to do this.

You removed this on Sep. 2, 2022 as part of the async/sync client redesign leading up to V40.

It wouldn't have worked anyway beyond very simple tests on a sync client and only for NETFRAMEWORK, it was a nice fix from #116 but I don't think it was used by more than the one person who created the issue #116 - so its failings never showed up.

So it was right that it was dropped.

Should one think about recreating this, it might be more streamlined as a config parameter, not giving the user the freedom to CCC at any time he wants?

robinrodricks:
I'm ok either ways. Whatever suits the usage model better.

FanDjango:
@tsvetko Are you still in on this? Opinion?

tsvetko:
I think the way to go is to add some validations to SwitchToCkearText and to extend the sync version.

Tsvetko

________________________________

robinrodricks:
@tsvetko Could you add the sync version?

tsvetko:
Sure, will update the current PR. Probably early next week.

Tsvetko

________________________________

FanDjango:
And I'll do the other stuff... bad weather coming.

FanDjango:
@tsvetko 

I've started work on this - and actually choosing to not make it an API but rather as a config parameter. So I hope this reaches you in time before you get too far.

FanDjango:
Ok. I have implemented this in #1730.

Have a look at it and then I will merge it.

The concerns (above) have either been addressed or are no longer relevant:

**Done** Provide the new functionality as sync as well.
Throw the apropriate exception if server FEAT response does not show CCC as being supported. ~~Perhaps provide a config variable to override this logic to enable CCC even if it is not mentioned in FEAT~~.
~~Throw "not implemented" if using a custom stream (like GnuTLS). Dunno if this will ever be supported.~~
**Not done, as this should actually work** Exception throw: Restricting the use to connections that don't use TLS1.3 or higher. ~~I think CCC would disrupt TLS1.3+ session resume and session ticket scheme, when data connections are then opened for "fresh encryption" from an unencrypted connection.~~
**Not done. Works fine** (Automatic?) restricting the use of the FluentFTP auto-restart mechanisms (on disconnect, on stale data) after a CCC is in effect.

Some additional work was needed for the AutoConnect() API, as a CCC failure caused this to continue with the next type of connection to try.

tsvetko:
Just verified your changes with our server. The command is handled successfully. Will close this PR.

FanDjango:
Oh, that's nice of you to test it right away. I'll merge it ASAP

FanDjango:
See also #1730

PR #1740, opened 2025-04-22
Cleanup

jnyrup:
Various cleanups suggested by analyzers:
* https://www.nuget.org/packages/Meziantou.Analyzer
* https://www.nuget.org/packages/Roslynator.Analyzers
* https://www.nuget.org/packages/Microsoft.VisualStudio.Threading.Analyzers

### DIFF
--- a/FluentFTP/Client/AsyncClient/AutoDetect.cs
+++ b/FluentFTP/Client/AsyncClient/AutoDetect.cs
@@ -23,9 +23,7 @@ namespace FluentFTP {
 
 			LogFunction(nameof(AutoDetect), config);
 
-			if (config == null) {
-				config = new FtpAutoDetectConfig();
-			}
+			config ??= new FtpAutoDetectConfig();
 
 			ValidateAutoDetect();
 
@@ -33,7 +31,7 @@ namespace FluentFTP {
 		}
 
 		/// <summary>
- 		/// LEGACY CALL FORMAT, to be deleted sometime in the future
+		/// LEGACY CALL FORMAT, to be deleted sometime in the future
 		/// 
 		/// Automatic FTP and FTPS connection negotiation.
 		/// This method tries every possible combination of the FTP connection properties, and returns the list of successful connection profiles.
--- a/FluentFTP/Client/AsyncClient/AutoDetect.cs
+++ b/FluentFTP/Client/AsyncClient/AutoDetect.cs
@@ -32,12 +32,13 @@ namespace FluentFTP {
 
 		/// <summary>
 		/// LEGACY CALL FORMAT, to be deleted sometime in the future
-		/// 
+		/// <para>
 		/// Automatic FTP and FTPS connection negotiation.
 		/// This method tries every possible combination of the FTP connection properties, and returns the list of successful connection profiles.
 		/// You can configure it to stop after finding the first successful profile, or to collect all successful profiles.
 		/// You can then generate code for the profile using the FtpProfile.ToCode method.
 		/// If no successful profiles are found, a blank list is returned.
+		/// </para>
 		/// </summary>
 		/// <param name="firstOnly">Find all successful profiles (false) or stop after finding the first successful profile (true)</param>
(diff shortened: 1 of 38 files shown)


FanDjango:
Oof - a lot of those are brainfarts of mine. But not all, by no means.

Working on double checking these, gimme a couple of days, ok.

Thanks for doing this.

jnyrup:
I forgot to mention, I've split the changes into a commit per change for easier review.
None of the changes in this PR should change behavior.

FanDjango:
Hmmm. I don't agree with this one: "Remove redundant null checks"

If it were written more nicely:

```cs
A * B || !A * C
```

```cs
ServerHandler == null && FtpCapability.MDTM || ServerHandler != null && !ServerHandler.DontUseMdtmEvenIfCapable(path)
```

In words:

If SvrHnd is null use the capability, if it is not null determine what to use.

Or am I missing something?

FanDjango:
All others, I totally agree with.

jnyrup:
> Hmmm. I don't agree with this one: "Remove redundant null checks"
> 
> If it were written more nicely:
> 
> ```cs
> A * B || !A * C
> ```
> 
> ```cs
> ServerHandler == null && FtpCapability.MDTM || ServerHandler != null && !ServerHandler.DontUseMdtmEvenIfCapable(path)
> ```
> 
> In words:
> 
> If SvrHnd is null use the capability, if it is not null determine what to use.
> 
> Or am I missing something?

I haven't looked into whether the overall logic is correct, only that `ServerHandler` doesn't need to null checked twice.

Since `||` is short-circuiting we don't need to null check `ServerHandler` on the right-hand-side.

[Example](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEUCuA7AHwAEAmARgFgAoagb2oAImGA3AQygbABsBLGPBgYBeBnhgB3BgGUAnnjABhPgIwAKAJS0ZMKC10AJNngAm3XSLE5u3BgF8A3I2ZEyATjU9+ggHQAxXnMAUQReAGcMMIB5KABzNQAiBI0NJypmBlcPL1V/QJgQ8MiABShE5NTqO2pqLM00umcmdk4cwUtxKTkFZW91LR09Q2MzC1FOzXs0jLq2jDzg0IjouPKU6Zd3TxVfAMWisNK1yqpqmio6k8b05hauHaFxyRl5JQfNbWldfSgjU3NOE8JJNHE1Mls5gsCktIjF4gkAA5sDAACwqGyYsweUMKyyOiORaPWVRq51JpBePQe1xmAGZBj8/qMoAB+BnDf4WbSxGAYBwMMK8/lnDJgoj04AQCC2PbQg5wtSuAAMDCRqI0IgAfOzfiMASJxtZbAQCAw1F8hrrOZwAISGmwMABkjoYNotjL1uh8ABEIIIAKqC6S8ABeBX0eAAkgAzRRsJHAcxqNUo4k3JhiiVSmX5XElMrK1WEjXCbXujnMg1WB0m13lq3Mn1+jCBmDBsNBCMxuMJpMpk5nWokHVMgE0lxZ6UMX0BoOh8MCbvxtiJmCKsgq/taouoqsE1EJNJ2IA==)

FanDjango:
IMO you are forgetting the **left side** of the "&&" (also in your gist) is `ServerHandler == null && FtpCapability.MDTM`.  Thus we might actually evaluate the right hand side in the case where `ServerHandler` **is** `null`, namely when `FtpCapability.MDTM` is `false`.

FanDjango:
In your gist you use

```
ServerHandler == null || (ServerHandler != null && !ServerHandler.DontUseSizeEvenIfCapable(path)
```
which is not the same.

Here we have
```
ServerHandler == null && a_third_parameter   ||   (ServerHandler != null && !ServerHandler.DontUseSizeEvenIfCapable(path)
```

FanDjango:
Why don't we separate that one out and commit/merge all the others and discuss this one further?

FanDjango:
I'd really like to get the others in...

PR #1741, opened 2025-04-23
Remove redundant null checks

jnyrup:
For the changed C# expressions, let:
* `a` be `HasFeature(FtpCapability.SIZE)`, 
* `b` be `ServerHandler == null` and
* `c` be `ServerHandler.DontUseSizeEvenIfCapable(path)`.

On mathematical form, the expressions are now `a && (b || !b && c)`.
As proven by [Wolfram Alpha](https://www.wolframalpha.com/input?i=%28a+%26%26+%28b+%7C%7C+%21b+%26%26+c%29%29+%3D+%28a+%26%26+%28b+%7C%7C+c%29%29) this is equal to writing `a && (b || c)`
image

Translating back to C# it means we can omit the `ServerHandler != null` checks.

### DIFF
--- a/FluentFTP/Client/AsyncClient/FileExists.cs
+++ b/FluentFTP/Client/AsyncClient/FileExists.cs
@@ -27,7 +27,7 @@ namespace FluentFTP {
 
 			// since FTP does not include a specific command to check if a file exists
 			// here we check if file exists by attempting to get its filesize (SIZE)
-			if (HasFeature(FtpCapability.SIZE) && (ServerHandler == null || (ServerHandler != null && !ServerHandler.DontUseSizeEvenIfCapable(path)))) {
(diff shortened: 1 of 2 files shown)


FanDjango:
Looks to me that I am on some different code than you. When I look at your PR and compare to my working branch, it does not coincide.

We have been talking at cross purposes, I am sorry about that. 

Let me research what the code **should** actually look like.

FanDjango:
Let me check the **intended logic** for validity first - your intended corrections are correct, but my current branch/working tree has differing code, thus the confusion (on my side).

FanDjango:
The code is correct and thus the change is also valid. Sorry for the confusion and thanks for your patience.

jnyrup:
Great to hear. 
I also began wondering if we were looking at the same code.

PR #1748, opened 2025-05-08
Fixes #1747: Make sure a OperationCancelledException is thrown when cancelling a AsyncConnect

### LINKED ISSUE #1747: TimeoutException is thrown instead of TaskCancelledException when FtpClient.ConnectAsync() is called with cancelled token and the token is cancelled
In the ConnectAsyncHelper the cancel token is linked with a timer:

		private async Task<bool> ConnectHelperAsync(IPAddress ipad, int port, CancellationToken token) {
			try {
					timeoutSrc.CancelAfter(ConnectTimeout);

Later the exception is caught but the OperationCancelled is treated as a timeout:

			}

Changing this to will fix this by checking if the source was the token of from the timeoutSrc:

				token.ThrowIfCancellationRequested();
			}

bruinsg:
The cancellation token was combined with the timeout source and the resulting cancellation was interpreted as a timeout instead of and cancelled operation.

### DIFF
--- a/FluentFTP.Tests/Unit/TimeoutTests.cs
+++ b/FluentFTP.Tests/Unit/TimeoutTests.cs
@@ -61,5 +61,28 @@ namespace FluentFTP.Tests.Unit {
 			}
 		}
 
+		[Fact]
+		public async Task ConnectTimeoutAsyncCancel() {
+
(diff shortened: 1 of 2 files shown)


FanDjango:
Thanks for improving this.

PR #1760, opened 2025-06-08
Remove FileZilla Integration Test

FanDjango:
I'm fed up.

They make it very difficult to put together an image for filezilla-server, from source code, from a .deb package or from binaries.

There is no pre-made docker image to be found.

The real problem is that their download pages, from where one could get the current version, uses java script and some php magic to prevent a "non-human-browser" download. But that would be exactly what we would need to build our image.

Anybody out there that wants this to work is heartily invited to try...


robinrodricks:
Thanks for this change!

Issue #1746, opened 2025-05-07
Handle ill-formed responses


kowalski-se:
Before I start, I know this is a completely stupid topic and I would be perfectly fine if you decide to reject this 😄 Nevertheless, I considered asking is free 😄

We're dealing with a custom FTP server that is not really RFC compliant. The server defines custom FTP commands and for some of them, it doesn't respond with a valid message, in particular the leading response code (e.g., 200) is missing. For those commands, the communication runs into timeout, although the server did respond, if also with an ill-formed message. Usually one would fix the server, but it's an old system, so we don't really have that possibility.

Would it be possible to adapt the implementation in a way, that the decision whether a response has arrived, is not bound to the leading response code? I don't really care how this is handled in the end, e.g., there could be an exception, but it would be ideal for this to happen immediately, not after the timeout. I'm not sure if there is a way to implement this without breaking anything else, but if there is, it would be awesome to have this 😄

FanDjango:
Do you invoke the custom commands using `Execute()`?

FanDjango:
I would envision a client config or an additional parameter causing all `Executed()` commands to be considered successful and that would give you the possibility of examining the response yourself. Would that work for you?

kowalski-se:
Yes, we use `IAsyncFtpClient.Execute(string, CancellationToken)`. That would work perfectly 😄

FanDjango:
I'll see what I can do and keep you posted. Pls standby.

FanDjango:
I've got a fork with a branch that has a tentative solution for your problem. Are you in a position to use (for testing this) such a branch instead of a NuGet package?

kowalski-se:
Of course, I assume it's the branch `No-Reply` since this is the most up to date one? This one is working perfectly :smile:
Can you please add the new method `Task<FtpReply> Execute(string, int, CancellationToken)` to the interface `IAsyncFtpClient` too?

Just one question to the code changes. I'm not sure whether these lines make sense:
![Image](https://github.com/user-attachments/assets/807b29e8-1033-4be9-9f94-c1928b600c86)

First, if we run into timeout before reaching the expecting lines, a timeout exception should still be thrown in my opinion.
Second, the semaphore is released in the finally-block already.
If I haven't overlooked anything, I think these lines can be removed :smile:

FanDjango:
> I assume it's the branch No-Reply since this is the most up to date one?

[...]

PR #1763, opened 2025-07-03
Rework accepting custom responses (#1746)

kowalski-se:
Sorry to bring this dogshit of #1746 up again, but after a while of testing I discovered that the previous approach wasn't quite good enough after all. We also have scenarios where we have a valid positive response, but the error case of the same command brings up an ill-formed response. In addition, there are multiline success responses while the error response is single-line, well, all of the possible ill-formed responses are single-liners (as far as I can tell currently).

As a result, I figured to change this `linesExpected` approach a bit to better fit these requirements. Feel free to have a look, I'd be also open to keep the special case code in my own library, in which case it would be fine to make the `GetReply` method(s) virtual so that I can inherit a subclass, overwriting the essential parts 😄

There is one small change in the handling of multiline responses (independent of the actual changes) though, previously whenever there was a "well-formed" FTP response within the multilines, the FtpReply would be handled as completed single line response. This is no longer the case, now we actually wait for the final closing line. This should technically be an improvement as per RFC, but on the other hand if we occur a disconnect in the middle of a multiline response, we're kinda stuck in the loop until timeout. Which ever way of handling this you find preferable, I'd be fine with, this is just an fyi. Example of this case could be something like this:
```
123-First line  
Second line  
  234 A line beginning with numbers  
123 The last line
```

### DIFF
--- a/FluentFTP.Tests/Unit/ParserTests.cs
+++ b/FluentFTP.Tests/Unit/ParserTests.cs
@@ -230,7 +230,7 @@ namespace FluentFTP.Tests.Unit {
 				new FtpListItem("data.0000", 9, FtpObjectType.Link, new DateTime(2011, 9, 27, 0, 0, 0, 0)),
 				new FtpListItem("data.6460", 512, FtpObjectType.Directory, new DateTime(2012, 6, 29, 0, 0, 0, 0)),
 				new FtpListItem("sys.0000", 8, FtpObjectType.Link, new DateTime(2011, 9, 27, 0, 0, 0, 0)),
-				new FtpListItem("sys.6460", 4096, FtpObjectType.Directory, new DateTime(2024, 6, 25, 16, 26, 0, 0)),
+				new FtpListItem("sys.6460", 4096, FtpObjectType.Directory, new DateTime(DateTime.UtcNow.Year, 6, 25, 16, 26, 0, 0)),
 				new FtpListItem("File001.xml dir", 512, FtpObjectType.Directory, new DateTime(1994, 4, 8, 0, 0, 0, 0)),
 				new FtpListItem("File003.xml dir", 512, FtpObjectType.Directory, new DateTime(1994, 4, 8, 0, 0, 0, 0)),
 				new FtpListItem("File-3.txt file", 1392298, FtpObjectType.File, new DateTime(1995, 11, 22, 0, 0, 0, 0)),
--- a/FluentFTP/Client/AsyncClient/Execute.cs
+++ b/FluentFTP/Client/AsyncClient/Execute.cs
@@ -17,17 +17,17 @@ namespace FluentFTP {
 		/// <param name="token">The token that can be used to cancel the entire process</param>
 		/// <returns>The servers reply to the command</returns>
 		public async Task<FtpReply> Execute(string command, CancellationToken token = default(CancellationToken)) {
-			return await Execute(command, -1, token);
+			return await Execute(command, false, token);
 		}
 
 		/// <summary>
 		/// Performs an asynchronous execution of the specified command
 		/// </summary>
 		/// <param name="command">The command to execute</param>
-		/// <param name="linesExpected">-1 normal operation, 0 accumulate until timeOut, >0 accumulate until n msgs received</param>
+		/// <param name="acceptIllFormed">A flag indicating whether to accept ill-formed responses.</param>
 		/// <param name="token">The token that can be used to cancel the entire process</param>
 		/// <returns>The servers reply to the command</returns>
-		public async Task<FtpReply> Execute(string command, int linesExpected, CancellationToken token = default(CancellationToken)) {
+		public async Task<FtpReply> Execute(string command, bool acceptIllFormed, CancellationToken token = default(CancellationToken)) {
 			FtpReply reply;
 
 			bool reconnect = false;
@@ -124,7 +124,7 @@ namespace FluentFTP {
 				LastCommandTimestamp = DateTime.UtcNow;
 
 				// get the reply
-				reply = await ((IInternalFtpClient)this).GetReplyInternal(token, command, false, 0, false, linesExpected);
+				reply = await ((IInternalFtpClient)this).GetReplyInternal(token, command, false, 0, false, acceptIllFormed);
 			}
 			finally {
 				m_daemonSemaphore.Release();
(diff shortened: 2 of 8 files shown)


FanDjango:
>  I discovered that the previous approach wasn't quite good enough after all

> As a result, I figured to change this linesExpected approach a bit to better fit these requirements

I can't say that I am surprised. I initially actually wanted to code a very lenient and flexible approach, allowing you to define all sorts of things to "accept", perhaps using some formal notation.

I'll go over your changes. A first quick look is positive. Have you run them through the integration tests?

FanDjango:
> even with help of the documentation

'twould be interesting where the problem lies, but.... before that:

The TimeoutTests ( ConnectTimeout ) fails. That is not an integration test. Here is the screenshot:

image

I will revert the merge until this is cleared up.

kowalski-se:
Problem with the integration tests could be anything. I have a complicated setup at work, a VM on a VSphere server with couple of proxy and firewall settings. Running the `.bat` file also throws some errors like this:
```
C:\Projects\FluentFTP\FluentFTP.Dockers>docker build common-mirror --pull --progress=plain -t common-mirror:fluentftp
#0 building with "desktop-linux" instance using docker driver

#1 [internal] load build definition from Dockerfile
#1 transferring dockerfile: 348B 0.0s done
#1 DONE 0.1s

#2 [internal] load metadata for docker.io/library/python:3.10-slim
#2 ERROR: failed to do request: Head "https://registry-1.docker.io/v2/library/python/manifests/3.10-slim": EOF
------
```
Didn't have too much time to dig into every issue I occur, at this point I simply gave up.

And regarding the unit tests, as I said, all succeed on my machine after I fixed the one with the date time parsing, so no idea what to do about this 😮

image

Or if you prefer the command line execution:

```
C:\Projects\FluentFTP> dotnet test --filter Unit
Restore complete (1,8s)
  FluentFTP net5.0 succeeded (0,5s) → FluentFTP\bin\Debug\net5.0\FluentFTP.dll
  FluentFTP net472 succeeded (0,3s) → FluentFTP\bin\Debug\net472\FluentFTP.dll
  FluentFTP net462 succeeded (0,3s) → FluentFTP\bin\Debug\net462\FluentFTP.dll
  FluentFTP netstandard2.1 succeeded (0,3s) → FluentFTP\bin\Debug\netstandard2.1\FluentFTP.dll
  FluentFTP netstandard2.0 succeeded (0,4s) → FluentFTP\bin\Debug\netstandard2.0\FluentFTP.dll
  FluentFTP net6.0 succeeded (0,5s) → FluentFTP\bin\Debug\net6.0\FluentFTP.dll
  FluentFTP net7.0 succeeded (4,9s) → FluentFTP\bin\Debug\net7.0\FluentFTP.dll
  FluentFTP.Xunit succeeded (1,3s) → FluentFTP.Xunit\bin\Debug\net7.0\FluentFTP.Xunit.dll
[xUnit.net 00:00:00.00] xUnit.net VSTest Adapter v3.0.2+dd36e86129 (64-bit .NET 7.0.20)
[xUnit.net 00:00:00.59]   Discovering: FluentFTP.Xunit
[xUnit.net 00:00:00.61]   Discovered:  FluentFTP.Xunit
  FluentFTP.Xunit test succeeded with 1 warning(s) (2,0s)
    C:\Program Files\dotnet\sdk\9.0.301\Microsoft.TestPlatform.targets(48,5): warning : No test matches the given testcase filter `Unit` in C:\Projects\FluentFTP\FluentFTP.Xunit\bin\Debug\net7.0\FluentFTP.Xunit.dll
  FluentFTP.Tests succeeded (0,9s) → FluentFTP.Tests\bin\Debug\net7.0\FluentFTP.Tests.dll
[xUnit.net 00:00:00.00] xUnit.net VSTest Adapter v3.0.2+dd36e86129 (64-bit .NET 7.0.20)
[xUnit.net 00:00:00.36]   Discovering: FluentFTP.Tests
[xUnit.net 00:00:00.42]   Discovered:  FluentFTP.Tests
[xUnit.net 00:00:00.44]   Starting:    FluentFTP.Tests
[xUnit.net 00:00:06.61]   Finished:    FluentFTP.Tests
  FluentFTP.Tests test succeeded (7,6s)

Test summary: total: 45, failed: 0, succeeded: 45, skipped: 0, duration: 10,4s
Build succeeded with 1 warning(s) in 19,0s
```

FanDjango:
> #1 [internal] load build definition from Dockerfile
#1 transferring dockerfile: 348B 0.0s done
#1 DONE 0.1s

#2 [internal] load metadata for docker.io/library/python:3.10-slim
#2 ERROR: failed to do request: Head "https://registry-1.docker.io/v2/library/python/manifests/3.10-slim": EOF

I'll look into that as time permits

FanDjango:
> I assume your unit test issue resolved itself, since you haven't reverted the changes yet? 

No it did not. So I reverted, you can see that in the master's commits.

I was hoping for you to find out which part of you changes caused that timeout test to fail.

What I did do, is to keep you other change to the parser tests, where the date was used incorrectly.

> Just one more question, do you have a new NuGet release already planned?

No, I don't know about the release schedule at this moment. But that also gives you a chance to investigate the above mentioned problem, so you can get it in **before** the next release.

kowalski-se:
> I don't understand your reaction.

I apologize if anything came across hostile here, please note that this was not my intention.

> You have in effect removed Noop logic. I suppose you do not use NOOP. I certainly need to look into this.

I didn't. The same logic is now a condition in the while loop, see `while (isMultiline || exhaustNoop);`. I was able to do so, since there are no more operations in the loop after `DecodeStringToReply`. So I just moved the conditional continue into the exit condition.

(`while (true)` loops, as well as `break` and `continue` operations are generally considered bad practice, since they confuse the control flow).

> Listen: This thing is useful but it needs to be integrated into the overall code as all users use it, not just you.

Exactly that's the thing I'm not sure about. I want for any FOSS to be as clean, efficient and compliant as possible. While covering non-standard, even non-compliant behavior like this issue, should in my opinion not have any rights to be part of FOSS code. But perhaps this is where I'm lacking experience.

> but what is wrong with taking this as far as possible and ironing out all problems.

If you really must know, I currently just don't have any time to do that, even though I really want to. At least until Oct/Nov this year. After that I'll be happy to check up on this again, I just honestly can't at the moment.

> Oh, I reworked your rework a little in a test branch and it passes the test now.

I'm kind of interested what fixed it :)

FanDjango:
@kowalski-se After some organisational communication with the owner, you might consider **re-opening** your PR, which I would recommend.

PR #1775, opened 2025-08-23
FluentFTP.Logging: fix dependency to Microsoft.Extensions.Logging

mus65:
there is no reason to reference the implementation package.


mus65:
> We have referenced MELA for a couple years now without issues. 

Well, the A in MELA stands for abstractions, so you actually haven't, at least not directly. :smile:

MELA only contains the interfaces, while MEL contains the default implementations. MELA is meant to be used by libraries so they don't pull in the implementations that they don't actually need. See also https://learn.microsoft.com/en-us/dotnet/core/extensions/logging?tabs=command-line#nuget-packages .

Granted, since FluentFTP.Logging is a separate package, this rarely makes a difference in practice since whoever uses this package, is most likely also using MEL (which also means that this change is unlikely to break things). The point is that FluentFTP.Logging pulls in a dependency that it doesn't actually need.

But if you don't want to do it, feel free to close. I don't care *that* much. :wink:

robinrodricks:
> since whoever uses this package, is most likely also using MEL 

Exactly.

PR #1782, opened 2025-09-16
Rework FtpMonitor class's access modifiers

Tiszteletes:
BaseFtpMonitor:
- Set properties internal access modifier to protected, so derived classes can reach them
- Set StartTimer and StopTimer functions access modifier to protected, so derived classes can call them
- Set GetListingOptions and HandleUnstableFiles functions access modifier to protected virtual, so derived classes can override them

AsyncFtpMonitor:
- Set PollFolder and GetCurrentListing functions access modifier to protected virtual, so derived classes can override them

BlockingAsyncFtpMonitor:
- Set Start function access modifier to public virtual, so derived classes can override it
- Set PollFolder and GetCurrentListing functions access modifier to protected virtual, so derived classes can override them

### DIFF
--- a/FluentFTP/Monitors/AsyncFtpMonitor.cs
+++ b/FluentFTP/Monitors/AsyncFtpMonitor.cs
@@ -57,7 +57,7 @@ namespace FluentFTP.Monitors {
 		/// <summary>
 		/// Polls the FTP folder for changes
 		/// </summary>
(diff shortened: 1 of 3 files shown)


robinrodricks:
Thanks, it's released!

https://www.nuget.org/packages/FluentFTP/53.0.2

PR #1790, opened 2025-11-22
Add SSL client authentication options configuration for legacy RSA FTPS servers

mamuleanu:
This PR adds a configuration hook that allows users to customize `SslClientAuthenticationOptions` before the TLS handshake, enabling support for Linux for legacy FTPS servers that only negotiate RSA key-exchange ciphers.

The problem: 

On Linux, .NET’s SslStream does not offer RSA key-exchange cipher suites by default. As a result, FluentFTP cannot complete the TLS handshake on these servers unless it exposes SslClientAuthenticationOptions and allows custom CipherSuitesPolicy.

Microsoft documents this behavior [here](https://learn.microsoft.com/en-us/dotnet/core/compatibility/cryptography/5.0/default-cipher-suites-for-tls-on-linux).

The commits added: 

- Added `ConfigureSslClientAuthenticationOptions` property to `FtpConfig` that allows users to customize `SslClientAuthenticationOptions` before authentication
- Updated TLS handshake code to use `SslClientAuthenticationOptions` pattern in both synchronous and async methods
- Added example file demonstrating how to configure FluentFTP for legacy RSA-only servers

I've created a repository with instructions on how to reproduce the handshake error. The repository, including instructions on how to setup the environment, can be found [here](https://github.com/mamuleanu/FTPS-legacy).

**UPDATE:**

As per PR feedback, the PR now adds an event handler that allows users to customize `SslClientAuthenticationOptions` before the TLS handshake, enabling support for legacy FTPS servers

Changes:

- **Added `FtpSslClientAuthenticationOptions` event delegate and event args class** - Infrastructure for the event
- **Added `ConfigureSslClientAuthenticationOptions` event to `BaseFtpClient`** - Event handler that allows customization of SSL options
- **Added `OnConfigureSslClientAuthenticationOptions` helper method** - Internal method to fire the event (following the same pattern as `ValidateCertificate`)
- **Added example file** - `ConnectFTPSLegacyRSA.cs` demonstrating how to configure FluentFTP for legacy RSA-only servers

### DIFF
--- a/FluentFTP.CSharpExamples/ConnectFTPSLegacyRSA.cs
+++ b/FluentFTP.CSharpExamples/ConnectFTPSLegacyRSA.cs
@@ -1,0 +1,133 @@
+using System;
+using System.Net;
+using System.Net.Security;
+using System.Security.Authentication;
+using System.Threading;
+using System.Threading.Tasks;
+using FluentFTP;
+
+namespace Examples {
+	internal static class ConnectFtpsLegacyRsaExample {
+
+		/// <summary>
+		/// Example: Connect to a legacy FTPS server that only supports RSA key-exchange ciphers.
+		///
+		/// Some legacy FTPS servers (especially on Linux) only negotiate RSA key-exchange TLS ciphers
+		/// (e.g., TLS_RSA_WITH_AES_256_GCM_SHA384) and fail when .NET on Linux offers only ECDHE ciphers.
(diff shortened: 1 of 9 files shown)


robinrodricks:
Thanks, good work.
Event handlers should be added to the ftp client classes and not the config. Config is only for simple properties.

PR #1831, opened 2026-04-30
Further fixes and improvement to uploadfilesinternal.cs

FanDjango:
Mostly cosmetic, but also in the course of investigating some problems with android/xamarin, some changes turned out to be good.


FluentFTP/Client/SyncClient/UploadFileInternal.cs
```diff
@@ -255,8 +255,25 @@ protected virtual FtpStatus UploadFileInternal(Stream fileData, string localPath
 					}
```
FanDjango, line 255: The flush here will be removed in a subsequent commit.

Issue #1839, opened 2026-05-06
No way to set NetworkStream WriteTimeout


sherlock1982:
We noticed that sometimes uploading to FTP is stuck forever.
We can' prove it but after analysis found out that it might hang on the following path:

`FtpDataStream.Write() → FtpSocketStream.Write() → NetworkStream.Write()`

[Basically WriteTimeout is infinite.](https://learn.microsoft.com/en-us/dotnet/api/system.net.sockets.networkstream.writetimeout?view=net-10.0)
As a workaround we can use async calls with cancellation token that timeouts after some time (most probably)

But still makes sense to have this ability in the library

FanDjango:
> [Basically WriteTimeout is infinite.](https://learn.microsoft.com/en-us/dotnet/api/system.net.sockets.networkstream.writetimeout?view=net-10.0)

> The default value, [Infinite](https://learn.microsoft.com/en-us/dotnet/api/system.threading.timeout.infinite?view=net-10.0#system-threading-timeout-infinite), specifies that the write operation does not time out.

We do not have a config value for WriteTimeout, you are right. **We have one for ReadTimeout, and that value is transferred on to the streams ReadTimeout on stream creation**, but not for WriteTimeout.

Your statements are essentially true, for SYNC and also for ASYNC operation, where you would need to use the cancellation token after some elapsed time yourself.

> But still makes sense to have this ability in the library

I agree. I wonder what reasoning led to omitting write timeout?

What should we do? Let's call it a bug, shall we?

FanDjango:
So this need a new config value similar to ReadTimeout, called WriteTimeout. And on stream creation, this value needs to be set in the stream, in the same places and in the same fashion as this takes place for ReadTimeout.

For WriteAsync, one should use the linked-token approach do this also, see ReadAsync.

FanDjango:
A PR is out. Do you want to test the fork/branch of mine that contains the changes?

sherlock1982:
We can't reproduce it locally but if you make a minor release we will try immediately.

Note the following: WriteTimeout is for the whole chunk (while ReadTimeout is merely on how long to wait before returning data).
15 seconds for ReadTimeout is ok (we put it as 120s BTW)

But for WriteTimeout situation is different.
For example I use chunks for 80Kb. 15s/80Kb = 5.5 KB/s. This is extremely low speed but possible. 
What is someone sends 1Gb? What is the reasonable write timeout for them?

[...]

PR #1840, opened 2026-05-06
Add a WriteTimeout to Config, control and data stream, FtpSocketstream

FanDjango:
See issue #1839

@robinrodricks this needs another pair of eyes. It runs all tests ok, runs my own tests too.


robinrodricks:
I think this looks great. You added a timeout with 30 min as the default. Matching the others.

Should be fine. 30 min seems high for ALL the timeouts, but that apart, I think you have implemented it very well.

I would just release it and then let future issues (if any) be reported by users. Impossible to fully test system level changes like this.

robinrodricks:
https://github.com/robinrodricks/FluentFTP/wiki/Timeouts#timeout-settings

PR #1843, opened 2026-05-07
Implement "TrackConfigChanges"

FanDjango:
/// Any configuration changes that may have happened that affect a streams parameters such as timeouts
		/// should be applied before the next read/write operation. This is to ensure that the stream is always
		/// up to date with the latest configuration settings.

This to make this Wiki text actually true and functional:

**How to execute an FTP command with a different timeout?**
**Just set Config.ReadTimeout before executing your FTP command and then set it back to whatever you want.**
```cs
client.Config.ReadTimeout = 100;     // 100 ms timeout
client.Execute(..);                  // execute whatever command here, you can also use high level API
client.Config.ReadTimeout = 10000;   // 10 seconds timeout
```

This was **not working** as described.

Up to the timeout properties were set to the stream at either connect time, or open active/passive datastream. From then on any config changes to timeout values were no applied on the stream.

### DIFF
--- a/FluentFTP/Streams/FtpSocketStream.cs
+++ b/FluentFTP/Streams/FtpSocketStream.cs
@@ -14,6 +14,7 @@ using FluentFTP.Client.BaseClient;
 using FluentFTP.Exceptions;
 using FluentFTP.Helpers;
 using FluentFTP.Streams;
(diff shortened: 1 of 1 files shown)


robinrodricks:
Subtle  but important. Thanks for catching this.

Issue #1847, opened 2026-05-24
Regression in 54.1.2: AsyncFtpClient.Connect ignores pre-cancelled CancellationToken


flozano:
## Summary

In **FluentFTP 54.1.2**, calling `AsyncFtpClient.Connect(CancellationToken)` with a token that is already in the cancelled state silently ignores the cancellation, attempts a real TCP connect, and then surfaces the failure as `System.IO.IOException` instead of `OperationCanceledException`. This is a behaviour change from **54.1.1**, where the same call honours the pre-cancelled token immediately and surfaces it as `OperationCanceledException` (`TaskCanceledException`).

## Minimal repro

Standalone .NET 10 console app, ~25 lines: https://gist.github.com/flozano/d2fd15078436828785809f9f417c7daa

```csharp
using var cts = new CancellationTokenSource();
cts.Cancel();

using var client = new AsyncFtpClient("127.0.0.1", "anonymous", "anonymous@example.com", 1);

try
{
    await client.Connect(cts.Token);
    Console.WriteLine("UNEXPECTED: Connect returned with no exception");
}
catch (OperationCanceledException ex)
{
    Console.WriteLine($"OK: OperationCanceledException ({ex.GetType().Name}: {ex.Message})");
}
catch (Exception ex)
{
    Console.WriteLine($"REGRESSION: {ex.GetType().FullName}: {ex.Message}");
}
```

### Observed output

| Version | Output |
|---------|--------|
| **54.1.1** | `OK: OperationCanceledException after 8 ms (TaskCanceledException: A task was canceled.)` |
| **54.1.2** | `REGRESSION: System.IO.IOException after 8 ms: Failed to connect to host.` |

The 8 ms numbers happen to match because `127.0.0.1:1` refuses the TCP connect immediately. Against an unreachable host (e.g. `192.0.2.1:21` — TEST-NET-1) the 54.1.1 case still completes in milliseconds, while 54.1.2 blocks for the full TCP connect timeout (default ~15 s).

## Why this matters

Most long-running .NET hosts signal graceful shutdown by cancelling the request's / job's `CancellationToken`. Container orchestrators (Kubernetes, AWS ECS, Azure Container Apps, Service Fabric, …) typically allow a bounded drain window (often 30 s) for the process to wind down before sending `SIGKILL`.

- **54.1.1**: in-flight FTP operations abort instantly on cancellation, sockets are released, the host drains cleanly inside the window.
[...]

PR #1848, opened 2026-05-24
Fix for regression / Issue #1847

FanDjango:
See issue #1847
Tests run OK / Regression sample code run OK


robinrodricks:
Released

Fixed in https://www.nuget.org/packages/FluentFTP/54.2.0

PR #1857, opened 2026-06-25
Support Native AOT

0x5bfa:
Resolves #1856

These were the new AOT-related issues, just FYI:

```diff
D:\source\FluentFTP\FluentFTP\Helpers\ValuePrinter.cs(15,21): warning IL2075: 'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicProperties' in call to 'System.Type.GetProperties(BindingFlags)'. The return value of method 'System.Object.GetType()' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to. [D:\source\FluentFTP\FluentFTP\FluentFTP.csproj::TargetFramework=net9.0]
D:\source\FluentFTP\FluentFTP\Client\BaseClient\Logger.cs(29,54): warning IL2026: Using member 'System.Reflection.MethodBase.GetCurrentMethod()' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Metadata for the method might be incomplete or removed. [D:\source\FluentFTP\FluentFTP\FluentFTP.csproj::TargetFramework=net9.0]
D:\source\FluentFTP\FluentFTP\Streams\FtpSocketStream.cs(1530,21): warning IL2072: 'type' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicParameterlessConstructor' in call to 'System.Activator.CreateInstance(Type)'. The return value of method 'FluentFTP.FtpConfig.CustomStream.get' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to. [D:\source\FluentFTP\FluentFTP\FluentFTP.csproj::TargetFramework=net9.0]
D:\source\FluentFTP\FluentFTP\Streams\FtpSslStream.cs(298,11): warning IL2070: 'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicFields', 'DynamicallyAccessedMemberTypes.NonPublicFields' in call to 'System.Type.GetFields(BindingFlags)'. The parameter 't' of method 'FluentFTP.Streams.ReflectUtil.GetAllFields(Type)' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to. [D:\source\FluentFTP\FluentFTP\FluentFTP.csproj::TargetFramework=net9.0]
[...]

PR #1867, opened 2026-09-10
feat: add NativeAOT support

### LINKED ISSUE #1856: Support Native AOT
It looks like this project doesn't officially support Native AOT builds. If already the code is compatible with it, please consider to annotate the project as AOT compatible; otherwise, please fix AOT issues then annonate.

BrycensRanch:
Supersedes https://github.com/robinrodricks/FluentFTP/pull/1857
Fixes https://github.com/robinrodricks/FluentFTP/issues/1856

Built off of the PR. This PR addresses your concerns and passes tests.
There are no AOT warnings now.
If there's any problem, let me know.

Also, I added comments to the ValuePrinter class explaining why `ToString` is added to `FtpAutoDetectConfig` to get around NativeAOT limitations regarding reflection. The behavior for older platforms is identical to master, as you requested.


FluentFTP/Model/Functions/FtpAutoDetectConfig.cs
```diff
@@ -43,5 +43,15 @@ public class FtpAutoDetectConfig {
 			// Do not EVER use "Default". It boils down to "SSL or TLS1.0" or worse.
 			// Do not use "None" - it can connect to TLS13, but Session Resume won't work, so a successful AutoDetect will be a false truth.
 		};
+
+		public override string ToString() {
```
robinrodricks, line 49: this is incorrect and does not match the original implementation. you need to call `ValueToString(x)` for each of the values. make sure the code is identical. also, use a stringbuilder and then join by ","..  dont make one big string.

BrycensRanch: Addressed. Nice spot.

robinrodricks:
Released.

https://www.nuget.org/packages/FluentFTP/55.0.0

Thanks!

PR #1869, opened 2026-09-17
Make BouncyCastle integration tests access all test servers

FanDjango:
I'd like the BouncyCastle stream to also test for access to non-FTPS, to non-Require-Session-Resume servers.

image


robinrodricks:
Thanks a lot!

PR #1870, opened 2026-09-18
One common base for docker integration tests, reduce duplication.

FanDjango:
All stream variants should use the current set of docker test servers. This eliminates the duplication of the (currently) three sets of stream intergration test calls and eases addition of further stream implementations (if ever).

image


robinrodricks:
very good thanks!

